PR #1948, opened 2015-01-15
Add toString to DeadEvent

paolofulgoni:
Given that the main purpose of DeadEvent is logging/debugging, I think that this simple string representation could be useful.


cpovirk:
https://github.com/google/guava/commit/252832d66a77cc19ba8b4d9c40afd2796127404c

PR #1970, opened 2015-02-11
Jdk5 backport v17.0 compatibility

paulduffin:
Reverts (at least in part) a number of changes made to jdk5-backport-v17.0-post to make it more compatible with code built against jdk5-backport-v14.0.1-post, specifically AOSP code. The reverted changes added back methods that had been present in v14 (deprecated and not) and since been removed. The intent is to allow AOSP code that built against v14.0.1 to also build against the compatibility version of v17.0 without modification (or with minimal atomic modifications). That will make upgrading itself easier and will allow other changes such as removing usages of deprecated methods to be done in stages.

### DIFF
--- a/guava-bootstrap/pom.xml
+++ b/guava-bootstrap/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>com.google.guava</groupId>
     <artifactId>guava-parent-jdk5</artifactId>
-    <version>17.0-rc2</version>
+    <version>17.0</version>
   </parent>
   <artifactId>guava-bootstrap-jdk5</artifactId>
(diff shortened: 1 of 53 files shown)


cgdecker:
Rather than merging this into jdk5-backport-master, I think I'll just pull it as a separate branch and give it a tag. Running the tests against a JDK5 rt.jar, I found one test in `ByteStreamsTest` that was calling `getBytes(Charset)` which didn't exist in JDK5. I'll just add a commit fixing that to the branch.

cgdecker:
I've pulled the branch in (plus the fix I mentioned) with the tag `jdk5-backport-v17.0-compatibility` (https://github.com/google/guava/commits/jdk5-backport-v17.0-compatibility).

PR #1971, opened 2015-02-12
google code conventions

rpau:
Hi,

I have applied the eclipse formatter using the Google configuration for Java projects available at https://code.google.com/p/google-styleguide/source/browse/trunk/eclipse-java-google-style.xml
Is there another formatter configuration for this project?

Thanks!

PR #1972, opened 2015-02-12
google code conventions

rpau:
Thi is an updated version with the CLA email that updates the whole codebase with the eclipse formatter that Google provides.


cgdecker:
Kevin mentioned our issues with the Eclipse formatter in #1971, so I'm going to close this too. In addition to the Eclipse formatter issue, I'd also note that anything affecting large swaths of the codebase like this really needs to be done internally and then synced out. We just don't have a good way to make changes in the other direction.

PR #1984, opened 2015-03-03
Add a weight() method to Cache

cheddar:
Cache objects currently do not have a method of introspecting on the current "weight" value.  I am replacing an implementation of a cache that is just a synchronized LinkedHashMap.  In my current implementation of the cache, I'm keeping track of the weights of things being stored and exposing that as a metric about my cache.  This is the _only_ feature of my current implementation that is not covered by Guava's Cache.

From looking at the code, it seemed relatively easy to just create another method parallel to `size()` to handle this case so I just went ahead and did it.  The guidelines do state that API changes should start as an issue, but I didn't realize that until I had already done this and it is such a simple fix, it doesn't bother me if it gets thrown away (as long as I do get some method of introspecting on the current weight value ;) ).

I do realize that this is a public API change, it is only an addition, but it does mean that if anyone else is implementing the Cache interface from scratch, their code will have to be updated before it can work with this.  I'm not sure what level of API compatibility is maintained with Guava, but as long as people are only using Guava-built cache objects, it should be a forward-compatible change.

Also, I updated unit tests as I could find ones that seemed relevant, but I will admit that I didn't read and comprehend all tests, so I might've missed a good place to add some more verification of the weight computation.  Let me know if there are other tests to update and I'll be happy to.

If you think this is going to work, let me know and I'll make sure to get a CLA.

### DIFF
--- a/guava-gwt/src-super/com/google/common/cache/super/com/google/common/cache/LocalCache.java
+++ b/guava-gwt/src-super/com/google/common/cache/super/com/google/common/cache/LocalCache.java
@@ -403,6 +403,11 @@ public class LocalCache<K, V> implements ConcurrentMap<K, V> {
     }
 
     @Override
+    public long weight() {
+      return localCache.weight();
+    }
+
+    @Override
     public ConcurrentMap<K, V> asMap() {
       return localCache;
     }
--- a/guava-tests/test/com/google/common/cache/CacheEvictionTest.java
+++ b/guava-tests/test/com/google/common/cache/CacheEvictionTest.java
@@ -77,15 +77,18 @@ public class CacheEvictionTest extends TestCase {
   }
 
   public void testEviction_maxWeightOneSegment() {
+    int maxWeight = 2 * MAX_SIZE;
+
     IdentityLoader<Integer> loader = identityLoader();
     LoadingCache<Integer, Integer> cache = CacheBuilder.newBuilder()
         .concurrencyLevel(1)
-        .maximumWeight(2 * MAX_SIZE)
+        .maximumWeight(maxWeight)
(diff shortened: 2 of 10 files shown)


cheddar:
It seems to be failing when trying to generate the gwt source.  I'll see if I can repro the failure on my local machine and try to fix, but any pointers to what might cause the gwt compilation to fail would be helpful.

cheddar:
Ok, on my local machine, I cannot seem to download com.google.guava:guava:java-source:19.0-SNAPSHOT

Looks like it is not in sonatype-nexus-snapshots

So, it's failing, but for a different reason...  I guess I'll wait on fixing that until the discussion of whether this API change could even be considered has happened.

ben-manes:
The predecessor, [ConcurrentLinkedHashMap](https://github.com/ben-manes/concurrentlinkedhashmap) introduced the concept of weights and provides a `weightedSize` method. That may be a simpler migration path. (Note: currently moving to Github due to Google Code shutdown)

The JDK8 rewrite, [Caffeine](https://github.com/ben-manes/caffeine), exposes additional inspection including the weighted size through the `Policy` abstraction. That project is maturing quickly but not yet released.

cheddar:
@ben-manes So, are you suggesting that instead of using a `Cache` in my code I use `ConcurrentLinkedHashMap`?

Can't really use something that isn't released yet :)

ben-manes:
I'm offering it as an option as your changes may not be accepted and you would have to wait for the next release if they are. If all you need are the basic features of its predecessor, then its a reasonable option. The old wiki is in the Github wiki branch since the migration didn't move it over properly.

cpovirk:

PR #1990, opened 2015-03-07
Created a Changelog

silentmoose:
Created a CHANGELOG in the root directory. With proper formatting the app could also read from this file and view the changes within the app. Changelog should be updated with each pull accept. Also, as main developers release updates to the code other contributors will now be able to see the change in the changelog.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,0 +1,1 @@
+#Changelog for each release/update: Should be updated as pulls are accepted.


cgruber:
So first of all, this should be CHANGELIST.md.  Much more useful as a markdown file. 

Second, I'm not sure this file adds much value, as a placeholder.  It doesn't include the massive back-fill of change log information, and doesn't provide much structure so it's not really a pull request so much as a documentation request (which is fair, but that's an issue, not a p/r)

Also, we have so much coming out from internal commits, and google's internal commits are usually unique, meaningful changes with a reasonable commit description, that it is nearly certain to be just as valuable to automatically crib the descriptions from the list of commits and use that as a first-draft log of changes.  Maintaining one manually and incrementally may be worth it, but I'm not convinced.

@cgdecker - any opinions?  I'm not sure what you had in mind for release changelogs.  My personal feeling is that we take the full list from the last release and trim it into a changelog.  I do like the idea of demonstrating the delta, the additional features/changes/etc.

silentmoose:
Appreciate your detailed explanation. I do feel that a changelog that users can quickly skim to see new additions is important. Now as far as actual implementation obviously that does not matter as much to me as the actual active devs.

ben-manes:
fyi, you should consider using [Github releases](https://github.com/blog/1547-release-your-software) instead of a changelog.

cpovirk:
I think Colin puts together a pretty good list for each release, e.g.:
https://github.com/google/guava/wiki/Release21

We also provide automated diffs between the current snapshot and the previous release:
http://google.github.io/guava/releases/snapshot/api/diffs/

We could try to keep a running changelog between releases, but I fear it will never become a high enough priority in comparison to the changes themselves, given the existing things we provide here.

PR #2002, opened 2015-03-17
Add callbacks to EventBus

smetarom:
I would appreciate a feature, where I can  add additional functionality to EventBus. My main goal is to extend logging and debugging capabilities of EventBus.

My proposal is to create callbacks, which are overwritten in the child class. As at is now, I have to do logging before I call EventBus.post and it creates code duplication.

The simple example:

EventBus eventBus = new EventBus() {
   @Override
   protected void onEventPosted(Object event) {
      // do some logging
   }
};

I added onRegister, onUnregister, onEventPosted, and onNoSubscriberRegistered callbacks to corresponding methods.

### DIFF
--- a/guava/src/com/google/common/eventbus/EventBus.java
+++ b/guava/src/com/google/common/eventbus/EventBus.java
@@ -183,7 +183,8 @@ public class EventBus {
    * @param object  object whose subscriber methods should be registered.
    */
   public void register(Object object) {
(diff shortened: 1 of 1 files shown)


kluever:
We're unlikely to add new features to `com.google.common.eventbus` in the near future (and will in fact be recommending alternatives).

PR #2012, opened 2015-03-23
Alternative way to get AssertionError instance in

songofsp3:
Alternative way to get AssertionError instance in com.google.common.util.concurrent.GeneratedMonitorTest, since 
public AssertionError(String message, Throwable cause) 
does not exist in Java 6 yet.

### DIFF
--- a/guava-gwt/src-super/com/google/common/collect/super/com/google/common/collect/ImmutableMultiset.java
+++ b/guava-gwt/src-super/com/google/common/collect/super/com/google/common/collect/ImmutableMultiset.java
@@ -363,7 +363,8 @@ public abstract class ImmutableMultiset<E> extends ImmutableCollection<E>
   }
 
   static class EntrySetSerializedForm<E> implements Serializable {
-    final ImmutableMultiset<E> multiset;
+	private static final long serialVersionUID = 0L;
+	final ImmutableMultiset<E> multiset;
 
     EntrySetSerializedForm(ImmutableMultiset<E> multiset) {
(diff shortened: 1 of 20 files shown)


cpovirk:
Our current policy is that Guava runs with Java 6+ but that it may require a newer version to build. I would not mind if we fixed this occurrence up, but we currently have no way to prevent future violations. Animal Sniffer, which we use for Java 6 enforcement, seems to work only on the main compile, not the test compile. (I was able to run it only by telling Maven that our test sources were normal sources and that all of our test-only deps were normal deps.) If anyone can figure out a way to run it on the test sources, we'd probably be interested.

(Also, see the CLA warning above. If it's something you'd rather not sign, I can submit my own workaround. (This is a problem we've had to confront in ours tests previously where it affected GWT compatibility.))

songofsp3:
I thought I have signed the cla, but only after I created the pull request.
Is that the reason, or do I need to sign a different one?

cpovirk:
I guess that the bot just takes a little while to run. Or maybe someone needs to come and say the magic work "CLA?" Anyway, looks good now. Thanks!

cgdecker:
I think the bot might just require someone (possibly the author?) to comment on the pull request again after the CLA has been signed for it to check again.

cpovirk:
Done in https://github.com/google/guava/commit/9131ddb3bd2ff2f3ad08dd06f59d0597d33b8bdf

It looks like we got the commit metadata wrong -- sorry.

PR #2017, opened 2015-03-29
Add missing serialVersionUID field to Converter and VerifyException classes.

mtuchowski:
Both Converter and VerifyException implement Serializable interface, so they should declare a static final serialVersionUID field.


mtuchowski:
I added second commit (*Deprecated Objects.toStringHelper()...*) by mistake to this pull request...

PR #2020, opened 2015-03-31
I suggest adding generators to com.google.common.collect.Iterators

eirikma:
I suggest adding (possibly infinite) generators to the Iterators class. The idea is to have an Iterator presenting a possibly infinite sequence of values, generated on the fly a generator function so you don't have to keep all of that in memory at once. 

Two use cases are provided as unit tests: 
a) some kind of endless computation 
b) iterating over possibly infinite input.

### DIFF
--- a/guava/src/com/google/common/collect/Iterators.java
+++ b/guava/src/com/google/common/collect/Iterators.java
@@ -33,6 +33,7 @@ import com.google.common.base.Objects;
 import com.google.common.base.Optional;
 import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
+import com.sun.istack.internal.NotNull;
 
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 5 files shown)


eirikma:
I've signed it now.

eirikma:
Found it, at last. There was a wrong email address in the commits. I
authored everything myself.

guava/src/com/google/common/collect/Iterators.java
```diff
@@ -1135,6 +1136,139 @@ public T nextElement() {
   }
 
   /**
+   * Create an possibly infinite Iterator that uses the provided Generator
```
peteryhwong: According to the javadoc standard in https://google-styleguide.googlecode.com/svn/trunk/javaguide.html, it should be 'Creates' and not Create'. This will also be consistent with other methods of this class.

cpovirk:

Issue #1742, opened 2014-10-31
Iterables.get() alternative returning Optional.absent()?


gissuebot:
_[Original issue](https://code.google.com/p/guava-libraries/issues/detail?id=1742) created by **fgaliegue** on 2014-05-05 at 07:38 AM_

---

Right now, there are two .get() methods in Iterables to retrieve an element at a given index: one which will return a default value if no element is there, and one which will throw an IndexOutOfBoundsException.

The problem is that both have a @﻿Nullable return type; and I am aware that the possibility exists that an Iterable _can_ yield null values.

Still, it would be nice if there were, say, a:

```
@Nonnull
public static <T> Optional<T> optionalGet(
    @Nonnull Iterable<T> iterable, int index)
```

which would throw an NPE if null elements are encountered in the Iterable, Optional.absent() if no such index, or Optional.of(theValue) if the value is found.

I am also aware that if you'd:

```
Iterables.optionalGet(iterable, 1)
```

where iterable has contents:

```
[ 0, 1, null ]
```

then the method wouldn't trigger the NPE; but then that can be documented:

```
Note that this method will only walk as much as needed to find the
required element; if you use this method on an Iterable potentially 
returning null elements but an element is found before a null is hit,
this method will return an Optional.of() the found element and not
throw a NullPointerException.
```

Comments?

PR #2025, opened 2015-04-07
Iterables.get() alternative returning Optional.absent()

gotenxds:
Hi, decided it will be nice to have an Optional version to the get methods of Iterables.
Also as suggested here https://github.com/google/guava/issues/1742 but with a minor difference:

Both methods will return Optional.absent if the position is grater then the requested index and will also return Optional.absent if the value in the requested index is null.

thoughts ?

### DIFF
--- a/guava-tests/test/com/google/common/collect/IterablesTest.java
+++ b/guava-tests/test/com/google/common/collect/IterablesTest.java
@@ -746,6 +746,23 @@ public class IterablesTest extends TestCase {
     } catch (IndexOutOfBoundsException expected) {}
   }
 
+  private void testOptionalGetOnAbc(Iterable<String> iterable) {
+    try {
+      Iterables.optionalGet(iterable, -1);
+      fail();
+    } catch (IndexOutOfBoundsException expected) {}
+
+    assertEquals("a", Iterables.optionalGet(iterable, 0).get());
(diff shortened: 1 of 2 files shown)


gotenxds:
I signed it.

lowasser:
This looks like you could easily do it yourself as Optional.fromNullable(Iterables.get(Iterable, into, null)).  Is it worth adding another method when it's that easy to roll yourself?

gotenxds:
Sure you could do 
```Java
Optional.fromNullable(Iterables.get(iterable, index, null));
```
But I think it's quite a mouthful, plus there is that null where the default value is which is not a big deal when you use the 
```Java
optionalGet(list , 4, myDeafultValue)
```
But it reduces readability when all you need is 

```Java
optionalGet(list , 4)
```

Debugging a code with this line
```Java
Optional.fromNullable(Iterables.get(iterable, index, null));
```
Will make you enter `iterables.get` implementation to understand the null value.

Loki-Afro:
:thumbsup: I agree with gotenxds.

drothmaler:
I like the idea, but i propose to change the name to `tryGet`, because it aligns better with `find`/`tryFind` and as a bonus it's even shorter :-)

cpovirk:
In general, we now steer users toward `Stream` over `Iterable`, including for this sort of thing.

PR #2045, opened 2015-05-06
Removed unnecessary null check from io.BaseEncoding.encode(bytes)

javajosh:
The check is done on line 166, so the first check is redundant.


javajosh:
I assume I just signed away my first-born. Well, you can have him, Google.

kluever:
Thanks Josh! We don't have a good mechanism for merging patches, so I'm going to submit a change internally and it should get mirrored out shortly.

There were actually a handful of unnecessary checkNotNulls in this file...I think I got 'em all.

Thanks again!

PR #2062, opened 2015-05-28
Add deprecation for compare(Boolean, Boolean)

andrewgaul:
This method exists to warn callers of `compare(Boolean, Boolean)` and
`compare(boolean, boolean)` with a deprecation warning.  Using the
private access modifier would still allow callers to use the generic
`compare(Comparable<?>, Comparable<?>)` method.


kluever:
I'm submitting a version of this change internally and it'll get mirrored out shortly. Thanks for pointing it out!

Issue #1190, opened 2014-10-31
Splitter.onPattern drops last token when zero width matches are used


gissuebot:
_[Original issue](https://code.google.com/p/guava-libraries/issues/detail?id=1190) created by **travis.downs** on 2012-11-04 at 10:17 AM_
---

If you split on a zero-width regex, the last element may be dropped.

String input = "foo";
String regex = "(?=o)|(?<=o)";
Splitter splitter = Splitter.onPattern(regex);
System.out.println(Arrays.asList(input.split(regex)));
System.out.println(Arrays.asList(Iterables.toArray(splitter.split(input),String.class)));

This does zero-width lookaround for 'o', so the string will be split before and after any o, but the o characters will also be returned as individual items.

Note that String.split works correctly here, but Guava splitter drops the last o.

Examining the code, this was probably introduced in this fix (granted, it didn't work at all before that):

https://github.com/google/guava/issues/936

In particular, this bit of logic in Splitter.java:

```
    if (offset == nextStart) {
      /*
       * (ommit comment)
       */
      offset++;
      if (offset >= toSplit.length()) {
        offset = -1;
      }
      continue;
    }
```

neglects to the the "last element" handling that the general logic above does (setting end to toSplit.lenth() if no more separators are found).

gissuebot:
_[Original comment](https://code.google.com/p/guava-libraries/issues/detail?id=1190#c17) posted by **cpovirk@google.com** on 2013-03-11 at 02:38 PM_

---

I just got a very helpful response on the Perl list:
http://www.xray.mpe.mpg.de/mailing-lists/perl5-porters/2013-03/msg00390.html

\[edit: [new link](https://www.nntp.perl.org/group/perl.perl5.porters/2013/03/msg199985.html) with [Internet Archive backup](https://web.archive.org/web/20231202221404/https://www.nntp.perl.org/group/perl.perl5.porters/2013/03/msg199985.html)\]

cpovirk:
We've just learned that this is particularly bad when the input string is only one character:

    public void testPatternSplitWordBoundary_singleCharInput() {
      String string = "f";
      Iterable<String> words = Splitter.on(Pattern.compile("\\b")).split(string);
      assertThat(words).containsExactly("f").inOrder();
    }

This test fails: `words` is completely empty.

jbduncan:
How is this going along? Is there a timeframe as to when a fix can be expected?

cpovirk:
[...]

PR #2086, opened 2015-06-19
Splitter fails with empty matches at the end of a String

JeffreyFalgout:
If you try to split "abc82" with the pattern "(?=\d)" you get [abc, 8].
The last number will be left out of the result.

http://stackoverflow.com/q/30941743/758280

### DIFF
--- a/guava-tests/test/com/google/common/base/SplitterTest.java
+++ b/guava-tests/test/com/google/common/base/SplitterTest.java
@@ -782,4 +782,10 @@ public class SplitterTest extends TestCase {
     } catch (IllegalArgumentException expected) {
     }
   }
+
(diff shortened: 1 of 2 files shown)


lowasser:
Do all the other tests pass?

IIRC, we've made a best-effort attempt to block patterns with zero width from working at all, because no matter what we did, something reasonable-looking wouldn't work.

JeffreyFalgout:
All of the other tests passed.

lowasser:
Previous discussion on the subject in https://github.com/google/guava/issues/1190

cpovirk:
Sorry for dropping this. We ended up accepting https://github.com/google/guava/pull/2615 instead (for no good reason).

PR #2099, opened 2015-07-07
Added missing @throws tag to ConcurrentHashMultiset.removeExactly

AlbertoGoffi:
When the method `removeExactly` from the class `ConcurrentHashMultiset` is invoked with a negative number as second parameter, the method correctly throws `IllegalArgumentException`. This behavior is not documented in the javadoc comment of the method, since the `@throws` tag is missing. I added the missing tag.


AlbertoGoffi:
I signed it!

cpovirk:
Committed: https://github.com/google/guava/commit/ee50c94031ae9449d52802d03241791cc2964194

Thanks!

PR #2106, opened 2015-07-14
Add missing @throws tags to methods of the class Verify

AlbertoGoffi:
Some methods of the class `Verify` miss the proper `@throws` tag. To be consistent with the method `Verify.verify(boolean, String, Object...)`, which already contains the proper `@throws` comment, I added the `@throws` tag where missing.


cpovirk:
Thanks. We'll get this merged shortly.

PR #2111, opened 2015-07-19
Use thread-local buffers to avoid memory allocation and improve performance

berndhopp:
# Abstract

goal of this PR is to replace the often seen but unneccessary 
```
byte[] buf = new byte[BUF_SIZE];
```
or 
```
ByteArrayOutputStream baos = new ByteArrayOutputStream();

```
 with thread-local instances of the respective types in all cases where they are used method-internal, this way avoiding unneccessary memory allocation, increasing performance and enhancing the memory profile. 

# Implementation

[ThreadLocalBuffers](https://github.com/berndhopp/guava/blob/master/guava/src/com/google/common/io/ThreadLocalBuffers.java) is the manager of thread-local buffers. It uses [SoftReferences](https://docs.oracle.com/javase/6/docs/api/java/lang/ref/SoftReference.html) to store buffers, this way enabling Garbage Collection to collect the buffers when the system runs low on memory. This class is package-local right now but can eventually be made public for other scenarios where a buffer is needed.
All allocations of byte[] or ByteArrayOutputStream where the allocated byte[]/BAOS was not the returned value of the method have been replaced with calls to ThreadLocalBuffers. 

ByteSource.contentEquals was challenging, because it used not one but two byte-arrays for buffering, so I used two ByteBuffers that wrap around the upper and lower part of the thread-local buffer, thereby mimicking the two buffers.

# Results

The microbenchmarks show that avoiding buffer allocation yields a significant performance boost. 

|tested method|bytes copied|runtime w/ TL buffers|runtime w/out TL buffers|performance increase in %|
|------------------|-----------------|---------------------------|--------------------------------|-----------------------------------|
|ByteStreams.copy|10|105569|681130|645,2|
|ByteStreams.copy|100|107411|698152|649,98|
|ByteStreams.copy|10000|840354|1186307|141,17|
|ByteStreams.readBytes|10|68917|633365|919,03|
|ByteStreams.readBytes|100|69802|660336|946,01|
|ByteStreams.readBytes|10000|360357|934588|259,35|
|ByteStreams.toByteArray|10|122681|750095|611,42|
|ByteStreams.toByteArray|100|131705|773561|587,34|
|ByteStreams.toByteArray|10000|1962961|4962929|252,83|
|ByteSource.contentEquals|10|647249|3614650|558,46|
|ByteSource.contentEquals|100|685559|3586983|523,22|
|ByteSource.contentEquals|10000|696307|9460941|1358,73|

you can find the results here:

[ByteStreamsBenchmark w/ ThreadLocal](https://microbenchmarks.appspot.com/runs/087a68fc-af92-4971-97ae-346be17ee50b)
[ByteStreamsBenchmark w/out ThreadLocal](https://microbenchmarks.appspot.com/runs/dfb1c1b7-b9e2-431c-91f9-3be2f19db2a7)
[ByteSourceBenchmark w ThreadLocal](https://microbenchmarks.appspot.com/runs/f5517421-8f23-4237-9de1-f7d9e4c8f3ce)
[ByteSourceBenchmark w/out ThreadLocal](https://microbenchmarks.appspot.com/runs/573ac1ff-a1e9-4870-9b0d-414e77e163e6)

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,12 +9,38 @@ concurrency libraries, common annotations, string processing, I/O, and so forth.
 
 Requires JDK 1.6 or higher (as of 12.0).
 
+Latest release
+--------------
+
+The most recent release is [Guava 18.0][], released August 25, 2014.
+
+- [18.0 API Docs][Release API Docs]
+- [18.0 API Diffs from 17.0][Release API Diffs]
+
+To add a dependency on Guava using Maven, use the following:
+
+```xml
+<dependency>
+  <groupId>com.google.guava</groupId>
+  <artifactId>guava</artifactId>
+  <version>18.0</version>
+</dependency>
+```
+
+Snapshots
+---------
+
+Snapshots of Guava built from the `master` branch are available through Maven
+using version `19.0-SNAPSHOT`. API documentation and diffs from version 18.0
+are available here:
+
+- [Snapshot API Docs][]
+- [Snapshot API Diffs from 18.0][Snapshot API Diffs]
+
 Learn about Guava
 ------------------
 
-- Our users' guide, [Guava Explained](https://github.com/google/guava/wiki/Home)
-- Browse [API docs for the most recent release](http://google.github.io/guava/releases/18.0/api/docs/)
-- Browse [API diffs for the most recent release](http://google.github.io/guava/releases/18.0/api/diffs/)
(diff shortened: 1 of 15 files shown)


berndhopp:
I signed it!

PR #2112, opened 2015-07-19
Use thread-local buffers to avoid memory allocation and improve performance

berndhopp:
# Abstract

goal of this PR is to replace

```
byte[] buf = new byte[BUF_SIZE];
ByteArrayOutputStream baos = new ByteArrayOutputStream();
```

with

```
Buffers.Lease<byte[]> lease = Buffers.leaseByteArray();

try{
   byte[] buffer = lease.getBuffer();
} finally {
    lease.returnLease();
}

Buffers.Lease<ByteArrayOutputStream> lease = Buffers.leaseByteArrayOutputStream();

try{
   ByteArrayOutputStream baos = lease.getBuffer();
} finally {
    lease.returnLease();
}


```

in all cases where byte-array or ByteArrayOutputStreams are used method-internal, this way avoiding unneccessary memory allocation, increasing performance and enhancing the memory profile. 
# Implementation

[ThreadLocalBuffers](https://github.com/berndhopp/guava/blob/master/guava/src/com/google/common/io/
Buffers.java) is the manager of thread-local byte-arrays and FastByteArrayOutputStreams. It uses [SoftReferences](https://docs.oracle.com/javase/6/docs/api/java/lang/ref/SoftReference.html) or [WeakReferences](https://docs.oracle.com/javase/6/docs/api/java/lang/ref/WeakReference.html) to store buffers, this way enabling Garbage Collection to collect the buffers when the system runs low on memory. All allocations of byte[] or ByteArrayOutputStream in where the allocated byte[]/BAOS is not the returned value of the method have been replaced with calls to ThreadLocalBuffers. 
ByteSource.contentEquals was challenging, because it used not one but two byte-arrays for buffering, so I used two ByteBuffers that wrap around the upper and lower part of the thread-local buffer, thereby mimicking the two buffers.
# Results

The microbenchmarks show that avoiding buffer allocation yields a significant performance boost. Tests that did not show significant differences are left out. There have been no tests where performance dropped more than 2.00%. 

| tested method | bytes copied | runtime w/ TL buffers | runtime w/out TL buffers | performance increase in % |
| --- | --- | --- | --- | --- |
| ByteStreams.copy | 10 | 105569 | 681130 | 545,2 |
| ByteStreams.copy | 100 | 107411 | 698152 | 549,98 |
| ByteStreams.copy | 10000 | 840354 | 1186307 | 41,17 |
| ByteStreams.readBytes | 10 | 68917 | 633365 | 819,03 |
| ByteStreams.readBytes | 100 | 69802 | 660336 | 846,01 |
| ByteStreams.readBytes | 10000 | 360357 | 934588 | 159,35 |
| ByteStreams.toByteArray | 10 | 122681 | 750095 | 511,42 |
| ByteStreams.toByteArray | 100 | 131705 | 773561 | 487,34 |
| ByteStreams.toByteArray | 10000 | 1962961 | 4962929 | 152,83 |
| ByteSource.contentEquals | 10 | 647140 | 3614650 | 458,56 |
| ByteSource.contentEquals | 100 | 685559 | 647694 | 453,81 |
| ByteSource.contentEquals | 10000 | 5987317 | 9460941 | 58,02 |

you can find the results here:

[ByteStreamsBenchmark w/ ThreadLocal](https://microbenchmarks.appspot.com/runs/087a68fc-af92-4971-97ae-346be17ee50b#r:scenario.benchmarkSpec.methodName,scenario.benchmarkSpec.parameters.n)
[ByteStreamsBenchmark w/out ThreadLocal](https://microbenchmarks.appspot.com/runs/dfb1c1b7-b9e2-431c-91f9-3be2f19db2a7#r:scenario.benchmarkSpec.methodName,scenario.benchmarkSpec.parameters.n)
[ByteSourceBenchmark w ThreadLocal](https://microbenchmarks.appspot.com/runs/f5517421-8f23-4237-9de1-f7d9e4c8f3ce#r:scenario.benchmarkSpec.parameters.n)
[ByteSourceBenchmark w/out ThreadLocal](https://microbenchmarks.appspot.com/runs/573ac1ff-a1e9-4870-9b0d-414e77e163e6#r:scenario.benchmarkSpec.parameters.n)


cpovirk:
Benchmark links with the `SCENARIO.VMSPEC` values hidden:

[ByteStreamsBenchmark w/ ThreadLocal](https://microbenchmarks.appspot.com/runs/087a68fc-af92-4971-97ae-346be17ee50b#r:scenario.benchmarkSpec.methodName,scenario.benchmarkSpec.parameters.n)
[ByteStreamsBenchmark w/out ThreadLocal](https://microbenchmarks.appspot.com/runs/dfb1c1b7-b9e2-431c-91f9-3be2f19db2a7#r:scenario.benchmarkSpec.methodName,scenario.benchmarkSpec.parameters.n)
[ByteSourceBenchmark w ThreadLocal](https://microbenchmarks.appspot.com/runs/f5517421-8f23-4237-9de1-f7d9e4c8f3ce#r:scenario.benchmarkSpec.parameters.n)
[ByteSourceBenchmark w/out ThreadLocal](https://microbenchmarks.appspot.com/runs/573ac1ff-a1e9-4870-9b0d-414e77e163e6#r:scenario.benchmarkSpec.parameters.n)

berndhopp:
new links are in the description, thank you

lukesandberg:
I think both SoftReferences and ThreadLocals can create issues
1. SoftReferences can induce poor gc behavior where (iirc) it takes 2 full gc for a softreference to be collected.  I believe the general advice is "don't use them, ever"
2. ThreadLocals can cause issues in certain application containers to leak classloaders.  I believe it is possible to workaround... i don't remember how.

Since preallocated buffers are such a win, the better option is probably to modify these apis so that users can supply their own buffer (which they could manage with a ThreadLocal+SofReference or whatever is appropriate)

berndhopp:
Hi Luke, thank you for your comment. 

It is often advised not to use ThreadLocals in webapplications (like [here](http://niklasschlimm.blogspot.de/2012/04/threading-stories-threadlocal-in-web.html) ) since webapplications typically use thread-pools and so threads never actually finish, leaving objects threadlocal-attached to a thread to never be collected by gc. However in this case, since we only store Soft-/WeakReferences, garbage collection is able to free the buffer-memory (except for the tiny SoftReference-Object itself),  thus avoiding the problem alltogether. 

As for the SoftReference-issue, are you referring to http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6912889 ? To circumvent this issue, it would be enough to just replace SoftReference with WeakReference, since WeakReference does not cause these problems, but it would come with some performance penalty. Maybe a good solution would be to use WeakReferences only for all ByteArrayOutputStreams and byte-arrays larger than the initial size of 8k, since you need to fill the heap up almost completely with softreferenced memory to cause the problem, what is hardly possible with 8k per thread. 
Also, guava uses SoftReferences in other places too, like http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/cache/CacheBuilder.html#softValues().

Regards
Bernd

Maaartinus:
AFAIK, there's already such a leak in [`com.google.common.escape.Platform`](https://github.com/google/guava/blob/master/guava/src/com/google/common/escape/Platform.java#L40), which could be fixed trivially by avoiding `initialValue`.

berndhopp:
@Maaartinus exactly, I removed the overriding of initialValue [here](
https://github.com/berndhopp/guava/commit/11973b139fc66741d4392abf3082db1c39662fab#diff-26e4970e348a674eadbfcccaac4fdb43L27)

berndhopp:
@cpovirk, I wonder if you can give me a hand in telling me why the fix in https://github.com/berndhopp/guava/commit/36bc7be81cc48325a7ac52ad599c4e58e7a50af8 does not work. I could not find too many hints as how to setup client side emulation, so I do not know what to do here. Regards

berndhopp:
@lukesandberg this is open for quite a time now, are there any plans to include or questions, requirements?

berndhopp:
@cgdecker @kluever @Bezier89 @cgruber @lowasser @lukesandberg @kevinb9n 

I'm sorry to bother all of you, but I've putted a lot of effort into this PR and I presented some numbers to prove that it offers a tremendous performance improvement, so I wonder why it had been fallen asleep. Many critical functions work a multitude of times faster than before, so it should be worth considering. 

I personally do not think that there should be additional API's to let a user supply its own buffers. I consider it the obligation of the framework to handle its buffers, otherwise most of the users would either not use buffer re-use at all or end up programming their own thread-local buffer implementation. 

What are your thoughts, what is needed to get this merged? Do you have concerns about the implementation? Should there be more control for the user?

best
Bernd

15characterlimi:
I'm not on the Guava team, but some brief comments:

* major comment: I'm not sure that ThreadLocal<Reference>s are the right approach here. Perhaps there is a better solution for the case that you care about (e.g. passing a pre-allocated array like in Collection.toArray())?
 - the Reference object itself cannot be garbage collected so this will necessarily create at least two objects per calling thread. Those two objects will then stick around for the rest of the lifetime of the virtual machine. 
 - (minor) this creates a new class of possible bugs because the methods are not safe to call from one another (because the global variable that is the thread local could be overwritten by the other method). I guess the tests should catch that.
- like others said, relying on Soft/WeakReferences often has nonobvious implications for GC performance which I'm guessing will not be apparent in your benchmarks but still exist.

- the optionality of the zeroing feels like it risks potential bugs for a (I'm guessing) marginal performance benefit. At the very least, there should be two differently named methods rather than one with a boolean argument https://ariya.io/2011/08/hall-of-api-shame-boolean-trap

- the javadoc of the gwt version claims that the class uses WeakReference / SoftReference but it actually keeps a static final byte[] instead. (yuck).

- Exceptions are more helpful if they quote the offending value. E.g.
   checkArgument(minSize >= 0, "minSize must not be negative");
   checkArgument(minSize <= MAX_BYTE_SIZE, "minSize must not exceed 2^16");
should probably be:
  checkArgument(0 <= minSize && minSize <= MAX_BYTE_SIZE, "Expected 0 <= value < %s, got %s", MAX_BYTE_SIZE, minSize);

berndhopp:
Hi Tobias, thanks for your comment. I'll go through it one by one.

>major comment: I'm not sure that ThreadLocals are the right approach here. Perhaps there is a >better solution for the case that you care about (e.g. passing a pre-allocated array like in >Collection.toArray()? or something else)?

There is no particular case I especially care about, this is a patch that should increase performance in every scenario. As I said, I don't think that passing a pre-allocated array or a similar api would help users along as re-using buffers properly is a tricky task that should be handled by a framework. 

>the Reference object itself cannot be garbage collected so this will necessarily create at least >two objects per calling thread. Those two objects will then stick around for the rest of the >lifetime of the thread.

What is the second object? The Reference object itself will not be collected until the thread finishes. But this object itself only allocates a negligible amount of memory. 

>(minor) this creates a new class of possible bugs because the methods are not safe to call from >one another (because the global variable that is the thread local could be overwritten by the >other method). I guess the tests should catch that.

Here you've got a point, I'm working on a solution for that currently

>like others said, relying on Soft/WeakReferences often has nonobvious implications for GC >performance which I'm guessing will not be apparent in your benchmarks but still exist. Yes the >CacheBuilder class supports SoftReferences but only if the caller explicitly asks for them, not >silently under the hood.

I have not found much on the internet on these problems, besides http://bugs.java.com/bugdatabase/view_bug.do?bug_id=6912889 which I have already addressed

>the optionality of the zeroing feels like it risks potential bugs for a (I'm guessing) marginal >performance benefit. At the very least, there should be two differently named methods rather >than one with a boolean argument https://ariya.io/2011/08/hall-of-api-shame-boolean-trap

well, I don't think there are any bugs in the current code. But in case the ThreadLocalBuffers-class is to be made public one day it may be better to make zeroing mandatory right now. 

>the javadoc of the gwt version claims that the class uses WeakReference / SoftReference but it >actually keeps a static final byte[] instead. (yuck).

you're right, my bad

>Exceptions are more helpful if they quote the offending value. E.g.
>checkArgument(minSize >= 0, "minSize must not be negative");
>checkArgument(minSize <= MAX_BYTE_SIZE, "minSize must not exceed 2^16");
>should probably be:
>checkArgument(0 <= minSize && minSize <= MAX_BYTE_SIZE, "Expected 0 <= value < %s, got >%s", MAX_BYTE_SIZE, minSize);

agreed

I will push some more commits in the coming days to address these issues.

berndhopp:
Update: ThreadLocal buffers are already used in Guava https://github.com/google/guava/blob/master/guava/src/com/google/common/escape/Platform.java#L29

berndhopp:
@15characterlimi please see my last commit where I hopefully addressed most of the possible problems you mentioned. I'm sorry that this commit does not have all the documentation yet, especially for the new Lease-concept, but I think the code should be readable as is.

guava-gwt/src-super/com/google/common/io/super/com/google/common/io/Buffers.java
```diff
@@ -0,0 +1,96 @@
...
+
+        }
+
+        @Override
+        public void close() throws IOException {
+
+        }
+    };
+
+    //since javascript is effectively single threaded, one buffer is sufficient
```
jbduncan: Hmm, what's this reference to JavaScript about?

berndhopp: it's the GWT code, which is compiled to JS eventually

jbduncan: Ah, I see! Thanks. :smile:

berndhopp:
@15characterlimi ok, thanks for the update

berndhopp:
Hi @15characterlimi, any news here?

cgdecker:
Sorry for how long this has taken, @berndhopp.

I'm going to take a look at the details of this and try to have a concrete response for you soon. I will say that my initial inclination is that this isn't something we want to do in Guava: I think reuse of buffers is a bit of a fraught issue and one that should be explicitly under the user's control, which would really require a different API design than our libraries currently have. But again, I'll try to respond more specifically soon.

berndhopp:
@cgdecker thanks for your answer. I'm a bit puzzled right now what the concerns are. Buffer reuse is a tricky task where you can do a lot of things wrong, it took me quite a lot of effort to come up with an implementation that is safe even when the same thread uses a buffer on different levels of the stack trace. Also, there is not really any good alternative algorithm or approach to the problem; if you want buffer reuse to be efficient, save and free of memory leaks, you will inevitably wind up with ThreadLocals, Weak-/SoftReferences and some sort of 'Lease' concept to avoid reuse by the same thread. Feel free to disagree on these points I mentioned, I'm willing to discuss them. But my point is: if you have a patch that increases performance in every possible scenario without any api break or side effects whatsoever, why would a user need more 'control' over it? I just cannot think of any scenario where a user-supplied buffer would make more sense than my implementation. It may outperform my implementation by a small margin if the user would cache a buffer in say a HttpSession, but there is just too much that could do wrong. 

If your concerns are that it 'looks' somewhat fragile and hard to test properly, I can totally understand that, but that is just the nature of the beast in this case.

jbduncan:
One concern I have is I don't know if thread-local buffers are thread-safe when used with libraries that introduce radically different threading models to the JVM, like [Parallel Universe's Quasar](http://docs.paralleluniverse.co/quasar/) with it's so-called "fibers" (or lightweight threads), or [Akka](http://akka.io/) with it's actor model.

berndhopp:
Both Quasar and Akka rely on plain old threads eventually, so the concurrency models may be different but they base on the same building blocks. In http://docs.paralleluniverse.co/quasar/ is says: 

"Using ThreadLocals in a fiber works as you’d expect – the values are local to the fiber."

Because a fiber is nothing but a wrapped Thread at the end of the day. Every concurrency library out there is based on plain old threads and ThreadLocal cannot be "tricked" out of working, so there is nothing to worry about. As long as every code execution in java happens in the context of a java thread, ThreadLocals will work and this patch will work.

wikier:
To learn from this PR and understand the background, while is still open? The feature request looks like a good idea to me. I'd need to dive deep in the implementation to provide further feedback. But what I still don't understand why is not closed (as `merge` or simply `closed`), at least not open forever.

cgdecker:
Sorry for letting this fall off my radar. While I think buffer reuse _can_ be useful, I don't think the way Guava's I/O utilities are designed lends itself to this. Because `ByteStreams` and `CharStreams` are all static methods, there is no (good) way for the user to have control over caching behavior. I want to avoid mutable static state like the lease provider map in `Buffers`. And I generally don't like the idea of potentially holding on to a buffer for each (reused) thread that has ever used one, which could add up to quite a bit, without the user having control over that.

A different design for the package where, say, you instead have an _instance_ of your utility class and can pass in some kind of buffer cache when creating it (which could use thread locals or some kind of pool or nothing) would work better for something like this, but we most likely won't be making a change like that to common.io anytime soon, if ever.

These are my main issues with the idea of trying to do buffer reuse in Guava in general. I think that if we _were_ to do some kind of (non-user-configurable) buffer reuse, though, I'd want it to be something  considerably simpler than what's in this PR currently, and something that has a strict bound on the amount of memory it holds on to. For example, for many applications I'd bet that you could get a significant percent of the benefit by by just using an `AtomicReference<byte[]>` or a fixed size `BlockingQueue<byte[]>` and letting each place that needs a buffer try to take that one and just make a new buffer if it's not there. If you have lots of threads doing I/O using Guava at once (web server maybe) there's considerably less benefit, of course.

berndhopp:
hm, let me reply to this real quick. 

I'm a little confused here, as to what the demands actually are. It had been requested that the user has full control, so I introduced the Lease-concept which gives full control by allowing the user to register it's own LeaseProvider-implementation. Now you say if you where to introduce buffer-reuse, it will be non-user-configurable, should be considerably simpler than this PR and has a strict bound on the amount of memory it holds on to. My first approach, that had no Lease-concept, did match all these requirements in the first place. It did not introduce any additional API, it's implementation was considerably simpler and the actual arrays had been soft-referenced in order to let the garbage collection free them in case memory runs low. Letting GC decide when to fee memory is a better approach than limiting the memory-consumption by assumptions or heuristics. GC will always have the best insights on runtime memory needs. AtomicReferences or BlockingQueues won't save you here, because they either introduce idling (if there are not enough buffers for all threads that need one), or eventually you wind up reinventing the very same mechanisms that ThreadLocal<Reference<byte[]>> had to begin with. No benefit either way. 

If you where to eventually introduce buffers with the objectives you mentioned, chances are you will either introduce the approach I presented in the beginning, or one that is inferior to that. my 2 cents.

PR #2122, opened 2015-08-01
Ensure the correct removal notification is published (fixes #2101)

### LINKED ISSUE #2101: Wrong RemovalCause published on cache.invaldateAll()
When the cache is being cleared, either through `Cache.invalidateAll()` or `asMap().clear()`, the removal cause is always `EXPLICIT`. However the map may contain expired or GC'd entries that would be removed on the next maintenance cycle. The removal cause should indicate the actual state of the entry, not merely the action that triggered the eventual removal.

Due to weak and soft references the key or value may be null. A listener should only expect to receive a null key or value if the cause is `COLLECTED`, as indicated by `RemovalNotification`:
>The key and/or value may be null if they were already garbage collected.

Therefore a listener might decide to inspect the properties without a null guard, leading to an NPE.

```java
@Test
public void clear_expireAfterAccess() {
  FakeTicker ticker = new FakeTicker();
  RemovalListener<Integer, Integer> listener = Mockito.mock(RemovalListener.class);
  ArgumentCaptor<RemovalNotification<Integer, Integer>> captor =
      ArgumentCaptor.forClass(RemovalNotification.class);
  Cache<Integer, Integer> cache = CacheBuilder.newBuilder()
      .expireAfterAccess(1, TimeUnit.MINUTES)
      .removalListener(listener)
      .ticker(ticker)
      .build();
  cache.put(1, 1);
  ticker.advance(10, TimeUnit.MINUTES);
  cache.invalidateAll();

  verify(listener).onRemoval(captor.capture());
  assertThat(captor.getValue().getCause(), is(RemovalCause.EXPIRED));
}
```

The fix is to change `LocalCache$Segment#clear()` to inspect the key, value, and expiration status when determining the cause. 

Unfortunately this race condition is more widespread due to `enqueueNotification` not being tolerant to a garbage collection nulling out the key/value. Prior to constructing the notification, the method must take a strong reference to the key and value, check if either was collected, and if so publish a `COLLECTED` instead of the caller's cause.

ben-manes:
Ensures that the correct removal notification is published by `Segment#clear()`. Previously the operation was using the cause `EXPLICIT` when pending maintenance for expiration or collection would take precedence.

For additional safety, publishing notifications require passing a strong reference of both the key and value. This ensures that compiler reordering and races with the GC do not result in a reference collection and publishing a null value. This may not have been a problem due to happens-before ordering across methods, but simplifies away having to make these assumptions.

In addition to adding an expiration test, verified by removing the workaround in Caffeine's fixture. The tests pass with the changes and fail with 19-rc1.

### DIFF
--- a/guava-tests/test/com/google/common/cache/CacheExpirationTest.java
+++ b/guava-tests/test/com/google/common/cache/CacheExpirationTest.java
@@ -22,6 +22,7 @@ import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 import com.google.common.cache.TestingCacheLoaders.IdentityLoader;
 import com.google.common.cache.TestingRemovalListeners.CountingRemovalListener;
+import com.google.common.cache.TestingRemovalListeners.QueuingRemovalListener;
 import com.google.common.collect.Iterators;
 import com.google.common.testing.FakeTicker;
 import com.google.common.util.concurrent.Callables;
@@ -31,6 +32,7 @@ import junit.framework.TestCase;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
 /**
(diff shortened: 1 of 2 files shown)


cpovirk:
Merged. Thanks, Ben.

PR #2129, opened 2015-08-11
Do something about RegularContiguousSet.hashCode() performance?

rcpoison:
Iterating over the whole ContiguousSet seems wrong.
RegularContiguousSet.equals() does not use Sets.equalsImpl() either but instead compares the DiscreteDomain and first+last.

### DIFF
--- a/guava/src/com/google/common/collect/RegularContiguousSet.java
+++ b/guava/src/com/google/common/collect/RegularContiguousSet.java
@@ -188,10 +188,12 @@ final class RegularContiguousSet<C extends Comparable> extends ContiguousSet<C> 
     return super.equals(object);
   }
 
-  // copied to make sure not to use the GWT-emulated version
   @Override
   public int hashCode() {
-    return Sets.hashCodeImpl(this);
+    int hash = this.domain.hashCode();
+    hash = 89 * hash + this.first().hashCode();
+    hash = 89 * hash + this.last().hashCode();
+    return hash;
   }
 
   @GwtIncompatible("serialization")


lowasser:
This change violates the spec of Set.hashCode().

> RegularContiguousSet.equals() does not use Sets.equalsImpl() either but instead compares the DiscreteDomain and first+last.

It falls back to super.equals(otherSet), which does use Sets.equalsImpl.

rcpoison:
@lowasser: Of course you're right :/ Sorry, should've thought harder.

kevinb9n:
This does bring up the fact that many DiscreteDomain implementations
*could* provide
an efficient hashCode() specialized for their type.

e.g. in Integers, Range.closed(0, n).asSet().hashCode() == n * (n + 1) / 2,
and all that.  Even if the implementation still had to loop it could still
avoid all the damned boxing.

This would be fun programming but... not sure whether
ContiguousSet.hashCode() is called enough to be worth worrying about or not.


On Tue, Aug 11, 2015 at 10:00 AM, rcpoison <notifications@github.com> wrote:

> @lowasser <https://github.com/lowasser>: Of course you're right :/ Sorry,
> should've thought harder.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/google/guava/pull/2129#issuecomment-129968386>.
>



-- 
Kevin Bourrillion | Java Librarian | Google, Inc. | kevinb@google.com

rcpoison:
> not sure whether
> ContiguousSet.hashCode() is called enough to be worth worrying about or not
It was in my case, and the next silly idea was to cache the hashCode. I've opted for storing the Ranges instead, which makes a lot more sense anyway.

kevinb9n:
Not sure I would close this yet. Given how awful what this method does really is, caching the result might be a very good idea.

cpovirk:

Issue #1070, opened 2014-10-31
Factory method to create a FluentIterable from an array


gissuebot:
_[Original issue](https://code.google.com/p/guava-libraries/issues/detail?id=1070) created by **drothmaler** on 2012-07-17 at 03:04 PM_

---

I think it would be nice to be able to create a FluentIterable from an array directly, with a factory method like this:
FluentIterable.from(E... elements) or FluentIterable.from(E[] elements)

This could simply call Arrays.asList(elements) internally, as I am doing now explicitly on the outside.
It is very usefull for working with legacy code or other APIs returning arrays.
Also it feels to be consistent, as there is a "toArray" method; so it would be a logical consequence to provide a from(Array) factory method also.

scr:
Thanks for linking this to my PR, @cpovirk.  I see some opinions of varargs, but not spelled out with reasoning.  

Why was it preferred to add `of(E[])` over `of(E...)` ? for methods under test that take an Iterable<E>, why wouldn't you want to just call the method with `FluentIterble.of(foo, bar, baz)`, which is much closer to the type that you want (Iterable)?  

Yes, I get it there's always that method out there somewhere that will do the same thing - i.e. `Arrays.asList(foo, bar, baz)` also implements Iterable, but it's a mental hoop to remember it, or understand why it was used as a reviewer.  

Certainly that mental-tax, however small, is worth not having to pay and choosing `FluentIterble.of(foo, bar, baz)` would be instantly intuitively obvious to everyone who sees it and stick in their minds more easily the next time they need to make an Iterable quickly for testing…

scr:
I should point out that it was done this way for `Iterators` and I don't see why not also `FluentIterable`.

http://docs.guava-libraries.googlecode.com/git/javadoc/com/google/common/collect/Iterators.html#forArray(T...)

stephenh:
Somewhat frustrating this hasn't been added. @scr FWIW I've found https://github.com/jOOQ/jOOL to be pretty nice, and generally either a) have the methods/overloads I expect, or b) the maintainer is pretty open about adding methods/overloads as requested. FWIW/YMMV.

Unfortunately GWT doesn't have java.util.stream support yet, so I'm using old-school Guava transform/etc. with lambdas until I can just use jOOL for both client- and server-side.

cpovirk:
Reopening and retitling this to reflect where the discussion has gone.

cpovirk:
No, wait. We already have an issue for varargs: #2136.

PR #2136, opened 2015-08-19
Improve FluentIterable.of to take variable # of arguments.

scr:
`FluentIterable.of(1, 2, 3)` would be easier to use/more succinct than `FluentIterable.of(new int[]{1, 2, 3})`

This change does not break the existing API as E... becomes E[] by the time it is used in the method.

### DIFF
--- a/guava/src/com/google/common/collect/FluentIterable.java
+++ b/guava/src/com/google/common/collect/FluentIterable.java
@@ -165,7 +165,7 @@ public abstract class FluentIterable<E> implements Iterable<E> {
    */
   @Beta
   @CheckReturnValue
-  public static <E> FluentIterable<E> of(E[] elements) {
+  public static <E> FluentIterable<E> of(E... elements) {
     return from(Lists.newArrayList(elements));
   }
 


scr:
I signed the CLA!

cpovirk:
See also past discussion on https://github.com/google/guava/issues/1070#issuecomment-61348831

scr:
@stephenh - I saw your comment on https://github.com/google/guava/issues/1070 but can't reply any longer.

FWIW, I rolled my own and have been getting good traction with it…

See https://github.com/scr/j8iterables for some useful crossovers between Guava, Stream, and Jdk8!

cgdecker:
We actually already have methods for this internally (`of()` and `of(E, E...)`) and are planning to release them soon.

billccn:
This change, however, did not come in Guava 19.

I am wondering what's the solution for the overloading confusion that would be introduced by such a signature change when there's one single array argument? (A potential break in source back-compatibility.)

The other Guava APIs avoided this by naming the var-arg method differently from the one that accepts an array (`of` vs. `copyOf`), but the original signature has obviously broken this naming convention.

PR #2148, opened 2015-09-03
byte array searches with start and end limits

kvr000:
This extends current indexOf function with possibility of specifying where to start and where to end with the search, similarly to java String.indexOf methods.


kvr000:
I signed it!

lowasser:
Hi, can you discuss more details about your use case for this method?

kvr000:
Hi, sure, I'm working on project with consumes ASCII input (potentially binary) and basically need similar functions for parsing as String provides.

lowasser:
Do you actually need an end parameter?  I observe String.indexOf does not have that.

kvr000:
If the data comes from structure I don't have control over - like ByteBuffer from NIO or ByteBuf from netty, this parameter will become pretty handy.

PR #2159, opened 2015-09-18
Update details of upcoming release

dpursehouse:
The README refers to 19.0-RC1, but 19.0-rc2 is now available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ Upcoming release
 ----------------
 
 [Guava 19.0][] is the next release. A release candidate is currently available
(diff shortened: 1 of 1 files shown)


cgdecker:
Thanks for the PR... I've updated this internally along with another change to the README and it should be mirrored out soon.

dpursehouse:
@cgdecker thanks, I can see it's updated now.  Do you have any idea when the final release for 19.0 is coming?  In the Gerrit project we're using 19.0-rc2 and it would be nice if we can upgrade to the final version before we cut the first RC for 2.12.

PR #2161, opened 2015-09-18
[Fixes #2160] Make it possible to halt the execution of expensive transforms on comparation objects via onResultOf variants.

### LINKED ISSUE #2160: ComparisonChain should support not performing expensive transforms: provide onResultOf variants
Consider the following snippet - even though comparison would be avoided beyond the first compare call, the parameter evaluation for the second one (`this.expensiveCalculation()`) would still be done.

```
ComparisonChain.start()
    .compare(this.lessThanThat(), that.greaterThanThis())
    .compare(this.expensiveCalculation(), that.expensiveCalculation())
    .result();
```

I propose an extension similar to `Ordering::onResultOf` that can be used to avoid operations that are not needed because the comparison chain is "inactive" (result has already been determined).

```
ComparisonChain.start()
    .compare(this.lessThanThat(), that.greaterThanThis())
    .compare(this, that, ThisClass::expensiveCalculation)
    .result();
```

scr:


cpovirk:
Closing now that we've closed the associated issue based on https://github.com/google/guava/issues/2160#issuecomment-1260074693

PR #2162, opened 2015-09-19
Jdk5 backport master

ashu-22:
Creating  a project with the help of Guava.


cgdecker:
Not sure what this is for, but we definitely don't want to merge the jdk5-backport branch into master.

PR #2163, opened 2015-09-19
Added ifNonNull to MoreObjects

paplorinc:
If input is non-null, applies the function to it.
If either the input or the result is null, will return the default value.

This is an alternative for `Optional`, similarly to `firstNonNull`:
```Java
Optional.fromNullable(person.getName()).transform(String::toUpperCase).or("");
```
or
```Java
String name = person.getName();
if (name != null)
  name = name.toUpperCase();
if (name == null)
  name = "";
```
can now be written as (Java 8, default value and static import):
```Java
ifNonNull(person.getName(), String::toUpperCase, "");
```

### DIFF
--- a/guava/src/com/google/common/base/MoreObjects.java
+++ b/guava/src/com/google/common/base/MoreObjects.java
@@ -16,14 +16,12 @@
 
 package com.google.common.base;
 
+import com.google.common.annotations.*;
+
+import javax.annotation.*;
+import java.util.*;
+
 import static com.google.common.base.Preconditions.checkNotNull;
-
-import com.google.common.annotations.GwtCompatible;
-
-import java.util.Arrays;
-
-import javax.annotation.CheckReturnValue;
-import javax.annotation.Nullable;
 
 /**
  * Helper functions that operate on any {@code Object}, and are not already provided in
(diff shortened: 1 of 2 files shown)


paplorinc:
Signed ;) 


     On Saturday, 19 September 2015, 12:55, googlebot <notifications@github.com> wrote:
   

 Thanks for your pull request. It looks like this may be your first contribution to a Google open source project, in which case you'll need to sign a Contributor License Agreement (CLA). Please visit https://cla.developers.google.com/ to sign.Once you've signed, please reply here (e.g. I signed it!) and we'll verify. Thanks.   
   - If you've already signed a CLA, it's possible we don't have your GitHub username or you're using a different email address. Check your existing CLA data and verify that your email is set on your git commits.
   - If you signed the CLA as a corporation, please let us know the company's name.

dant3:
How is it different from `Optional` approach besides not creating an `Optional` instance?
```java
Optional.fromNullable(name).transform(n -> n.toUpperCase()).or("");
```

paplorinc:
Hello @dant3,

The point of this modification is to provide an alternative for `Optional` for another very common case.
The same as for Guava's `firstNonNull`, e.g. instead of
```Java
    Optional.fromNullable(first).or(Optional.of(second)).get();
```
we can *choose* (this is the keyword here) to simply call
```Java
    firstNonNull(first, second);    
```
Which delimits the different parts of the computation way better, than `Optional` could.


The same applies to this changeset also, e.g. instead of
```Java
    Optional.fromNullable(name).transform(n -> n.toUpperCase()).or(-1);
```
we can *choose* to simply call
```Java
    ifNonNull(name, n -> n.toUpperCase(), -1);
```
Which keeps the bare essentials also.

ps1. I need the second Optional wrapping in the first example in order to have the same behavior as the alternative function does, for all-null parameters
ps2. I statically imported `MoreObject`, because its methods are stateless and have meaning without the surrounding object, unlike `Optional`, where simply calling `of(second)` would look very unintuitive.

paplorinc:
(related to https://github.com/google/guava/pull/2164)

zsolt-donca:
I like the general idea presented here, as there definitely are use cases when the above-defined `ifNonNull` would be the preferable choice.

What is the expected behavior when `thenFunction` returns null? Should `ifNonNull` return 
- null (the current behavior), 
- throw an exception, like the `Optional.transform` does in the above example
- `elseValue`, like the equivalent of the above example would do if it was using `java.util.Optional` from Java 8.

guava/src/com/google/common/base/MoreObjects.java
```diff
@@ -57,6 +56,43 @@
...
+   * }</pre>
+   *
+   * @param input        The value that should be null-checked
+   * @param thenFunction The function to apply to the non-null input
+   * @param elseValue    The value that should be used in case the input was null
+   * @return if input is null, return the default value, otherwise apply a function to it
+   * @since 19.0
+   */
+  @Nullable
+  public static <T, R> R ifNonNull(@Nullable T input, @Nonnull Function<T, R> thenFunction, @Nullable R elseValue) {
```
zsolt-donca: Another overload with a `Supplier` for the `elseValue` could be useful, in order to make lazy evaluation possible.

lowasser:
I'm not sure we're ready to accept new features that are dependent on Java 8, in the sense that if you're working with Java 7, this is much less readable than the "conventional" code you would write with a perfectly normal ternary or if statement.

paplorinc:
@zsolt-donca, very good observations, thanks, will apply them.
@lowasser, I understand your concern, and would like to add a few comments:
* Guava has many methods that look horrible with Java 7, e.g.
```Java
Integer find(List<Integer> inputs) {
  return Iterables.find(inputs, new Predicate<Integer>() {
    @Override
    public boolean apply(Integer input) {
      return input == 2;
    }
  });
}
```
but can be replaced with a simple constructs, e.g.
```Java
Integer find(List<Integer> inputs) {
  for (Integer input : inputs)
    if (input == 2) return input;  
  throw new NoSuchElementException();
}
```
or
```Java
Iterables.find(inputs, Predicates.equalTo(2))
```
The same is true in this case, i.e. one could still extract commonly used functions and use them in a similar manner.
```Java
name = ifNonNull(person.getName(), Strings.toUpperCase(), "");
```

* The other argument is a more subjective one. Idea already presents these cases as if it were lambda style, making it more terse in reality than the variable&ternary style, i.e. instead of:
```Java
String name = person.getName();
if (name != null)
  name = name.toUpperCase();
if (name == null)
  name = "";
```
we can still see in the IDE
```Java
name = ifNonNull(person.getName(), (n) → { return n.toUpperCase(); }, "");
```

* And as a closing argument, in my personal opinion, Guava could really complement Java 8's shortcomings in more ways than it does now.
Not providing useful functionality because in a deprecated Java version the syntax is uglier, is a pitty.

cgdecker:
Personally, I feel that `Optional` is a much better way to approach this. `firstNonNull` preceded the existence of our `Optional` class and has the advantage of being very simple to understand. I think when you get into something like this `ifNotNull(T, Function, R)` it becomes much less obvious to readers what you're doing, whereas `Optional.fromNullable(value).transform(function).or(fallback)` is pretty clear.

Thanks for the idea, but I'm fairly certain this isn't an API we want to add, so I'm going to close this. For future reference, if you want to see an API added or changed it's better to start by just opening an issue for discussion of it. Pull requests are better for changes that don't affect the APIs themselves or changes that we've decided should be made and that we'd like a contribution for.

paplorinc:
Thanks for the explanation :)

paplorinc:
@cgdecker, could you please take a look at https://github.com/google/guava/pull/2164 also?

PR #2164, opened 2015-09-19
Added a vararg overload for firstNonNull

paplorinc:
Avoiding nullability in Java is a recurring problem.

Other languages offer an `Optional`, which makes nullability composable.
This however is very verbose in Java and cannot consistently be inserted into an existing, nullable codebase (especially since the `Optional` can be null also).

Guava tries to simplify some of the common cases, e.g. we have a value, but if it's not set, we want a default (but the same can be true for the default).
Unfortunately this method wasn't called `defaultIfNull`, but `firstNonNull`, therefore people assumed it's a search.

This commit tries to solve that by _making_ it a search, and letting the users decide which approach they want:
ternary,

``` Java
(first != null) ? first : ((second != null) ? second : checkNotNull(third));
```

Find,

``` Java
Iterables.find(asList(first, second, third), Predicates.notNull());
```

Optional,

``` Java
Optional.fromNullable(first).or(second).or(third).get();
```

or the same approach Guava already provided:

``` Java
firstNonNull(first, second, third);
```

### DIFF
--- a/guava-gwt/test-super/com/google/common/base/super/com/google/common/base/MoreObjectsTest.java
+++ b/guava-gwt/test-super/com/google/common/base/super/com/google/common/base/MoreObjectsTest.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright (C) 2006 The Guava Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.common.base;
+
+import com.google.common.annotations.*;
+import junit.framework.*;
+
+import static com.google.common.base.MoreObjects.firstNonNull;
+
+/**
+ * Tests for {@link MoreObjects}.
+ */
+@GwtCompatible(emulated = true)
+public class MoreObjectsTest extends TestCase {
+  public void testFirstNonNull_withNonNull() throws Exception {
+    String s1 = "foo";
+    String s2 = firstNonNull(s1, "bar");
+    assertSame(s1, s2);
+
+    Long n1 = (long) 42;
+    Long n2 = firstNonNull(null, n1);
+    assertSame(n1, n2);
+  }
+
+  public void testFirstNonNull_throwsNullPointerException() throws Exception {
+    try {
+      firstNonNull(null, null);
+      fail("expected NullPointerException");
+    } catch (NullPointerException ignored) {
+    }
+  }
+
+  public void testFirstNonNull_varArg() throws Exception {
+    Integer input1 = 1;
+    assertEquals(input1, firstNonNull(input1, null, null));
+
+    Character input2 = '2';
+    assertEquals(input2, firstNonNull(null, input2, null));
+
+    Double input3 = 3.0;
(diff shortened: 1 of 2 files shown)


paplorinc:
I signed it as `paplorinc`! :) 


     On Saturday, 19 September 2015, 12:58, googlebot <notifications@github.com> wrote:
   

 Thanks for your pull request. It looks like this may be your first contribution to a Google open source project, in which case you'll need to sign a Contributor License Agreement (CLA). Please visit https://cla.developers.google.com/ to sign.Once you've signed, please reply here (e.g. I signed it!) and we'll verify. Thanks.   
   - If you've already signed a CLA, it's possible we don't have your GitHub username or you're using a different email address. Check your existing CLA data and verify that your email is set on your git commits.
   - If you signed the CLA as a corporation, please let us know the company's name.

kevinb9n:
See issue 384

paplorinc:
Thanks, updated the commit message and posted a response to the Issue :)

guava/src/com/google/common/base/MoreObjects.java
```diff
@@ -39,21 +38,34 @@
...
    */
   @CheckReturnValue
   public static <T> T firstNonNull(@Nullable T first, @Nullable T second) {
-    return first != null ? first : checkNotNull(second);
+    return (first != null) ? first
+                           : checkNotNull(second);
+  }
+
+  @CheckReturnValue
+  public static <T> T firstNonNull(@Nullable T first, @Nullable T second, @Nullable T third, @Nullable T... rest) {
```
zsolt-donca: I would add javadoc to this method as well, otherwise there is no documentation visible in IDEs at the _call site_.

guava/src/com/google/common/base/MoreObjects.java
```diff
@@ -39,21 +38,34 @@
...
    * first.or(supplier)}.
    *
-   * @return {@code first} if it is non-null; otherwise {@code second} if it is non-null
-   * @throws NullPointerException if both {@code first} and {@code second} are null
+   * @return the first non-null parameter
+   * @throws NullPointerException if all parameters are null
    * @since 18.0 (since 3.0 as {@code Objects.firstNonNull()}).
    */
   @CheckReturnValue
   public static <T> T firstNonNull(@Nullable T first, @Nullable T second) {
```
zsolt-donca: I would keep the original javadoc here (referring to the parameters as _first_ and _second_); this is a simpler and more specific case of the below method, then the javadoc could be simpler and more specific as well.

zsolt-donca: I would keep the original javadoc here (referring to the parameters as _first_ and _second_); this is a simpler and more specific case of the below method, then the javadoc could be simpler and more specific as well.

guava-gwt/test-super/com/google/common/base/super/com/google/common/base/MoreObjectsTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+    Character input2 = '2';
+    assertEquals(input2, MoreObjects.firstNonNull(null, input2, null));
+
+    Double input3 = 3.0;
+    assertEquals(input3, MoreObjects.firstNonNull(null, null, input3, null));
+    
+    String input4 = "4";
+    assertEquals(input4, MoreObjects.firstNonNull(null, null, null, input4));
+
```
zsolt-donca: The cases when there are nulls elements in the array should also be tested.

zsolt-donca:
I strongly support this change, as I've been in the same pitfall: when I've first seen the method `firstNonNull` in use, I've immediately assumed that it accepts any number of arguments (as it would be logical, given it's name and intent); and then I was disappointed when I tried to use it with three arguments.
The existing alternatives are far less elegant and concise, and also fragment the code base (while for the two-parameter cases I'm using `firstNonNull`, for the three-parameter cases I'm using something different).

paplorinc:
@zsolt-donca, awesome comments, will apply them :)

mmontag:
What became of this one? Continued in a different PR?

PR #2174, opened 2015-09-29
ImmutableMap.copyOf(Map): minor performance improvement

rgoldberg:
Minor performance improvement in `ImmutableMap.copyOf(Map)`:

If `map instanceof ImmutableSortedMap`, no longer unnecessarily run the following, since it will always be `false`:

```
else if (map instanceof EnumMap)
```


rgoldberg:
This change also clarifies the intent of the method

cpovirk:
This is minor enough that I think it's not worth even a couple extra lines, sorry. If you know of a place in which this is a bottleneck and can provide a benchmark, we could reconsider.

PR #2180, opened 2015-10-08
Add tryParse methods to UnsignedBytes, UnsignedInts, and UnsignedLongs.

leeyikjiun:


leeyikjiun:
Okay, I will make the changes. Do I have to close this pull request and create a new one?

lowasser:
I don't think so?  I think you can just add more changes to this pull request?

leeyikjiun:
I can't add the tryParse method for UnsignedBytes as there is no UnsignedByte class to return as a null object.

lowasser:
Then probably leave it out -- it's not clear what we should do there.

leeyikjiun:
I've found some further inconsistencies. In UnsignedLongs.parseUnsignedLong, a NumberFormatException is thrown if the radix is not within the min or max radix which is consistent with Java's Long.parseLong. However, in Longs.tryParse, an IllegalArgumentException is thrown instead.

I think the new tryParse methods should throw NumberFormatException as well. But what should we do with the old methods?

PR #2183, opened 2015-10-10
Release19

hugo-vrijswijk:


cgdecker:
We don't need to merge the release branch into master.

Issue #2075, opened 2015-06-12
Consider phasing out our use of interruptTask() in favor of done()


cpovirk:
(This part of of our work to figure out the best way to do cancellation logic -- related to https://github.com/google/guava/issues/1905)

Without `interruptTask()`, we could just write:

```
done() {
  if (wasInterrupted()) {
    whatever();
  }
  anyOtherCleanup();
}
```

On the one hand, I don't like making users write a conditional when we could provide a callback method. On the other hand, when there are multiple callbacks, there are questions about the order in which they run. Users wonder whether their fields might have been nulled out. Users have the opportunity to rely on poorly specified behavior like whether listeners have been executed or `get()` callers have been released.

Of course, right now, out-of-package users don't have the ability to override `done()`, so we won't eliminate `interruptTask()` yet if at all. (And it's non-`@Beta`, so we'd be waiting a while to eliminate it.) But it's worth doing some investigation here, in part because we'll have a decision to make for `FluentFuture` someday.

Issue #1958, opened 2015-01-29
Do we need an onCancelled method for setFuture(cancelledFuture) to invoke?


cpovirk:
(Not a high priority, just something I realized recently.)

`myFuture.setFuture(cancelledFuture);`

This causes `myFuture` to be cancelled. But under the current `AbstractFuture` implementation, the cancellation doesn't invoke `myFuture.cancel`.

That seems fine on the face of it, but it got me thinking: Don't we rely on `myFuture.cancel` for cleanup in some cases? Investigating, I found that the only classes that both call `setFuture` and override `cancel` are those in `Futures`. (Hopefully I got my search right. The results sound sane enough.) Here's my analysis:

- `FallbackFuture`: `cancel` merely cancels the original `Future`, which is unnecessary in all the cases that we call `setFuture` (namely, cases in which the original `Future` is already done and, specifically, has failed)
- `TimeoutFuture`: `cancel` (a) cancels the input `Future` and (b) cancels the timeout timer. But no worries: We don't call `setFuture` until the input `Future` is done, and at that point, we don't need to do (a), and we also take care of doing (b) manually.
- `ChainingListenableFuture`: `cancel` cancels only the *original* `Future`. If we've called `setFuture`, though, that means that the original `Future` is already done, so the cancellation would be a no-op.

What does this mean for adding `onCancel`? Mainly, it doesn't appear to be necessary at present: Things already seem to work right. The most I see is an opportunity for code reuse. But I also see that many of our cancel/done methods call many of our other cancel/done methods, and any change has the chance to introduce infinite loops, unexpected nulling out of fields, and other problems. Our tests might catch these (we hope), but I point it out also to say that the change here likely wouldn't be trivial.

cpovirk:
Forgot to add: Through the same `TrustedFuture` path, it's possible for an `AbstractFuture` to transition to an "interrupted" state without invoking `interruptTask`. That is probably even *less* likely to be a problem.

If we're concerned, though, one possibility is to invoke `interruptTask` conditionally from `complete`. We could do the same with the hypothetical `onCancelled`.

Another possibility is to re-expose `done`. Since we added `wasInterrupted` a while back, the implementer of that method has the ability to determine what kind of completion occurred. This would let us provide a hook for cleanup+cancellation+interruption all in one, avoiding questions about which is called first... except that `interruptTask` and the overrideable `cancel` aren't going away anytime soon -- or perhaps we'd omit them in a forthcoming `AbstractFluentFuture`?

[...]

Issue #1905, opened 2014-12-05
AbstractFuture.done() hook


cpovirk:
We have `interruptTask()` for the interrupt case, but we seem to have a number of `AbstractFuture` subclasses where we need to do work on a non-interrupting cancellation (usually propagating that cancellation to a related `Future` or clearing some resources that would have been used during execution). Currently we override `cancel` to call `super.cancel`, but I hate to encourage overriding of `Future` methods, which are too easy to get wrong, and the necessary boilerplate is a minor annoyance.

The main difficulty now is that `interruptTask()` already exists. That leaves us with some unappealing options:
- Add `onCancel(boolean)` to be called on `cancel(anything)` (but before or after `interruptTask()`?)
- Add `cancelSuccessfulButInterruptionNotRequested()` to be called only on `cancel(false)`.
- Add `onCancel(boolean)` with a default behavior of calling `interruptTask` if interruption was requested.

(One of the `onCancel` approaches is probably what the most users want (certainly users inside Guava itself), but it either raises questions about the order of operations (what if `onCancel` nulls out something that `interruptTask` needs? Is that plausible? Can reversing the order save us?) or makes `interruptTask` meaningless if someone overrides `onCancel()`.)

cpovirk:
See also the justification in #1958, which I have closed in favor of this issue.

cpovirk:
A couple key points that aren't emphasized clearly on this thread:

- `setFuture(futureThatEndsUpCancelled)` does not invoke `cancel` *if* `futureThatEndsUpCancelled` is a `TrustedFuture`. If `cancel` is supposed to do important cleanup, it won't get done. But this is probably mostly OK. First, if we called `setFuture`, we probably don't have any real work left to do: The work is being done by the other `Future`. Still, users may be surprised that `cancel` isn't invoked, especially because this happens conditionally based on the type of input `Future` (so tests might not catch it). Second, most real cleanup in our `Future` implementations happens in `done`. But we don't expose `done` outside the package. (On the occasion that it comes up, we advise that users add a listener. But we've found that we don't like this ourselves: First, there's the extra allocation (perhaps avoidable in some cases if you already have an object that can implement `Runnable`). Then, I think we may also have had concerns about initialization order? I forget. Maybe we should expose `done` again? There's some discussion related to this on https://github.com/google/guava/issues/760

[...]

PR #2184, opened 2015-10-10
allow inheriting from SettableFuture and overriding cancel() method

kvr000:
I'd like to be able to implement custom ListenableFuture class where I could easily override cancel(boolean) method. The SettableFuture seems to be best fit but unfortunately it's declared final. Removing this restriction makes it much more reusable.

### DIFF
--- a/guava/src/com/google/common/util/concurrent/SettableFuture.java
+++ b/guava/src/com/google/common/util/concurrent/SettableFuture.java
@@ -30,7 +30,7 @@ import javax.annotation.Nullable;
  * @since 9.0 (in 1.0 as {@code ValueFuture})
  */
 @GwtCompatible
-public final class SettableFuture<V> extends AbstractFuture.TrustedFuture<V> {
+public class SettableFuture<V> extends AbstractFuture.TrustedFuture<V> {
 
   /**
    * Creates a new {@code SettableFuture} in the default state.
@@ -43,7 +43,7 @@ public final class SettableFuture<V> extends AbstractFuture.TrustedFuture<V> {
    * Explicit private constructor, use the {@link #create} factory method to
    * create instances of {@code SettableFuture}.
    */
-  private SettableFuture() {}
+  protected SettableFuture() {}
 
   @Override public boolean set(@Nullable V value) {
     return super.set(value);


cpovirk:
The class you're looking for is `AbstractFuture`. I will add some doc to `SettableFuture` to point there. (We currently have doc on `ListenableFuture`, so I may just add a pointer back there.)

Another thing we need to do is figure out our recommendations for extending `AbstractFuture`. Our hope is that users will rarely have to override `cancel()` (which doesn't always get invoked on cancellation, surprisingly enough: https://github.com/google/guava/issues/1958 https://github.com/google/guava/issues/1905). In many cases, all you need is to override `interruptTask()`. For cases in which that is not enough, we are probably going to expose a `done()` hook for users to override (https://github.com/google/guava/issues/2075).

I'll leave this request open until I at least add the doc.

kvr000:
Thanks for mentioning `interruptTask()`, this is definitely better than overriding full `cancel()` method.

However, I still need both the `cancel()` and `set()` / `setException()` method. I actually think that this is pretty common use case for asynchronous tasks that are finished not by returning from method but after some notification is delivered (e.g. another `ListentableFuture`, completion callback etc.). In such case I need future for which it's possible to set the result and custom method for interrupting the task.

I believe that SettableFuture is best candidate here. Is there any drawback by making it non-final?

cpovirk:
`cancel()`, `set()`, and `setException()` are all available on `AbstractFuture`, too. (I haven't formed a full picture of your requirements, but I'm confident that anything that can be done with `SettableFuture` can be done with `AbstractFuture`.)

But I guess I have been dodging the question :) A reason for `SettableFuture` to be final is that it gives us freedom to perform [optimizations](https://github.com/google/guava/blob/60fe0bb1bf3257779474af8fd80a59cb9a2f402a/guava/src/com/google/common/util/concurrent/AbstractFuture.java#L701). Well, OK, for that particular optimization, we need only `get()` to be `final`, not `cancel()`. But we want to leave ourselves the freedom to do additional optimizations in the future. In fact, this `get()`-based optimization is new for Guava 19, so if we had permitted overriding of arbitrary methods, we would have prevented ourselves from introducing it. We also have the freedom to make other changes besides optimizations. (I don't remember if this has been important for `SettableFuture` yet, but that's the general principle at work. And again, for people who do need to extend the type, we provide `AbstractFuture`, so we're just pushing users to take a slightly different approach.)

kvr000:
Well, I see the point that with current design it can save some execution time in rather special situations though I can't judge how common they are in guava or the software based on this.

However, as you mentioned - this is mainly about optimizing `get()`, not `cancel()`... I think marking `get()` as final in `TrustedFuture` would make it safer, more readable and clear on purpose and on the other hand wouldn't affect extendability of `TrustedFuture` and as a result the inherited classes would take advantage of those performance optimizations as well :-)

Personally I agree it's not a big deal to inherit it from `AbstractFuture` in exactly the same way how `SettableFuture` is currently implemented. But I don't like I have to reinvent the wheel / duplicate the code. So maybe we can consider introducing another intermediate class `AbstractSettableFuture` and final `SettableFuture`. Although as described above it probably would hardly bring any value from the originally proposed approach.

Additionally, I'm considering implementing a `Future` class where in case of code is already running and cancellation is requested, `get()` and listener processing would be called after `set()`/`setException()` methods are called instead of immediately after invoking `cancel()`. Do you think it would be worth merging into guava or shall I skip the step for this? :-)

cpovirk:
I've added some documentation to `SettableFuture`.

When you refer to "duplicating the code," are you talking about duplicating the overrides of `set` and `setException` that exist only to make them accessible? Often this ends up being unnecessary if you call the methods from within your `AbstractFuture` subclass.

To your `cancel()` request, I think you're describing a `Future` that diverges from the `Future` spec. [That spec](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/Future.html#cancel-boolean-) requires that the `Future` be done when `cancel()` returns.

kvr000:
Yes, I'm talking about duplicating `set` and `setException`. If I was able to inherit from `SettableFuture` it would save me few common lines of code. I need to call them externally as in this case it's basically used implementing integration pattern between 3rd-party framework and guava based code. I.e. (simplified version) `3rdparty.method().addListener(() -> { guavaFuture.set(3rdpartyFuture.get()); }`

For the `cancel()` request - I don't think it's somehow in conflict with specification. The specification only says that `isDone()` must return true and `isCancelled()` true as well but doesn't say when. And actually as the future is handled asynchronously by listener anyway it doesn't make any difference when you really "finish" the future.

cpovirk:
If the need to redeclare `set` and `setException` is distasteful, you can put your `3rdparty.method().addListener` call inside your `AbstractFuture` subclass, probably declared inside your `MyThirdPartyUsingService` class. This might turn out to be the simplest place to define your subclass, anyway, to keep the `set` and `cancel` logic near one another.

RE: "subsequent calls to `isDone()` will always return true": I think that the natural reading of "always" is "at any time," not "after some arbitrary future time." Otherwise, all the statement would be saying is "Once the `Future` is done, it's always done," and that doesn't seem like something to document on `cancel()`.

kvr000:
That would be even worse because I would lose reusability of such class. And secondly I understand the original purpose of `SettableFuture` as the creating code doesn't yet know who and how updates the future. Otherwise it could be rather called Forwarded which in this case is just the listener.

RE: "subsequent calls to isDone()will always return true": But it still does. The only difference is that listeners invocation is postponed until the task is finished (if it already runs). The state of future would be always CANCELLED since the completion of `cancel()`and all the affected methods would return proper value.

cpovirk:
You seem to be suggesting that the `isDone` concept of "completed" ("Returns `true` if this task completed") has no connection to the concept of "completed" used elsewhere in the `Future` docs ("The result can only be retrieved using method `get` when the computation has completed").

Returning to `SettableFuture` and `final`: I regret that I haven't formed a clear enough picture of how `SettableFuture` would be used in your code to be more helpful. As far as the feature request goes, though, the Guava team is satisfied with `AbstractFuture` as a solution here.

kvr000:
Never mind. I can either copy the setters or base the code on spring SettableListenableFuture which is more open in these terms.

Thanks anyway.

PR #2185, opened 2015-10-12
Remove old method from ImmutableList

Vladlis:


cpovirk:
`readObject` is a method automatically invoked during deserialization, so it's still "used."

But this is a weird case: Because `ImmutableList` has a serialization proxy, `readObject` should never be called. The only case in which it can be called is the case in which a malicious user imitates a serialized `ImmutableList` that doesn't use the proxy. In that case, `readObject` serves to block this attempt, keeping the attacker from thwarting `ImmutableList`'s immutability guarantee. (This is covered in Effective Java, item 78 in the second edition.)

All that said, we decided a while ago that we weren't going to try to keep our `Immutable*` types immutable even in the presence of malicious users, so we can remove this if we want.

Vladlis:
@cpovirk, thanks for the explanation!

PR #2190, opened 2015-10-15
Fast size() impl for FileBackedOutputStream.asByteSource

tcataldo:
FileBackendOutput holds data in either a ByteArrayOutputStream or a File. Both of them provides quick access to content size. Use that knowledge to provide fast sizeIfKnown & size implementations.

### DIFF
--- a/guava/src/com/google/common/io/FileBackedOutputStream.java
+++ b/guava/src/com/google/common/io/FileBackedOutputStream.java
@@ -18,6 +18,7 @@ package com.google.common.io;
 
 import com.google.common.annotations.Beta;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Optional;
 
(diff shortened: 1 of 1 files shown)


tcataldo:
I signed it !

tcataldo:
I signed it with my other email address too.

jbduncan:
`size()` can be optimised to not instantiate an `Optional` internally each time it's called. See the code below for an example of what I mean by this.

```java
@Override
public long size() throws IOException {
  return file != null ? file.length() : memory.getCount();
}

@Override
public Optional<Long> sizeIfKnown() {
  return Optional.of(size());
}
```

jbduncan:
Hey @tcataldo, I'm not entirely clear on why you closed this PR. It looked rather promising to me.

tcataldo:
Because I will not have time to rebase it & apply review comments quickly.

PR #2192, opened 2015-10-15
add EMPTH in Strings

Qunzer:
When we define variable,like that, String foo = "";
we can use that, String foo = Strings.EMPTH;


kluever:
-1 on this PR. It's fewer characters, and no difference (performance-wise) to just write `""` instead of `EMPTH`

PR #2198, opened 2015-10-21
Added SuppressWarnings annotation to a used member which is updated using reflection

kilaka:
Added SuppressWarnings annotation to a used member which is updated using reflection


kilaka:
I signed it!

kilaka:
Closed because I commited using an unknown user :(

PR #2200, opened 2015-10-22
Added SuppressWarnings annotation to a used member which is updated using reflection

kilaka:
Added SuppressWarnings annotation to a used member which is updated using reflection.


cpovirk:
This is good information to put in the source code. Probably we should have a `@UsedFromReflection` annotation for just this purpose. In fact, we do, but it's internal only... and there are at least two :( ... I've started a discussion about what we should do about that.

kilaka:
I indeed couldn't find the `@UsedFromReflection` annotation.
Found almost nothing when searching for "UsedFromReflection discussion" using google's search.
@cpovirk, can you share the discussion link? Thanks.

cpovirk:
Sorry, I should have said that it was a discussion inside Google (since that's where the teams that currently use the annotation are). I'll let you know how it goes.

PR #2208, opened 2015-10-27
Made it possible to provide a custom StatsCounterSupplier in CacheBuilder

jebl01:
To simplify the process of collecting cache stats using e.g. _io.dropwizard.metrics_, `CacheBuilder.recordStats()` should have an overloaded version that accepts a StatsCounter Supplier.

> Just read the _guidelines for contributing_ and realized that I should have opened an issue for this instead of going directly to a PR. I do that, but I keep this PR as a reference...

### DIFF
--- a/guava/src/com/google/common/cache/CacheBuilder.java
+++ b/guava/src/com/google/common/cache/CacheBuilder.java
@@ -29,7 +29,6 @@ import com.google.common.base.Supplier;
 import com.google.common.base.Suppliers;
 import com.google.common.base.Ticker;
 import com.google.common.cache.AbstractCache.SimpleStatsCounter;
-import com.google.common.cache.AbstractCache.StatsCounter;
 import com.google.common.cache.LocalCache.Strength;
 
 import java.lang.ref.SoftReference;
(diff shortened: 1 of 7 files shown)


jebl01:
I signed it!

guava/src/com/google/common/cache/CacheBuilder.java
```diff
@@ -760,8 +759,28 @@ Ticker getTicker(boolean recordsTime) {
...
 
+  /**
+   * Enable the accumulation of {@link CacheStats} during the operation of the cache. Without this
+   * {@link Cache#stats} will return zero for all statistics. Note that recording stats requires
+   * bookkeeping to be performed with each operation, and thus imposes a performance penalty on
+   * cache operation.
+   *
+   * @param statsCounter a custom stats counter.
+   */
+  public CacheBuilder<K, V> recordStats(final StatsCounter statsCounter) {
```
ben-manes: This has to accept a Supplier, as the builder is reusable and a counter is generally per-cache. That's not horrible at the call site due to`Suppliers.ofInstance` or `CustomCounter::new`

guava/src/com/google/common/cache/StatsCounter.java
```diff
@@ -0,0 +1,82 @@
...
+import com.google.common.annotations.GwtCompatible;
+
+/**
+ * Accumulates statistics during the operation of a {@link Cache} for presentation by {@link
+ * Cache#stats}. This is solely intended for consumption by {@code Cache} implementors.
+ *
+ * @since 10.0
+ */
+@GwtCompatible
+public interface StatsCounter {
```
ben-manes: This could be argued as a breaking API change

ben-manes: It would probably need to extend the original, with that marked as deprecated. Then it could go through the graceful removal process.

ben-manes: Actually just have the original deprecated and extend this one would be the right way

cpovirk:
As you can see, we're mostly not making changes to `common.cache` anymore. We recommend Caffeine, which I believe [offers this functionality](https://static.javadoc.io/com.github.ben-manes.caffeine/caffeine/2.8.0/com/github/benmanes/caffeine/cache/Caffeine.html#recordStats-java.util.function.Supplier-).

jebl01:
👍

PR #2210, opened 2015-10-27
Add Futures.getUnchecked(Future<V> future, long timeout, TimeUnit unit)

sebgymn:
Time limited waiting is performed frequently on `Future` and this method overloads `Futures.getUnchecked()` method with one that accepts timeout value.

### DIFF
--- a/guava/src/com/google/common/util/concurrent/Futures.java
+++ b/guava/src/com/google/common/util/concurrent/Futures.java
@@ -1452,6 +1452,53 @@ public final class Futures extends GwtFuturesCatchingSpecialization {
     }
   }
 
+  /**
+   * Returns the result of calling {@link Future#get(long, TimeUnit)} uninterruptibly on a task
+   * known not to throw a checked exception. This makes {@code Future} more suitable for lightweight,
+   * fast-running tasks that, barring bugs in the code, will not fail. This gives it exception-handling
+   * behavior similar to that of {@code ForkJoinTask.join}.
+   *
+   * <p>Exceptions from {@code Future.get} are treated as follows:
+   * <ul>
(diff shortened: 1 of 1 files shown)


sebgymn:
I signed it!

Vinidel:
I signed it!

cpovirk:
We may end up doing this. The reasons that I haven't just merged it are:

- Our theory is that few operations are (a) slow and yet (b) unable to fail. Most are either predictably fast (in which case you don't need the timeout) or able to fail (in which case you need the checked exception -- say, for an RPC).
- Users who do have slow yet unable to fail operations may be better served by the full ForkJoin framework, which... hmm, I thought that it provided timed gets, but it seems not, at least not on `ForkJoinTask`? Maybe it's defined somewhere else?
- In most cases, it's hard to see a timeout as a "programming error." (Now, we do already have `UncheckedTimeoutException` for `TimeLimiter`, as you've noted... but that's an API that we've felt kind of icky about since we created it.)
- It's not 100% clear what to do with interruption. I think that what you do is probably right. But if the wait is expected to sometimes be long, then waiting uninterruptibly may not be ideal for all users.
- I have some general regret around the `getChecked`/`getUnchecked` family of methods. This comes mainly from the fact that `getChecked` is poorly suited for Android: It's likely to be slow (especially in the failure case but even somewhat slow in the success case on old versions), and more importantly, it doesn't play nice with Proguard. But how does that tie in to `getUnchecked`? Well, if people can't use `getChecked`, they're going to be tempted to use `getUnchecked`, even if it's not the right thing for them. So there's a part of me that wishes that we hadn't added either method.

Again, we still might do this. It just requires some weighing of pros and cons. Having the code waiting for us is nice, at least :)

PR #2211, opened 2015-10-28
Fix typo in Subscriber's Javadoc

mkordas:


mkordas:
I signed it!

lowasser:
I don't know whether it was the original intent, but "sporting" isn't necessarily wrong here:

https://en.wiktionary.org/wiki/sport#Verb

(transitive) To display; to have as a notable feature. 
Jen's sporting a new pair of shoes;  he was sporting a new wound from the combat

mkordas:
Thanks for explanation. I was looking in dictionary under 'sporting' and I haven't found anything. Apparently under 'sport' there is something that may fit.

cgdecker:
FYI, while we think "sporting" was intentional, it's not a very clear word to use, so we are changing this to use the word "with" instead.

Issue #1835, opened 2014-10-31
ImmutableSortedSet with 1 non-Comparable element


gissuebot:
_[Original issue](https://code.google.com/p/guava-libraries/issues/detail?id=1835) created by **tadamson** on 2014-08-19 at 11:20 PM_

---

Creating a singleton immutable sorted set by calling of() with a non-Comparable object is not allowed (via "FauxverideShim"). However, one can be created by first instantiating a single element Collection and creating a copyOf() it.

Example (throws Exception):
&nbsp;&nbsp;Object notComparable = new Object();
&nbsp;&nbsp;ImmutableSortedSet.of(notComparable);

Example (works):
&nbsp;&nbsp;Object notComparable = new Object();
&nbsp;&nbsp;ImmutableSortedSet.copyOf(ImmutableSet.of(notComparable));

I submit that they should both work or both fail. While technically a singleton non-comparable SortedSet violates the SortedSet contract ("All elements inserted into a sorted set must implement the Comparable interface"), it make senses to me that an immutable SortedSet of 1 element could be an exception (and thus not throw an Exception). In fact, TreeSet does not fail with a single non-Comparable element:

&nbsp;&nbsp;SortedSet&lt;Object> sortedSet = new TreeSet&lt;Object>();
&nbsp;&nbsp;sortedSet.add(notComparable);

cjosw:
I have submitted a pull request to fix this.

cpovirk:
We tried to do this last year, but we found that it would break some internal users. That means two things:
- We need to fix those users.
- It's probably going to break external users, too.
It might still be a good idea to fix, but things are no longer as simple as hoped :(

PR #2243, opened 2015-12-04
ImmutableSortedSet.copyOf should fail if input is not Comparable

cjosw:
ImmutableSortedSet.copyOf(Collection) should fail if the input collection is not comparable.
This currently applies for most collections, but fails if the input is a singleton.
This is recorded in issue #1835.

I have noted that this is caught within TreeSet by invoking the comparator on the singleton item, so this change makes a similar call to catch the un-comparable items in a singleton ImmutableSortedSet.

I have added tests for a two-element input collection (which was already caught), and a new test for a single-element input; this latter test only passes once the code change has been inserted.

### DIFF
--- a/guava-tests/test/com/google/common/collect/ImmutableSortedSetTest.java
+++ b/guava-tests/test/com/google/common/collect/ImmutableSortedSetTest.java
@@ -665,6 +665,22 @@ public class ImmutableSortedSetTest extends AbstractImmutableSetTest {
     assertSame(STRING_LENGTH, set.comparator());
   }
 
+  public void testCopyOf_immutableSingletonSetNotComparable_throwsClassCastException() {
+    try {
+      ImmutableSortedSet.copyOf(ImmutableSet.of(new Object()));
+      fail("Should complain that the element isn't Comparable");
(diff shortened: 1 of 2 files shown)


cpovirk:
Thanks. The trick will be seeing if any existing Google code fails and needs to be fixed. I have a feeling that we tried this before and had problems, but I'm hoping for the best. I'll let you know how it goes.

cpovirk:
Actually, I went and found the CL from last year. It does break people. This includes some people who define Comparators that throw if given equal elements! That's terrible, but it's also hard to prioritize fixing. I've teased the project to a couple people without success.

Sorry for not putting this information in the bug report. I'll do so now.

cjosw:
*This includes some people who define Comparators that throw if given equal elements!*

Note (just for info) that this is exactly what TreeSet does when you insert its first element.
The comparator call is what triggers the failure here:

     SortedSet<Object> sortedSet = new TreeSet<Object>();
     sortedSet.add(notComparable);

cpovirk:
Closing, since we need to do cleanup internally to take this :(

Issue #2083, opened 2015-06-17
Empty immutable collections should be explicitly documented as singletons


eamonnmcmanus:
`ImmutableList.of()` says this:

> Returns the empty immutable list. This set behaves and performs comparably to `Collections.emptyList()`, and is preferable mainly for consistency and maintainability of your code.

The use of the phrase "the empty immutable list" suggests that this method always returns the same object, as indeed it does. But this is implied only by the use of the definite article and the reference to `Collecitons.emptyList()`. It would be better to state it more explicitly. By comparison, `Collections.emptyList()` itself says this:

> Implementation note: Implementations of this method need not create a separate `List` object for each call. Using this method is likely to have comparable cost to using the like-named field. (Unlike this method, the field does not provide type safety.)

Of course the same applies to the other empty-collection methods.

kevinb9n:
Yes, this seems worth noting. We use `<b>Performance note:</b> ...` for things like this.

cjosw:
I have raised a pull request for this: https://github.com/google/guava/pull/2247

bedla:
Question: Why is ImmutableList.of() preferable to Collections.emptyList() ? JavaDoc: "for consistency and maintainability of your code" does not clarify it, imo.

Stephan202:
The return type of `ImmutableList.of()` is an `ImmutableList`. Guava recommends using such `Immutable*` types throughout, because this way immutability is communicated and propagated at the level of the type system. Not only does this increase maintainability (because it makes it easier to reason about the code), the compiler will then also warn about code that invokes a mutator method on such types. (Because they're marked deprecated.)

bedla:
I understand what you mean, but it works only when you have ImmutableList at signatures, but when you prefer coding against interface, you have List there...

ronshapiro:
We treat ImmutableList as an interface in that sense. It has meaningful
semantics that are frequently relevant for a caller to know. In the same
way that you can return a Collection vs. a Set, even though Set doesn't
have any different methods, it is clearly a different type.

bedla:
Ok, i see :] thanks a lot

cpovirk:
[...]

PR #2247, opened 2015-12-07
Update documentation to clarify that empty immutable collections may be singletons

cjosw:
This updates the documentation (no functional changes), to clarify that empty immutable collections may be singletons, as raised in a bug: https://github.com/google/guava/issues/2083

### DIFF
--- a/guava/src/com/google/common/collect/ImmutableBiMap.java
+++ b/guava/src/com/google/common/collect/ImmutableBiMap.java
@@ -35,6 +35,11 @@ public abstract class ImmutableBiMap<K, V> extends ImmutableMap<K, V> implements
 
   /**
    * Returns the empty bimap.
+   *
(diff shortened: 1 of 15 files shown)


nglorioso:
Hi there! Sorry it's taken a while to get to reviewing this PR.

Looking over the changes, I have a couple of primary notes:
- The "implementations of this method" language doesn't seem to make sense for static methods that have one implementation (while they follow the notes on Collections.emptyList(), etc.). I might suggest:

```
Performance notes: This method will always return the same instance of {@link FooType}.
```

- If we continue along this line, please update the linked data type to the Immutable version (e.g.: `ImmutableList` as opposed to `List`). We prefer to keep the Immutable type as the interface (see https://github.com/google/guava/issues/2083#issuecomment-388354744)

PR #2248, opened 2015-12-09
modify testCheckArgument_simpleMessage_failure to clean code

intonormal:
verifySimpleMessage(expected)  change to assertThat(expected).hasMessage("A message"), 

so that the  compared target("A message") can be saw straightway


intonormal:
I signed it!
this is my first contribute in open source
exciting and hope to be accepted

cpovirk:
Sorry that we sat on this (and so many pull requests) for so long.

If "A message" appeared inline in the setup for the test, I would agree that there's a clear benefit to including it inline in the assertion. But it's already defined elsewhere, so this test already fails to be readable on its own :( You could probably fine a way to move all the information inline if you want, but I'd encourage you to look for something more substantial. Maybe https://github.com/google/guava/issues/2394 if you're still interested.

PR #2344, opened 2015-12-28
Remove unaccessible and unnecessary legacy XML_ESCAPER.

leeca:
During some analysis with DepAn, I notices that XML_ESCAPER is never used internally, nor is it accessible externally.  Seems like a good idea to delete this cruft, especially after 2 years as dead code.

### DIFF
--- a/guava/src/com/google/common/xml/XmlEscapers.java
+++ b/guava/src/com/google/common/xml/XmlEscapers.java
@@ -113,7 +113,6 @@ public class XmlEscapers {
     return XML_ATTRIBUTE_ESCAPER;
   }
 
-  private static final Escaper XML_ESCAPER;
   private static final Escaper XML_CONTENT_ESCAPER;
   private static final Escaper XML_ATTRIBUTE_ESCAPER;
   static {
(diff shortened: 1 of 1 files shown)


leeca:
I signed it!

kluever:
So this is actually some deprecated code that we have internally, and is publicly accessible for our internal users (the method is deprecated).

Usually we're able to strip out that stuff, but since it's being constructed in a static block, it's a little more difficult. Since it's not hurting anything, I'm inclined to leave it as-is.

leeca:
You have write permission, so it is your call.  But it does seem to violate the normal Googley convention that no dead or commented out code should be checked into the repository.  I guess all rules have exceptions.

lowasser:
To be clear, the code is used internally, and we mirror out our internal codebase by passing it through a stripping tool, which has more trouble stripping out this code than usual because of the way it's written.  Internally, we're _extremely_ aggressive about stripping out unused code -- much more aggressive than we can be externally -- but this happens to hit an edge case where we can't strip out something that we've eliminated externally.

PR #2345, opened 2015-12-29
Add Splitter methods to return CharSequence types.

mspiegel:
Not sure if this has been discussed previously but it is trivial to add methods that return CharSequence types to the Splitter API and maintain the existing API. We have some use cases for this feature where we do not want to create temporary char[] array copies of the intermediate substrings.

The Splitter API accepts CharSequence types as input and internally manipulates CharSequence types. This patch maintains the existing methods that return String types and adds additional methods that
return CharSequence types. The performance overhead on the existing methods that return String type is one additional call to Iterators#transform() that calls toString() on the CharSequence results. Note the #toString() call was already present in the existing implementation.

### DIFF
--- a/guava-tests/test/com/google/common/base/SplitterTest.java
+++ b/guava-tests/test/com/google/common/base/SplitterTest.java
@@ -66,6 +66,12 @@ public class SplitterTest extends TestCase {
     assertThat(letters).containsExactly("a", "b", "c").inOrder();
   }
 
+  public void testCharacterSimpleSplitToCharSequenceList() {
+    String simple = "a,b,c";
+    List<CharSequence> letters = COMMA_SPLITTER.splitToCharSequenceList(simple);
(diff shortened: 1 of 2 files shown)


kluever:
The `Splitter` API has been part of Guava since 1.0, and has been fairly stable (it's non-`@Beta` at this point). I'm somewhat doubtful that this use-case is popular enough for us to start expanding the API at this point.

mspiegel:
It looks like the API of Splitter was created at the time when the HotSpot
JVM did not allocate new char[] arrays for subsequences. That behavior was
changed in their release of JDK 7u6:
http://bugs.java.com/view_bug.do?bug_id=4513622. Might this warrant the
need for new methods that do not create new char[] arrays?

kluever:
That's an interesting issue, but I think I'm still -1 on this. Any thoughts @lowasser ?

kluever:
Unlikely to add this w/o a demonstrated, proven performance issue.

Issue #2132, opened 2015-08-13
Add video MediaTypes


Alexuy:
Missed several video MimeTypes in MediaType class:

    video/x-flv
    video/3gpp
    video/3gpp2

and some audio types:

    audio/basic
    audio/L24
    audio/aac
    audio/vorbis
    audio/x-ms-wma
    audio/x-ms-wax
    audio/vnd.rn-realaudio
    audio/vnd.wave

kluever:
I started on this, but the `audio/L24` one is interesting because our MediaType class requires that all subtypes be lowercased. I didn't dig too much into it, but I suspect we do this for normalization (e.g., "image/gif" and "image/GIF" are treated as equal).

Maybe we shouldn't do this though? I'm not really sure. Any thoughts @gk5885 ?

kluever:
From our own javadocs:
> All portions of the media type that are case-insensitive (type, subtype, parameter attributes) are normalized to lowercase.

So I wonder if we can just define it as:
`createConstant(AUDIO_TYPE, "l24");`

kluever:
And @cpovirk points out:

> Both top-level type and subtype names are case-insensitive.

via https://tools.ietf.org/html/rfc6838#section-4.2

Alexuy:
I found some other missed types:

image

    image/pjpeg: Progressive JPEG
    image/vnd.wap.wbmp: WBMP

text

    text/cmd: commands
    text/php: PHP scripts

message

    message/http (RFC 2616)
    message/imdn+xml: IMDN (RFC 5438)
    message/partial: E-mail (RFC 2045 and RFC 2046)
    message/rfc822: E-mail; EML files, MIME files, MHT files, MHTML files (RFC 2045 and RFC 2046)

model

For 3D models

    model/example: (RFC 4735)
    model/iges: IGS files, IGES files (RFC 2077)
    model/mesh: MSH files, MESH files (RFC 2077), SILO files
    model/vrml: WRL files, VRML files (RFC 2077)
    model/x3d+binary: X3DB files
    model/x3d+vrml: X3DV VRML files
    model/x3d+xml: X3D XML files

x-pkcs (https://en.wikipedia.org/wiki/PKCS)

    application/x-pkcs12: p12 files
    application/x-pkcs12: pfx files
    application/x-pkcs7-certificates: p7b files (https://tools.ietf.org/html/rfc2315.html#section-9.1)
    application/x-pkcs7-certificates: spc files
[...]

PR #2354, opened 2016-01-07
Add new mediatypes according to #2132

steven-maasch:
Add new mediatypes from [#2132](https://github.com/google/guava/issues/2132).

The mediatype for php (text/php) and cmd (text/cmd) are not included,
because these are not specified by IANA.


ILapitan:
яр
07.01.2016 19:43 пользователь "Steven Maasch" <notifications@github.com>
написал:

kluever:
I have a CL for this internally as well (CL/102147951)...we had a question about L24 vs. l24, but I think we decided just to go with l24.

kluever:
Submitted internally...it should be mirrored out shortly.

Also for these MediaType additions, the hard work isn't adding the constants, but rather writing the javadocs, and digging up the proper citations for the javadocs.

PR #2377, opened 2016-01-30
Feature/tables.synchronized table 2156

mostlymagic:
Had deleted previous fork by accident.
Suggested changes implemented in this one.

### DIFF
--- a/guava-gwt/src-super/com/google/common/collect/super/com/google/common/collect/Synchronized.java
+++ b/guava-gwt/src-super/com/google/common/collect/super/com/google/common/collect/Synchronized.java
@@ -16,8 +16,6 @@
 
 package com.google.common.collect;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-
 import com.google.common.annotations.GwtCompatible;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
@@ -29,7 +27,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
-import java.util.Map.Entry;
 import java.util.Queue;
 import java.util.RandomAccess;
 import java.util.Set;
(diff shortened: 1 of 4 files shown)


guava-gwt/src-super/com/google/common/collect/super/com/google/common/collect/Synchronized.java
```diff
@@ -1303,5 +1303,192 @@ public E remove() {
...
+      synchronized (mutex) {
+        return set(delegate().columnKeySet(), mutex);
+      }
+    }
+
+    @Override
+    public Map<C, Map<R, V>> columnMap() {
+      synchronized (mutex) {
+        Function<Map<R, V>, Map<R, V>> mapFunction = synchronizedMapWrapper(mutex);
+          Map<C, Map<R, V>> columnMap = delegate().columnMap();
```
lowasser: Wait, I wasn't referring to indenting these; they should be at the same level as the rest of the code...

guava-tests/test/com/google/common/collect/SynchronizedTableTest.java
```diff
@@ -0,0 +1,329 @@
...
+        @Override
+        public int hashCode() {
+            assertTrue(Thread.holdsLock(mutex));
+            return super.hashCode();
+        }
+
+        private static long serialVersionUID = 0;
+    }
+
+  /*
```
lowasser: I think it's this comment that should be..indented +2?  The leading / of the comment should be at the same level as the method signatures.

mostlymagic: seems to have been an IDE settings mixup. will provide a fix, sorry

shapowang:
Does anyone resolve the conflict and merge the code?
I cannot find the `Tables.synchronizedTable` function in guava 20.0-rc1.This function would be useful,please.

cpovirk:
I'm not sure why we ended up with our own implementation instead of this one, but we ended up adding one in https://github.com/google/guava/commit/e3f9ba6a707b03559910bceffb74244ac9a8db13

Issue #1409, opened 2014-10-31
Add more uninterruptible methods


gissuebot:
_[Original issue](https://code.google.com/p/guava-libraries/issues/detail?id=1409) created by **Ash2kk** on 2013-05-12 at 10:33 AM_

---

Currently Uninterruptibles have putUninterruptibly() and takeUninterruptibly() methods for BlockingQueue. Consider adding helper methods for other interruptible methods: offer(Object, long, TimeUnit) and poll(long, TimeUnit).

Also there is BlockingDeque interface with many more interruptible methods.

Related to issue #﻿1212 and issue #﻿1315.

RamAnvesh:
Also there are the Object.wait methods. I am guessing they are also used quite frequently.

a86c6f7964:
is there a reason that poll or offer has not been added? do they introduce something that will make people program incorrectly? such as https://github.com/google/guava/issues/1118 ?

cpovirk:
I think we just haven't needed them yet. They seem fine to add as far as I can tell.

thecoop:
Could this be made more generic, for running any arbitrary operation that throws `InterruptedException`?

    Uninterruptibles.runUninterruptibly(() -> opThrowingInterruptedException())

RamAnvesh:
@thecoop https://github.com/google/guava/issues/2384

leventov:
Maybe there are enough voices to add more methods now? We would benefit from BlockingQueue.offer() and poll() too.

Pr0methean:
@thecoop For time-limited operations, we'd need to pass the total time into the helper method, which in turn would pass the *remaining* time as an argument to the functional parameter.

tanmauec:
@cpovirk l am beginner when it comes to contributing to open-source of Guava. I would like to work on this issue.

I am planning to implement two methods offer(long timeout, TimeUnit unit) and poll(long timeout, TimeUnit unit).

Please let me know if I can start working on this!

Thanks

cpovirk:
That sounds great. Go for it!

(I have some other things that I need to get back to people about, so I might not review it immediately, but I'll be sure to get to it.)

tanmauec:
Ack.

tanmauec:
@cpovirk Please review per your convenience.

tanmauec:
All errors fixed, Please review [4034](https://github.com/google/guava/pull/4034)

tanmauec:
[...]

PR #2383, opened 2016-02-07
In Uninterruptibles, added waitUninterruptibly methods which call Object#wait methods

RamAnvesh:
https://github.com/google/guava/issues/1409

### DIFF
--- a/guava/src/com/google/common/util/concurrent/Uninterruptibles.java
+++ b/guava/src/com/google/common/util/concurrent/Uninterruptibles.java
@@ -326,7 +326,67 @@ public final class Uninterruptibles {
     }
   }
 
-  // TODO(user): Add support for waitUninterruptibly.
+  /**
+   * Invokes {@code object.}{@link Object#wait() wait()} uninterruptibly.
+   *
+   * @param object to wait upon.
+   *
+   * @throws IllegalMonitorStateException if the current thread is not
+   *               the owner of the object's monitor.
+   */
+  @GwtIncompatible // concurrency
+  public static void waitUninterruptibly(Object object) {
+    Preconditions.checkNotNull(object);
+    boolean interrupted = false;
+    try {
+      while (true) {
+        try {
+          object.wait();
+          return;
+        } catch (InterruptedException e) {
+          interrupted = true;
+        }
+      }
+    } finally {
+      if (interrupted) {
+        Thread.currentThread().interrupt();
+      }
+    }
+  }
+
+  /**
+   * Invokes {@code object.}{@link Object#wait(long, int) wait(millis, nanos)} uninterruptibly.
+   *
+   * @param object to wait upon.
+   *
+   * @throws IllegalMonitorStateException if the current thread is not the owner of the object's
+   *          monitor.
+   */
+  @GwtIncompatible // concurrency
+  public static void waitUninterruptibly(Object object, long timeout, TimeUnit unit) {
+    Preconditions.checkNotNull(object);
+    Preconditions.checkNotNull(unit);
+    boolean interrupted = false;
+    try {
+      long remainingNanos = unit.toNanos(timeout);
+      long end = System.nanoTime() + remainingNanos;
+
+      while (true) {
+        try {
+          // timedWait treats negative timeouts just like zero.
+          NANOSECONDS.timedWait(object, remainingNanos);
+          return;
+        } catch (InterruptedException e) {
+          interrupted = true;
+          remainingNanos = end - System.nanoTime();
+        }
+      }
+    } finally {
+      if (interrupted) {
+        Thread.currentThread().interrupt();
+      }
+    }
+  }
 
   private Uninterruptibles() {}
 }
--- a/guava-tests/test/com/google/common/util/concurrent/UninterruptiblesTest.java
+++ b/guava-tests/test/com/google/common/util/concurrent/UninterruptiblesTest.java
@@ -19,6 +19,7 @@ package com.google.common.util.concurrent;
 import static com.google.common.util.concurrent.InterruptionUtil.repeatedlyInterruptTestThread;
 import static com.google.common.util.concurrent.Uninterruptibles.joinUninterruptibly;
 import static com.google.common.util.concurrent.Uninterruptibles.putUninterruptibly;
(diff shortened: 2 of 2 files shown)


RamAnvesh:
I signed it!

cpovirk:
Sorry for the delay on this. Now that `UninterruptiblesTest` is released, can you add some tests for your new methods, patterned off the existing tests? Then I'll take a look.

Ram-Kasam:
Sure. will do...

RamAnvesh:
Sure. will do...

RamAnvesh:
Hi,
Sorry about the delay. I have added the 11 tests for the following cases:

- Not holding monitor throws IllegalMonitorStateException. 
- wait, waitSuccessfullyWithTimeout, waitUnseccessfullyWithTimeout for the following cases each:

1. No interrupt
2. Single Interrupt
3. Multi Interrupt

guava/src/com/google/common/util/concurrent/Uninterruptibles.java
```diff
@@ -326,7 +326,67 @@ public static boolean tryAcquireUninterruptibly(
...
+  public static void waitUninterruptibly(Object object) {
+    Preconditions.checkNotNull(object);
+    boolean interrupted = false;
+    try {
+      while (true) {
+        try {
+          object.wait();
+          return;
+        } catch (InterruptedException e) {
+          interrupted = true;
```
cpovirk: Hmm. I am pretty sure that we need to `return` here.

`wait` is different than the other interruptible methods because `notify` is a one-shot thing: If you aren't `wait`-ing during the call, then you don't get the notification. This is different than something like `latch.await()`, where you can start waiting after the `countDown()` call and still "receive the notification."

Past discussion converged on the idea that we'd probably need to do something like this:

```
public static void waitUninterruptibly(Object waitOn) {
  boolean reinterrupt = Thread.interrupted();
  try {
    waitOn.wait();
  } catch (InterruptedException e) {
    reinterrupt = true;
  }
  if (reinterrupt) {
    selfInterrupt();
  }
}
```

Sorry for not digging this up earlier. I'd forgotten about it until I actually saw the code :(

cpovirk: http://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.2.1

And for my reference:
- http://b/4198432#comment4
- http://cl/19874683

Ram-Kasam: Interesting. I completely overlooked that difference. I am fine with the code which you shared. But wanted to discuss a possibility:

In the code which you shared, the usage pattern of the users of this API would be something like this:

```
while(!checkNotificationCondition()){
  Uninterruptibles.waitUninterruptibly(waitOn);
}
```

if we could change the API to something like this: 
```
public static void waitUninterruptibly(Object waitOn, Supplier<Boolean> notificationCondition) {
  boolean reinterrupt = Thread.interrupted();
  while(!notificationCondition.get()){
    try {
      waitOn.wait();
    } catch (InterruptedException e) {
      reinterrupt = true;
    }
  }
  if (reinterrupt) {
    selfInterrupt();
  }
}
```

And the usage will look like this:

```
Uninterruptibles.waitUninterruptibly(waitOn, this::notificationCondition())
```

Your thoughts?

guava/src/com/google/common/util/concurrent/Uninterruptibles.java
```diff
@@ -326,7 +326,67 @@ public static boolean tryAcquireUninterruptibly(
...
+      long remainingNanos = unit.toNanos(timeout);
+      long end = System.nanoTime() + remainingNanos;
+
+      while (true) {
+        try {
+          // timedWait treats negative timeouts just like zero.
+          NANOSECONDS.timedWait(object, remainingNanos);
+          return;
+        } catch (InterruptedException e) {
+          interrupted = true;
```
Ram-Kasam: Do you suggest we return here? or keep retrying till the timeout expires?

guava/src/com/google/common/util/concurrent/Uninterruptibles.java
cpovirk: That seems like a reasonable API. I think the main question is whether we want to get in the business of supplying high-level wrappers around `Object.wait` or whether we want our high-level APIs to be based on other primitives, like [`Monitor`](http://google.github.io/guava/releases/19.0/api/docs/com/google/common/util/concurrent/Monitor.html) or even `ReentrantLock`.

I'll see if I can get some other opinions on this.

cpovirk: The general opinion internally seems to be that, if people are willing to change their code, they should probably move to something like `Monitor`. Arguably we shouldn't even provide `waitUninterruptibly(Object)`, but I can at least see the value in providing a method that users can easily plug in for a broken `try-catch` block without otherwise touching the structure of their code and without introducing a usage of `Supplier` (which we'd like to phase out after Java 8).

RamAnvesh: Ok. I too feel that taking a `Supplier` as an argument is too high level. 
Also, `Object.wait()` is too low level to provide and API for. I too moved away from `Object.wait()` to something like `Condition` in my code. 
I don't want to force this into the API unless it adds substantial value. I will let you take the final call on whether we should go forward with this before I proceed further.

On a side note, I am interested to know more about what you said about phasing out the usage of `Supplier`(?) after Java 8. Would be great if you could point to some issue/article etc. regarding this.

RamAnvesh:
Keeping in mind the reasons you have given and some open questions, I am not going forward with this change.
 Closing this pull request. Please reopen if you need any further discussion.

Open question:
-  What do we do in case of timedWait methods-  return on interrupt? or keep retrying till timeout? Or take a supplier to check and retry?
- Returning on interrupt will mean that the client might need to implement code to retry with reduced time which will essentially render the API pointless
- Not returning on interrupt will mean that in some cases, the method will wait more than required
- Taking a supplier as parameter is not preferred as discussed above.

Looks like Object.wait() is not a good candidate for uninterruptibility.

Issue #2384, opened 2016-02-07
Clean up Uninterruptibles to reduce code duplication


RamAnvesh:
Currently the Uninterruptibles class has many methods which do basically the same thing:

1. No timeout: 
    - Wait on something 
    - catch InterruptedException 
    - repeat
    - set interrupted state if required 
    - return result if any

2. Timeout: 
    - Wait on something with timeout 
    - catch InterruptedException 
    - update timeout
    - repeat 
    - set interrupted state if required 
    - return result if any

This same logic is repeated multiple times in every method. And presumably, given the requests for more such methods to be added to the class, the duplication will increase. I suggest cleaning up the code with a generic implementation. This genericization will be private initially with no change in public API. Later on we might choose to publicize this generic implementation so that clients can take advantage of this class even in cases where they have blocking implementations which cannot be supported directly by this class

RamAnvesh:
I have submitted a sample implementation. Please review

cpovirk:
We used to have the code like this, but we ultimately felt it wasn't worth the extra allocations. I do wish that we had another way to reduce duplication.

RamAnvesh:
1. By allocation you mean creating a new object once in every method? 
2. Is it so much of a performance hit to worry about, given that it happens only once per call? 
3. What will it take to reconsider the decision no to go with the above design?
One of the reasons I insist is that by going down this path, we will be in a position some time in the near future to expose a generic method for clients to use with their own and other third party blocking implementations - making this class useful in almost all cases. 

If you feel strongly about avoiding object allocation at the cost of code duplication, I would like to try out an alternative approach of using reflection - Pass the required method and arguments to a single method. 
4. What is your take on such an approach?

cpovirk:
Right, an allocation per method. It's not clear how big the cost is, but `getUninterruptibly` specifically is used through our `Futures` methods, which are heavily used, including in our Android apps, where allocation costs more. The right way to determine the costs of allocation (or reflection, which is unfortunately probably worse on Android) is probably a full benchmark of such an app, which of course I can't exactly ask you to produce :(

I understand the appeal of the generic API. It's a big part of the reason that we originally did the code that way, too. But we had it for years, and no on used it. I'm sure that a handful of people could have, but that's a lower bar than we like to have for our APIs.

RamAnvesh:
Understood. Seems the the solution is not as simple as it seemed when i saw the code initially.

Thanks for your time :)
You can close this issue.

PR #2385, opened 2016-02-07
Refactored code in Uninterruptibles.

RamAnvesh:
All the major logic is now moved to a single private method.
Now there is no code duplication, and this should be help in adding new methods to this class by writing very little code without much risk of introducing bugs


RamAnvesh:
This is a sample implementation for https://github.com/google/guava/issues/2384. The design needs to be discussed before proceeding.

cpovirk:
As noted on #2384, we decided to move away from this approach a while back.

PR #2395, opened 2016-02-18
Reuse the mac instance just constructed for bit-length calculation

harningt:
No use in constructing a fresh copy of the Mac just to get the bit-length.


kluever:
Thanks - this will be mirrored out shortly!

PR #2398, opened 2016-02-21
UncheckedIOException introduced in Throwables.propagate()

### LINKED ISSUE #2397: Throwables.propagate should throw UncheckedIOException where appropriate
Currently, `Throwables.propagate` wraps all checked exceptions in a `RuntimeException`. Java 8 added `UncheckedIOException` specifically to wrap an `IOException` in an unchecked exception. `Throwables.propagate` should check for `IOException` and use this class rather than the raw `RuntimeException`.

This is a compatible change, as `UncheckedIOException` extends `RuntimeException`.

pivovarit:
This closes #2397


pivovarit:
I signed it!

PR #2417, opened 2016-03-10
Make LittleEndianDataInputStream#readAndCheckByte public

SpOOnman:
In my opinion LittleEndianDataInputStream cannot be easily used with composition. I think readAndCheckByte method should be public.

For example I need to have a method that reads 3 bytes and constructs int from that array. My implementation would be the same as readInt with first byte being 0x00. I'd like to use readAndCheckByte but it's private.


SpOOnman:
Any news on this? Is it fine or needs more work? Thanks.

PR #2420, opened 2016-03-13
Allowing Range.intersect to return empty sets when ranges are not connected

alfredodiaz:
When I am working with ranges that sometimes are connected and sometimes not, It is a litle cumbersome to have to check if the ranges are connected before any calling to the intersection method.

I added a new intersection method with a new optional flag parameter "_allowEmptyRanges_". When this flag is true, the method will return an empty set when ranges are not connected (instead of throwing an IllegalArgumentException).

The change is backward-compatible but I have included specitifc unit tests for this little change.

I would be glad to hear any comments from you.

### DIFF
--- a/guava-tests/test/com/google/common/collect/RangeTest.java
+++ b/guava-tests/test/com/google/common/collect/RangeTest.java
@@ -461,6 +461,36 @@ public class RangeTest extends TestCase {
     }
   }
 
+  public void testOverlapping() {
+    assertEquals(Range.openClosed(2, 2), Range.open(0, 1).intersection(Range.open(2, 3), true));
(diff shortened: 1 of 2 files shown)


alfredodiaz:
I signed it!

cpovirk:
This is surely useful to you and to some other users, but I think it's fundamentally no longer just an "intersection," and it's hard to explain what it is without saying "This makes sense only if you consider all empty ranges to be equivalent." That's fine, but I don't think it belongs in Guava itself.

The way I look at it is that it changes every paragraph in the documentation. Since you're adding a new method, not changing behavior, that's in some sense OK. But I think the change is too large even for an overload, and I think the operation is too niche to justify adding a separate name for.

"Returns the maximal range... _unless_..."

"The intersection exists if and only if... _unless_..."

"The intersection operation is commutative, associative and idempotent... _unless_..."

PR #2422, opened 2016-03-15
digitsOfNFactorial, getNumbersOfDigit  Methods are added

ozanonurtek:
#### public static int digitsOfNFactorial(int n)

//This method calculates digits number of number n's factorial without calculating 'n!'
//For example:for n=5--> 5!=120 --> returns 3
#### public static int getNumbersOfDigit(int n)

//This method calculates digits number of n
//For example:for n=4129--> returns 4

### DIFF
--- a/guava/src/com/google/common/math/IntMath.java
+++ b/guava/src/com/google/common/math/IntMath.java
@@ -709,6 +709,19 @@ public final class IntMath {
   //For example:for n=4129--> returns 4
 		return (n == 0 || n == 1) ? 1 : (int) (Math.floor(Math.log10(n))) + 1;
 }
+public static boolean isPrime(int p) {
+  //Check given number is prime number or not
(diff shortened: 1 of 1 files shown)


ozanonurtek:
I signed it!

ozanonurtek:
I signed it!

netdpb:
What use cases do you see for these methods? How often do you expect people to need to calculate these?

ozanonurtek:
It might be helpful for scientific use. @netdpb

netdpb:
I'm sure that's the case, but I'm not certain that that's useful frequently enough among Guava's users. We tend not to add features unless we have good evidence that they'll be generally useful and frequently used. Otherwise we could add thousands of correct features that won't get used.

I'm going to close this PR for now. If you find data that suggests that these methods would be more widely used than it seems, please reopen it.

Thank you for this suggestion.

PR #2424, opened 2016-03-16
concatBytesToArray

TeknoCepni:
Created a function that returns byte array. Take paremeters of byte values, concat all of them and returns a single array.
Example:
byte array = new byte[2];
array = concatBytesToArray(0x06,0x01);

### DIFF
--- a/guava/src/com/google/common/primitives/Bytes.java
+++ b/guava/src/com/google/common/primitives/Bytes.java
@@ -172,6 +172,25 @@ public final class Bytes {
     }
     return result;
   }
+  
+  /**
+   * Take paremeters of byte values, concat all of them and returns a single array. For example,
+   * {@code concatBytesToArray((byte)a,(byte)b,(byte)c,(byte)d) returns the array {@code {a, b,c,d}}.
+   *
+   */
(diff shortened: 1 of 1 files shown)


TeknoCepni:
I signed it!

Mihai238:
Hello,

Why not simply use `byte[] array = new byte[]{0x64, 0x61};` ?

TeknoCepni:
Hello,
If you wanna add many byte values and you don't know how many. You can't this structure.
Thanks your comment.

jonahgreenthal:
But you have to either know how many you have (to pass them into the function) or pass in an array to begin with. Calling this method creates the array anyway (that's how varargs works), and your method just copies it. I see no use for such a method.

TeknoCepni:
If you want to concat many byte values. Its useful. 
Thanks your comment.

kluever:
I agree with @TeknoCepni - I don't see why this method is useful. If you're passing var-args, then you might as well just use `new byte[] {a, b, c}`.

TeknoCepni:
I am new here.@kluever  I used to merge this method for fetching mbus datas. But you are right. 	
This can also be done in such this way.
Thanks your comment.

PR #2426, opened 2016-03-16
public static boolean isPrime(int p) 

ozanonurtek:
//Check given number is prime number or not. If number is prime return true, else return false.

### DIFF
--- a/guava/src/com/google/common/math/IntMath.java
+++ b/guava/src/com/google/common/math/IntMath.java
@@ -709,6 +709,19 @@ public final class IntMath {
   //For example:for n=4129--> returns 4
 		return (n == 0 || n == 1) ? 1 : (int) (Math.floor(Math.log10(n))) + 1;
 }
+public static boolean isPrime(int p) {
+  //Check given number is prime number or not
+  //If number is prime return true, else return false.
+		if (p == 2)
+			return true;
+		if (p % 2 == 0)
+			return false;
+		for (int i = 3; i * i <= p; i = i + 2)
+			if (p % i == 0)
+				return false;
+		return true;
+	
+}
 
   private IntMath() {}
   
--- a/guava/src/com/google/common/math/IntMath.java
+++ b/guava/src/com/google/common/math/IntMath.java
@@ -699,6 +699,17 @@ public final class IntMath {
     // The alternative (x + y) >>> 1 fails for negative values.
     return (x & y) + ((x ^ y) >> 1);
   }
+  public static int digitsOfNFactorial(int n) {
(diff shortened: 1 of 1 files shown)


guava/src/com/google/common/math/IntMath.java
```diff
@@ -699,6 +699,29 @@ public static int mean(int x, int y) {
...
+  //This method calculates digits number of number n's factorial without calculating 'n!'
+  //For example:for n=5--> 5!=120 --> returns 3
+		return (n == 0 || n == 1) ? 1 : (int) Math.floor(((n + 0.5)* Math.log(n) - n + 0.5 * Math.log(2 * Math.PI))	/ Math.log(10)) + 1;
+	}
+  public static int getNumbersOfDigit(int n) {
+  //This method calculates digits number of n
+  //For example:for n=4129--> returns 4
+		return (n == 0 || n == 1) ? 1 : (int) (Math.floor(Math.log10(n))) + 1;
+}
+public static boolean isPrime(int p) {
```
arthurv89: It returns the wrong result for p<2 since that should return false.

What about the following, which is faster and correct?
(Inspired from http://stackoverflow.com/questions/2385909/what-would-be-the-fastest-method-to-test-for-primality-in-java)

`public static boolean isPrime(int n) {
        if(n < 2) return false;
        if(n == 2 || n == 3) return true;
        if(n%2 == 0 || n%3 == 0) return false;
        int sqrtN = (int)Math.sqrt(n)+1;
        for(int i = 6; i <= sqrtN; i += 6) {
            if(n%(i-1) == 0 || n%(i+1) == 0) return false;
        }
        return true;
    }`

guava/src/com/google/common/math/IntMath.java
```diff
@@ -699,6 +699,29 @@ public static int mean(int x, int y) {
     // The alternative (x + y) >>> 1 fails for negative values.
     return (x & y) + ((x ^ y) >> 1);
   }
+  public static int digitsOfNFactorial(int n) {
```
lowasser: This is not a function I'm aware of anyone needing ever.  Can you discuss your use case?

guava/src/com/google/common/math/IntMath.java
lowasser: If we're going to provide primality testing at all, we have a much more heavily optimized implementation Google-internally that we haven't open-sourced primarily because we weren't convinced anyone would use it.

cpovirk:
We added an isPrime implementation.

ozanonurtek:
Thanks :)

PR #2428, opened 2016-03-18
Remove `@Beta` from Verify class

### LINKED ISSUE #2409: Remove @Beta from com.google.common.base.Verify
At work we use a lot the Verify utility class to make postconditions checks, it's a simple and effective way to do such checks. Nevertheless, it's marked with the Beta annotation. I read about Beta APIs https://github.com/google/guava/wiki/PhilosophyExplained#beta-apis and would like to ask for Beta removal from Verify, taking into account that the usage of Verify in our codebase has grown. Would it be possible? Thanks for your attention.

xasx:
Fixes #2409

### DIFF
--- a/guava/src/com/google/common/base/Verify.java
+++ b/guava/src/com/google/common/base/Verify.java
@@ -86,7 +86,6 @@ import javax.annotation.Nullable;
  *
  * @since 17.0
  */
-@Beta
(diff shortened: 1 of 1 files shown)


xasx:
This is my first PR here. Let's see whether the Beta API is asserted in tests...

PR #2432, opened 2016-03-22
Object can not be serializabled

heyunxia:
fix bug:
List<String> resultList = Lists.transform(list, new Function<Bean, String>() {
                        @Override
                        public String apply(Bean input) {
                            return input.getId();
                        }
                    });

after using this , the resultList can't be serializable

### DIFF
--- a/guava/src/com/google/common/collect/TransformedIterator.java
+++ b/guava/src/com/google/common/collect/TransformedIterator.java
@@ -20,6 +20,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.annotations.GwtCompatible;
 
+import java.io.Serializable;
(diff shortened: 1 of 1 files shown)


heyunxia:
I signed it!

cpovirk:
You've changed the code to make the `Iterator` serializable, but your description suggests that you want `resultList` to be serializable. Which one are you interested in?

heyunxia:
Yes, my resultList can do serializable depends on it's implementor, so i marked the top interface Iterator serializable.

lowasser:
Making the Iterator serializable shouldn't affect the actual `List` at all, and even so it's really not clear to me that single-use stateful objects like Iterators should ever be serialized.

The root issue seems to be that your `Function` isn't serializable.

PR #2441, opened 2016-04-01
Handle escape characters properly while scraping ClassPaths.

### LINKED ISSUE #2152: ClassPath.getTopLevelClasses() returns empty list.
**Summary**
ClassPath.getTopLevelClasses() returns empty list when the path of the classloader contains special characters.

**How to reproduce this error**
1. Create a packeage with three classes
2. Create the following unit test

``` java
    @Test
    public void testMain() throws Exception {
        ClassLoader loader = this.getClass().getClassLoader();
        ClassPath p = ClassPath.from(loader);
        ImmutableSet<ClassPath.ClassInfo> list = p.getTopLevelClasses("testpackage");
        Assert.assertEquals(3, list.size());
    }
```
1. Copy the whole project to a folder with special characters in its name (like % for example) and execute it.
2. The test will fail. If you copy the project to a folder without special characters in its path the test will pass.

ColinChartier:
The URI constructor for file is much more robust when it comes to handling encodings.
It saves a lot of headache of worrying about implementation specific edge cases.
(Fixes #2152)

### DIFF
--- a/guava/src/com/google/common/reflect/ClassPath.java
+++ b/guava/src/com/google/common/reflect/ClassPath.java
@@ -37,6 +37,7 @@ import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
+import java.net.URISyntaxException;
 import java.nio.charset.Charset;
 import java.util.Enumeration;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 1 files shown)


ColinChartier:
CLA done

ColinChartier:
Hmm, looks like there is something strange about the way manifests are parsed versus how ClassLoader paths are parsed.

eamonnmcmanus:
Making a regression test isn't trivial because ClassPathTest doesn't currently have a simple way to obtain sample jars, as far as I can see. We would want a test that would put the jar in something like "strange %directory"/test.jar but we'd need an input jar to put there. So it might not be feasible right now to make a test. (There's an additional layer of complexity because the test also has to work when run from Google's internal build system.)

ColinChartier:
In the mean time, is it possible to merge just this fix for the non-classpath loader? It shouldn't break any backwards compatibility, and we can fix file names with special characters here at least.

cpovirk:
We ended up attempting what I think is this with a fallback:

https://github.com/google/guava/commit/896c51abd32e136621c13d56b6130d0a72f4957a

We added some tests to create actual jars to test against. If you know of cases that still fail, let us know.

PR #2456, opened 2016-04-15
Min and Max methods for BigIntegerMath class

Sumu-Ning:


lowasser:
I'm not sure why you would need this?  You can just do `Collections.max(Arrays.asList(bigIntegerArray))` or use any of the other ways to get a max of `Comparable`s.

Sumu-Ning:
of course.
Just want to type less ...

lowasser:
Please stick with standard methods for Comparables.  If you really feel the available options for Comparables are too long, please submit a feature request for something that works for all Comparables.

PR #2457, opened 2016-04-15
Emulate Throwables in GWT

nordligulv:


cpovirk:
Thanks! I've tweaked a couple things (mostly to leave the soon-to-be-deprecated methods as `@GwtIncompatible`) and will get this submitted internally and mirrored back out.

PR #2481, opened 2016-05-16
Make Cache class keys type safe

juhovh:
Maybe this should be an issue instead of pull request, if you find this to be against the guidelines for contributing, please feel free to close this request immediately. I wanted to send a pull request instead to demonstrate how invasive the change would be. (not very)

Basically I was using `LoadingCache<String, Value>` in my application and used it to cache data based on a string key, and I also needed to invalidate the cache, so invalidate(key) was called whenever necessary. Later on I needed to change the type of the `LoadingCache` to `LoadingCache<Wrapper, Value>` in order to pass on some extra parameters for the `CacheLoader`. Everything compiled and seemed to work fine, but after a while some cache invalidation issues started to come up.

After debugging I figured out that the signature of `invalidate(key)` is actually `invalidate(Object key)`, instead of using the actual key type. The lack of cache invalidation in turn caused inconsistency in our actual data and at least a day of extra work trying to fix everything manually. So I thought if having the key type included in the signature could've saved me hours of work, maybe I'm not the only one.

In theory it would be possible that someone is using an object for invalidation that is not a subclass of key, but would still equal to it. In practice I find this rather unlikely, but I would like to hear some comments about how to proceed with this suggested change.

Also, I have not signed the Google CLA yet, but I can do it if necessary.

### DIFF
--- a/guava-tests/test/com/google/common/cache/AbstractCacheTest.java
+++ b/guava-tests/test/com/google/common/cache/AbstractCacheTest.java
@@ -79,15 +79,15 @@ public class AbstractCacheTest extends TestCase {
   }
 
   public void testInvalidateAll() {
-    final List<Object> invalidated = Lists.newArrayList();
+    final List<Integer> invalidated = Lists.newArrayList();
     Cache<Integer, Integer> cache = new AbstractCache<Integer, Integer>() {
       @Override
-      public Integer getIfPresent(Object key) {
+      public Integer getIfPresent(Integer key) {
         throw new UnsupportedOperationException();
       }
 
       @Override
-      public void invalidate(Object key) {
+      public void invalidate(Integer key) {
         invalidated.add(key);
       }
     };
--- a/guava/src/com/google/common/cache/AbstractCache.java
+++ b/guava/src/com/google/common/cache/AbstractCache.java
@@ -61,15 +61,13 @@ public abstract class AbstractCache<K, V> implements Cache<K, V> {
    * @since 11.0
    */
   @Override
-  public ImmutableMap<K, V> getAllPresent(Iterable<?> keys) {
+  public ImmutableMap<K, V> getAllPresent(Iterable<? extends K> keys) {
     Map<K, V> result = Maps.newLinkedHashMap();
-    for (Object key : keys) {
+    for (K key : keys) {
       if (!result.containsKey(key)) {
-        @SuppressWarnings("unchecked")
-        K castKey = (K) key;
         V value = getIfPresent(key);
         if (value != null) {
-          result.put(castKey, value);
+          result.put(key, value);
         }
       }
     }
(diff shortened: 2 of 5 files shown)


jvah:
It looks like https://github.com/google/guava/commit/9eed06311194b0060c42ea9662cc6d3cb76e6a43 has intentionally erased types from getIfPresent and getAllPresent, so I would be ok by not touching these two and modify the commit accordingly. But for invalidate the case still stands...

juhovh:
It looks like 9eed063 has intentionally erased types from getIfPresent and getAllPresent, so I would be ok by not touching these two and modify the commit accordingly. Don't know the details behind that change, but for invalidate the case still stands in my opinion...

juhovh:
Looks like a similar issue is discussed in http://smallwig.blogspot.co.uk/2007/12/why-does-setcontains-take-object-not-e.html blog post and the end result would be that the current behaviour is more correct. However, in my particular case the `Well, when I typed that code into IntelliJ, it flagged a warning for me right away. ` didn't seem to stand and there were no warnings whatsoever with IntelliJ IDEA 15.

Again, feel free to close this pull request. Guess I just need to solve this issue with big warning messages, but it feels quite wrong...

cpovirk:
Sorry for the trouble that this caused you. This is indeed intentional. The best documentation I know of at the moment is [Error Prone's](http://errorprone.info/bugpattern/CollectionIncompatibleType). We do want to extend Error Prone to catch bugs not just with the use of JDK collections but also with `Cache`, but we haven't gotten there yet. Our hope is to do so by introducing some kind of `@TypeCompatibleWith` annotation. If we do that, then IntelliJ could read the annotation and provide the same warning. (Or they could just special-case `Cache` today.)

kevinb9n:
We are hoping to extend that Error Prone check to catch Guava classes like
Cache, but actually FindBugs has beaten us to that by a few years. I would
suggest adopting it in your project. It's unfortunate, but static analysis
is the only real solution here.

On Mon, May 16, 2016 at 6:44 AM, Chris Povirk <notifications@github.com>
wrote:

> Sorry for the trouble that this caused you. This is indeed intentional.
> The best documentation I know of at the moment is Error Prone's
> <http://errorprone.info/bugpattern/CollectionIncompatibleType>. We do
> want to extend Error Prone to catch bugs not just with the use of JDK
> collections but also with Cache, but we haven't gotten there yet. Our
> hope is to do so by introducing some kind of @TypeCompatibleWith
> annotation. If we do that, then IntelliJ could read the annotation and
> provide the same warning. (Or they could just special-case Cache today.)
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/google/guava/pull/2481#issuecomment-219427998>
>



-- 
Kevin Bourrillion | Java Librarian | Google, Inc. | kevinb@google.com

juhovh:
I forgot to reply to this, thanks to both of you for suggestions and I'll consider adding FindBugs to the project development process.

For now I'm avoiding the issue by having a generic wrapper around the different cache types we're using with the key type always defined, and unit tests for the cache invalidation. Need to see if that becomes insufficient at some point.

PR #2500, opened 2016-06-11
Fix usage example of ImmutableSortedMultiset.Builder

FRedEnergy:
Currently, example code in the description of ImmutableSortedMultiset.Builder class contains call for constructor which doesn't exist. This PR replaces constructor with `ImmutableSortedMultiset.naturalOrder()` call


FRedEnergy:
I signed it!

cpovirk:
Thanks. I think I'm going to keep it as a constructor call but add the required parameter.

Issue #2509, opened 2016-06-27
DoubleMath incorrectly rounds 0.99999999999999994 to 2 with rounding mode UP or CEILING


terzerm:
**Failure cases:**
_1) Positive_
Values rounded incorrectly: `0.99999999999999984` to `0.99999999999999994`
Rounding modes: CEILING or UP
Expected: `1`
Actual: `2`

_2) Negative_
Values rounded incorrectly:` -0.99999999999999984` to `-0.99999999999999994`
Rounding modes: FLOOR or UP
Expected: `-1`
Actual: `-2`

**Remarks:**
The reason is that values between 0 and 1 have higher fraction precision than values between 1 and 2. The code adds 1.0 to the value x, and double arithmetic uses HALF_EVEN rounding by default. This leads to unexpected results for edge cases when HALF_EVEN triggers rounding UP to 2.

**Proposed FIX:**
The FIX is to cast the double value `x` to a long before the addition of `1.0`:
Current:` x + 1.0`
Fixed: `(long)x + 1.0`

**Pull Request:**
See pull request [#2511](https://github.com/google/guava/pull/2511)

lowasser:
Fixed the bug; I think we ended up slightly changing the implementation (and doing much more aggressive testing).  The fix should be mirrored out shortly.

PR #2511, opened 2016-06-27
#2509: Fix for 'DoubleMath incorrectly rounds 0.99999999999999994 to 2 with rounding mode UP or CEILING

terzerm:
See Issue #2509

### DIFF
--- a/guava/src/com/google/common/math/DoubleMath.java
+++ b/guava/src/com/google/common/math/DoubleMath.java
@@ -66,14 +66,14 @@ public final class DoubleMath {
         if (x >= 0.0 || isMathematicalInteger(x)) {
           return x;
         } else {
-          return x - 1.0;
+          return (long)x - 1.0;
         }
 
       case CEILING:
         if (x <= 0.0 || isMathematicalInteger(x)) {
           return x;
         } else {
-          return x + 1.0;
+          return (long)x + 1.0;
         }
 
       case DOWN:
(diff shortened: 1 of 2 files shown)


terzerm:
I signed an individual CLA.

guava/src/com/google/common/math/DoubleMath.java
```diff
@@ -66,14 +66,14 @@ static double roundIntermediate(double x, RoundingMode mode) {
         if (x >= 0.0 || isMathematicalInteger(x)) {
           return x;
         } else {
-          return x - 1.0;
+          return (long)x - 1.0;
```
eamonnmcmanus: I think this will do the wrong thing if x is too large to fit in a long.

terzerm: valid point --- should check that long boundaries are not breached.

weird that there's no test coverage for this --- cause i have run all DoubleMath tests.

lowasser: I concur, Eamonn: if x doesn't fit in a long, it's a mathematical integer anyway.

eamonnmcmanus: OK.

cpovirk:
Fixed a little differently in 1b1163b7e2c121d4a5b25b8966714201551976c4. Thanks for reporting.

PR #2515, opened 2016-06-29
Add possibility to create Striped lock with 'fair ordering policy' = true

armsargis:
Added static method to create Striped with Locks with 'fair ordering policy' = true


armsargis:
I signed it!

JFSoftwareServices:
hi

blekinge:
Could we get this request pushed or rejected? Just today I had to hack a new StripedFactory just to be able to set fair=true.

Issue #332, opened 2014-10-31
Ordering.explicit should handle equals values not only strictly ordered values, and should provide default lowest or highest order for unknown values


gissuebot:
_[Original issue](https://code.google.com/p/guava-libraries/issues/detail?id=332) created by **amertum** on 2010-02-24 at 01:10 PM_

---

I need explicit ordering with elements that are equals such as

<pre>
Ordering.explicit(a, [b1, b2], c).compare(b1, b2) returns 0
</pre>

explicit signature should be discussed, may be 
Ordering.explicit(List&lt;Iterable&lt;T>> elements) should be enough ?

I need also ordering other non-explicit elements such as

<pre>
Ordering.explicit(UNKNOWN_LOWER, a, [b1, b2], c).compare(d, a) returns -1

or

Ordering.explicit(UNKNOWN_GREATER, a, [b1, b2], c).compare(d, a) returns 1
</pre>

because right now, com.google.common.collect.ExplicitOrdering.rank(T) 
throws IncomparableValueException when the T value is known

I joined a first implementation that handle equals elements fully inspired 
from package visible com.google.common.collect.ExplicitOrdering

What do you think of this need ?

kevinb9n:
We still want to do some internal experimentation and potential-usage research.

cowwoc:
@kevinb9n Can we please decouple these two unrelated issues?

1. Add `Ordering.explicit(Iterable)`
2. Ability to specify handling for unknown values

The first one could be implemented right away. My use-case is needing to invoke `Ordering.explicit(Queue)`. The code in question is very performance-sensitive (invoked millions of times) so ideally I want to avoid converting the `Queue` to a `List` per invocation.

cpovirk:
(1) is #1542. Can you post there with some benchmarks for the performance gains you'd get from `Iterable` handling?

Performance concerns are something we haven't see much in the past. (Possibly that's because passing a plain `Iterable` would perform about as poorly as `newArrayList(iterable)` already does, since we basically need to copy it to a list internally, while a `Queue` or other `Collection` should be better.)

cowwoc:
@cpovirk I must have missed something then. Where in the code do you depend on the input being a `List`? As far as I can tell, you simply invoke a for-each loop on it, which means that an `Iterable` or `Collection` input would result in identical behavior without needing to copy the input to a `List` beforehand.

cowwoc:
[...]

PR #2517, opened 2016-07-04
Issue google/guava#332: Ability to specify handling for unknown values

dtrunk90:


dtrunk90:
I signed it!

Am 04.07.2016 um 19:37 schrieb googlebot:

cpovirk:
We do expect to support this use case eventually, but we're still looking to settle the API. Hopefully someday :(

PR #2518, opened 2016-07-06
Fix HttpHeaders constants to be lowercase.

swankjesse:
HTTP/2 requires lowercase, and it's our collective future. Any code that cares
about the casing of headers is broken, so any code that breaks from this change
is destined to break when it's used with HTTP/2.

String.toLowerCase() and Ascii.toLowerCase() both have optimizations that make
code run faster if the input is already lowercase. I estimate that 0.0000001%
of Google's extensive compute resources are currently wasted converting headers
to lowercase when they could be lowercase already.

### DIFF
--- a/guava-tests/test/com/google/common/net/HttpHeadersTest.java
+++ b/guava-tests/test/com/google/common/net/HttpHeadersTest.java
@@ -17,16 +17,10 @@
 package com.google.common.net;
 
 import com.google.common.base.Ascii;
-import com.google.common.base.Joiner;
-import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableBiMap;
 import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Lists;
-
-import junit.framework.TestCase;
-
 import java.lang.reflect.Field;
-import java.util.List;
+import junit.framework.TestCase;
 
 /**
  * Tests for the HttpHeaders class.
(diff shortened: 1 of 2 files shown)


kluever:
Oh man, I think this is going to cause some serious breakages inside of google if we merge it. This is pretty scary.

> HTTP/2 requires lowercase

Can you send a reference to the RFC or whatever? I'm going to start an internal thread about this and see what folks think.

kluever:
Ahh, I think I found it:

> header field names MUST be converted to lowercase prior to their encoding in HTTP/2. A request or response containing uppercase header field names MUST be treated as malformed

https://httpwg.github.io/specs/rfc7540.html#rfc.section.8.1.2

kluever:
Hey Jesse,

We discussed this internally quite a bit, and decided to leave it as is.

> The fact that HTTP headers are lowercase on the wire is an implementation detail. It does not need to be surfaced to users.

And...

> The HTTP/2 protocol is mostly binary, so you'd be using an encoder anyway.

Yes, there is a slight performance penalty for lowercasing, but it's not even close to large enough for us to try to push a change of this magnitude through.

Cheers,
-kak

swankjesse:
Works for me. Two quick notes about why this is too bad:

1. Any code that cares about case sensitivity here is almost certainly incorrect. Until recently this was academic. But any header that travels through an HTTP/2 channel comes out the other end in lowercase. ”Pushing a change of this magnitude through” is fixing buggy code.

2. Because HTTP/2 exists, attempting to preserve case in HTTP headers is now a waste of code and CPU cycles. For example we use slow awkward TreeMaps with case-insensitive comparators rather than fast convenient LinkedHashMaps with lowercase keys.

PR #2541, opened 2016-08-03
while-statement did not loop

leksak:
A return statement occurred in the first
iteration of the loop making the loop
statement redundant, as it is semantically
equivalent to an if-statement.

Replaced the loop-statement with the
equivalent if.


leksak:
I signed it!

PR #2544, opened 2016-08-08
Implement the indexOf method of the CartesianList

hadrienk:
The implementation of the method indexOf(Object o) in the AbstractList iterates over the iterator. In the case of Cartesian list, the index can be computed instead.

### DIFF
--- a/guava/src/com/google/common/collect/CartesianList.java
+++ b/guava/src/com/google/common/collect/CartesianList.java
@@ -71,6 +71,27 @@ final class CartesianList<E> extends AbstractList<List<E>> implements RandomAcce
   }
 
   @Override
+  public int indexOf(Object o) {
+    if (!(o instanceof List)) {
+      return -1;
+    }
+    List<?> list = (List<?>) o;
+    if (list.size() != axes.size()) {
+      return -1;
+    }
+    ListIterator<?> itr = list.listIterator();
+    int computedIndex = 0;
+    while (itr.hasNext()) {
+      int axeIndex = itr.nextIndex(), elemIndex;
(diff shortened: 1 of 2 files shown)


hadrienk:
Hi! Can I help with something to get this PR merged?

hadrienk:
@lowasser @kluever

guava/src/com/google/common/collect/CartesianList.java
```diff
@@ -70,6 +70,27 @@ private int getAxisIndexForProductIndex(int index, int axis) {
...
+      return -1;
+    }
+    List<?> list = (List<?>) o;
+    if (list.size() != axes.size()) {
+      return -1;
+    }
+    ListIterator<?> itr = list.listIterator();
+    int computedIndex = 0;
+    while (itr.hasNext()) {
+      int axeIndex = itr.nextIndex(), elemIndex;
```
ronshapiro: Google Style doesn't allow for multiple variables to be declared in the same statement. Can you split this up?

I think it would also make the next conditional much clearer:

```java
int axisIndex = itr.nextIndex();
int elemIndex = axes.get(axisIndex).indexOf(itr.next());
if (elemIndex == -1) {
  return -1;
}
...
```

ronshapiro: `axisIndex` is more appropriate - `axes` is not the multiple of axe here, rather it's the multiple of `axis`

cpovirk:
(IIRC Louis may already have imported this and made the style fixes. Not on my laptop to check though.)

ronshapiro:
Ah, I see this is in https://github.com/google/guava/pull/3197, but @lowasser didn't attribute it correctly

PR #2545, opened 2016-08-10
Emulate Queue/Deque impls in GWT.

nordligulv:
GWT 2.8 has ArrayDeque implementation


cpovirk:
Nice, didn't realize that GWT had added this.

And thanks. I didn't realize that we'd fully duplicated `TreeTraverser`, either.

atmiklas:
<empty>

PR #2546, opened 2016-08-11
Provide emulation for classes that use BitSet.

nordligulv:
GWT 2.8 added BitSet emulation.


guava/src/com/google/common/base/CharMatcher.java
```diff
@@ -596,7 +596,6 @@ private static boolean isSmall(int totalCharacters, int tableLength) {
   /**
    * Sets bits in {@code table} matched by this matcher.
    */
-  @GwtIncompatible // java.util.BitSet
```
cpovirk: FYI, I will probably leave these as `@GwtIncompatible`, since they're used only from other `@GwtIncompatible` code. (We might as well spare the GWT compiler the trouble of compiling and discarding them, and we might as well give ourselves the option of adding non-GWT-compatible code to them in the future.) I'll update the comment, though.

cpovirk:
Thanks. Looking at this now.

spirylics:
Since, this error appears now : 
[INFO]    Tracing compile failure path for type 'com.google.common.collect.BinaryTreeTraverser'
[INFO]       [ERROR] Errors in 'jar:file:/Users/cyril/.m2/repository/com/google/guava/guava-gwt/20.0-SNAPSHOT/guava-gwt-20.0-SNAPSHOT.jar!/com/google/common/collect/BinaryTreeTraverser.java'
[INFO]          [ERROR] Line 134: No source code is available for type java.util.BitSet; did you forget to inherit a required module?
[INFO]    [ERROR] Aborting compile due to errors in some input files

Somebody has an idea ?

nordligulv:
@spirylics https://github.com/google/guava/issues/2547

PR #2555, opened 2016-08-23
Fix Issue #2554, Helpers#isEmpty

ghonzo:
As detailed in issue #2554 there was a bug in com.google.common.collect.testing.Helpers#isEmpty when passed an Iterable that is not a Collection.

Added a test that fails before the patch, succeeds after.


lowasser:
I think we're actually going to aim to kill off this class entirely and replace it with Truth rather than try to fix its brokenness.

ghonzo:
Fair enough! Thanks for the direction.

lowasser:
Bleahhhhh it looks like we might have to live with this class due to some dependency weirdnesses.

Going to reopen this so we can fix this issue, at least.

kluever:
@lowasser do you mind handling this request? thanks

lowasser:
I'm pretty sure I got this already, not sure if it's sync'd out yet though.

PR #2558, opened 2016-09-03
Adding checkPositive MathPreconditions before using value of x

dpsingh016:
Adding checkPositive MathPreconditions before using value of x in ceilingPowerOfTwo and floorPowerOfTwo methods.

### DIFF
--- a/guava/src/com/google/common/math/BigIntegerMath.java
+++ b/guava/src/com/google/common/math/BigIntegerMath.java
@@ -57,6 +57,7 @@ public final class BigIntegerMath {
    */
   @Beta
   public static BigInteger ceilingPowerOfTwo(BigInteger x) {
+    checkPositive("x", checkNotNull(x));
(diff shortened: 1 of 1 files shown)


dpsingh016:
Updated email address

lowasser:
Not sure why you feel this is necessary?  log2, which these methods call immediately, already does this check.

dpsingh016:
Presently we are tightly coupled with log2 method to enforce the method contract "@throws IllegalArgumentException if {@code x <= 0}"

I'm following fail fast concept at this place. If we know that for non positive values of x this method is going to fail we can fail it even before calling log2 method.

lowasser:
There's no reason to check twice.  We have tests to make sure the exception gets thrown somewhere, and throwing the exception is part of log2's contract, which we can depend on.

cpovirk:

PR #2562, opened 2016-09-11
Fix MultiInputStream.read(byte[]): do not stop after first stream...

jbgi:
if there is still bytes to read.


jbgi:
CLA signed.

jbgi:
After examination of the javadoc, the current behavioir is not a bug:

> An attempt is made to read as many as `len` bytes, but a smaller number may be read.

It is just that the code that exposed bad behavior did not use `com.google.common.io.ByteStreams#read` or equivalent.

PR #2580, opened 2016-09-28
Remove reference to a deprecated method

Xaerxess:
I found these comments slightly confusing. (I also wonder whether you removed `sameThreadExecutor` internally, since it's scheduled for removal in August 2016, but I guess it means in "when Guava 20 is out".)


cpovirk:
Yeah, we've fallen behind on removing things. `sameThreadExecutor` is on our to-do list. We won't hold up Guava 20 over it, though, so it might hang on longer :(

Anyway, thanks for the comment fix.

PR #2581, opened 2016-09-29
Apply IntelliJ-suggested language feature migrations.

jbduncan:
All code which initializes an IOException with a cause exception now uses the appropriate constructor (which was introduced in Java 6), rather than initCause. Since Guava now depends on Java 6, this (as far as I can tell) is a safe change to make.

A change was made from using an Iterator to an enhanced for-loop for (arguably) improved readability and, by my understanding, no performance impact.

All other changes should hopefully be self-explanatory. :)

### DIFF
--- a/guava/src/com/google/common/collect/RegularImmutableSortedSet.java
+++ b/guava/src/com/google/common/collect/RegularImmutableSortedSet.java
@@ -162,9 +162,7 @@ final class RegularImmutableSortedSet<E> extends ImmutableSortedSet<E> {
     if (SortedIterables.hasSameComparator(comparator, that)) {
       Iterator<?> otherIterator = that.iterator();
       try {
-        Iterator<E> iterator = iterator();
-        while (iterator.hasNext()) {
-          Object element = iterator.next();
+        for (E element : this) {
           Object otherElement = otherIterator.next();
           if (otherElement == null || unsafeCompare(element, otherElement) != 0) {
             return false;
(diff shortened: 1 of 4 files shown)


jbduncan:
I signed it!

guava/src/com/google/common/collect/RegularImmutableSortedSet.java
```diff
@@ -162,9 +162,7 @@ public boolean equals(@Nullable Object object) {
     if (SortedIterables.hasSameComparator(comparator, that)) {
       Iterator<?> otherIterator = that.iterator();
       try {
-        Iterator<E> iterator = iterator();
-        while (iterator.hasNext()) {
-          Object element = iterator.next();
+        for (E element : this) {
```
cpovirk: Exhaustive, definitive, guaranteed-correct analysis of this change by me and one other guy shows that the code was better how it was.

Or, you know, we just kind of prefer that way :-P We like the parallel `next()` calls. I'm submitting the change internally (minus the one part here), and we'll mirror it out shortly.

jbduncan: Haha, fair enough!

Seems IntelliJ isn't _quite_ as infallible as I thought. 😋 And I also agree to some extent that consistent, parallel iterator calls are better than replacing one of them with a for loop, so I'm happy with this.

PR #2582, opened 2016-09-29
Replace unnecessary StringBuilder calls with String concat

jbduncan:
According to IntelliJ, these usages of StringBuilder can be simplified. By my understanding, this is because they do not contain any loops, and thus, as of Java 6, JVMs are able to optimize these concatenations into more-efficient StringBuilder calls at compile time. Thus this change improves readability without losing performance.


lowasser:
Heh.  I'm at least partially responsible for that change in upstream JDK, so it's appropriate we apply those simplifications to Guava now...

PR #2567, opened 2016-09-17
Update Mockito version to 2.1.0-RC.1

TimvdLippe:
Last week we released our first release candidate of Mockito 2.
This release should be backwards compatible with the codebase
of Google Guava.

To read about all changes in Mockito 2, check out
https://github.com/mockito/mockito/wiki/What%27s-new-in-Mockito-2

I have ran your test suite locally before and after this change and
did not discover any regressions nor test failures.
While this version is a release candidate, we are not planning
to introduce any changes for the final release. There might be
possible bug fixes included in the final release, but so far
we have not received any reports of regressions in our library.

Thank you for using Mockito and please let us know what you think!

PR #2588, opened 2016-10-04
Update Mockito version to 2.1.0

Xaerxess:
Mockito 2.1.0 is out! It should be a smooth upgrade from RC-1.


Xaerxess:
See #2567 for context.

PR #2592, opened 2016-10-06
micro sign

yecril71pl:
The micro sign is a different character than Greek small letter mu.  The current rendering is in error.  Moreover, it does not print on an 8-bit console unless the code page is Greek, which is insane.


simon04:
See #2593 for a discussion on this change.

cpovirk:

PR #2595, opened 2016-10-06
Update StopwatchTest.java

yecril71pl:
Use micro sign for microsecond unit.


simon04:
This PR seems to be a subset of #2592 and should be closed?

PR #2614, opened 2016-10-26
fixes serialization errors when serializing a RegularImmutableAsList

### LINKED ISSUE #1841: RegularImmutableAsList CustomFieldSerializer depends on RegularImmutableSet but did not include it into Serializationpolicy
_[Original issue](https://code.google.com/p/guava-libraries/issues/detail?id=1841) created by **jbjrke** on 2014-09-03 at 09:52 AM_
---

if you have a RemoteService with only 1 method and the signature:

```
void testImmutableList(ImmutableList<String> list);
```

and you call it with

```
testImmutableList(ImmutableSet.of("a", "b").asList())
```

you will get an exception:

```
com.google.gwt.user.client.rpc.SerializationException: Type 'com.google.common.collect.RegularImmutableSet' was not included in the set of types which can be deserialized by this SerializationPolicy or its Class object could not be loaded. For security purposes, this type will not be deserialized.
    at com.google.gwt.user.server.rpc.impl.StandardSerializationPolicy.validateDeserialize(StandardSerializationPolicy.java:158)
```

this happens because the delegate of RegularImmutableAsList is a RegularImmutableSet which is not included into the serialization policy but is written by the RegularImmutableAsList_CustomFieldSerializer

```
writer.writeObject(instance.delegateCollection());
writer.writeObject(instance.delegateList());
```

bjrke:
the issue occur when ImmutableList is included in gwt's serialization
policy and ImmutableSet not and a list created by ImmutableSet.asList
needs to be serialized

fixes #1841

### DIFF
--- a/guava-gwt/src/com/google/common/collect/RegularImmutableAsList_CustomFieldSerializer.java
+++ b/guava-gwt/src/com/google/common/collect/RegularImmutableAsList_CustomFieldSerializer.java
@@ -16,10 +16,13 @@
 
 package com.google.common.collect;
 
+import java.util.ArrayList;
+
 import com.google.common.annotations.GwtCompatible;
 import com.google.gwt.user.client.rpc.SerializationException;
 import com.google.gwt.user.client.rpc.SerializationStreamReader;
 import com.google.gwt.user.client.rpc.SerializationStreamWriter;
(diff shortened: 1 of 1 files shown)


bjrke:
I signed it!

bjrke:
I signed it!

PR #2615, opened 2016-10-28
Fix #1190

### LINKED ISSUE #1190: Splitter.onPattern drops last token when zero width matches are used

jbduncan:
Commit text body:

> Before this fix, splitting a single character input string with a
> Splitter.onPattern instance created with a zero-width regex pattern,
> would have caused the input string to be dropped from the output,
> resulting in an empty iterable being returned rather than a single
> element one.
> 
> This fix ensures that the input passes through untouched.
> 
> For example, whereas before in this code snippet, 'words' would have
> been initialized as an empty iterable...
> 
> ```
> String string = "f";
> Iterable<String> words =
>     Splitter.on(Pattern.compile("\\b")).split(string);
> // words is empty!
> ```
> 
> ...it is now initialized as ["f"].

### DIFF
--- a/guava-tests/test/com/google/common/base/SplitterTest.java
+++ b/guava-tests/test/com/google/common/base/SplitterTest.java
@@ -383,6 +383,20 @@ public class SplitterTest extends TestCase {
   }
 
   @GwtIncompatible // java.util.regex.Pattern
+  public void testPatternSplitWordBoundary_singleCharInput() {
+    String string = "f";
+    Iterable<String> words = Splitter.on(Pattern.compile("\\b")).split(string);
+    assertThat(words).containsExactly("f").inOrder();
+  }
+
+  @GwtIncompatible // java.util.regex.Pattern
+  public void testPatternSplitWordBoundary_singleWordInput() {
+    String string = "foo";
+    Iterable<String> words = Splitter.on(Pattern.compile("\\b")).split(string);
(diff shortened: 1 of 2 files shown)


cpovirk:
Sorry for dropping this. We ended up accepting similar https://github.com/google/guava/pull/2615 instead (for no good reason).

PR #2620, opened 2016-11-01
Link local address

rudenvla:
link-local address validation with percent and scope id


rudenvla:
I signed it

nglorioso:
Hi there! Thanks for sending us this PR!

One thing I want to talk about here is that, while you've updated the validation function, `InetAddresses.forString` doesn't appear to preserve the scope part of the IPV6 address.

cpovirk:
I am optimistically assuming that this got handled by https://github.com/google/guava/commit/8a7d36a8e5044096919a3512aa6734c95032f38c#diff-1207ec0a4b5d3f5e5c2236b7373eefc9, but please reopen if not.

PR #2621, opened 2016-11-02
fix warning on eclipse4.6 (com.google.common.collect.MutableClassToInstanceMap)

hand515:
Warning on eclipse , 

`
Bound mismatch: The generic method cast(Class<T>, B) of type MutableClassToInstanceMap<B> is not applicable for the arguments (capture#8-of ? extends Class<? extends B>, capture#12-of ? extends B). The inferred type capture#10-of ? extends B is not a valid substitute for the bounded parameter <T extends B>
`


hand515:
I signed it!

PR #2624, opened 2016-11-03
Make SloppyTearDown.logger private.

dlazerka:
Being public, it always gets into way of our class (import is done automatically by auto-importing tools).
It's not used anywhere else, so can be `private`.
If it's intended to be used by inheritance, then it can be `protected` as well.
If it really neede to be public, then it's better be available through a getter method.


dlazerka:
I signed it!

PR #2625, opened 2016-11-04
Use a newer JDK 8 for travis build

cushon:
The travis docker image uses JDK 8u31, which is ancient and has type
inference bugs that are fixed at head. This change causes it to use 8u111:

https://github.com/travis-ci/travis-ci/issues/3259#issuecomment-130860338

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,12 @@ language: java
 
 jdk:
   - oraclejdk8
-  - oraclejdk7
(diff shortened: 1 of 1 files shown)


cushon:
The JDK 8 build looks good now: https://travis-ci.org/google/guava/jobs/173130822

I update the PR to also turn off the JDK 7 builds.

Issue #2617, opened 2016-10-29
Add lastIndexOf with array target in primitives' utility classes


sim642:
Primitives' utility classes such as `Bytes` have methods like
* `indexOf(byte[] array, byte target)`
* `indexOf(byte[] array, byte[] target)`
* `lastIndexOf(byte[] array, byte target)`
for working with arrays of those primitives.

Unfortunately the `indexOf` and `lastIndexOf` aren't symmetric in this case as there isn't a method like `lastIndexOf(byte[] array, byte[] target)`, taking an array as target in the `lastIndexOf` case. Hereby I propose that such method be added to `Bytes` and other analogous primitives' utility classes.

blzb:
I really need this too.

fdesu:
I come across this functionality and wanted to contribute it. I've already created PR with a draft of the changes for `int lastIndexOf(array[], target[])`. I've created PR to gather an early feedback to not waste anyone's time to understand if there is still need in such an enhancement and if yes wether I go the right way.

PR #2626, opened 2016-11-04
Added the lastIndexOf method for the Bytes primitives class for both parameters as byte array types.

luisalduucin:
Added the lastIndexOf method which receives two byte array as parameters for the Bytes primitive class and its corresponding tests.

It is a fix for the issue #2617 .


luisalduucin:
I signed it!

luisalduucin:
I signed it!

PR #2627, opened 2016-11-04
Added lastIndexOf(byte[] array, byte[] target) method to Bytes primit…

luisalduucin:
Added lastIndexOf(byte[] array, byte[] target) method to Bytes primitive class and its corresponding tests.

It is a fix for the issue #2617 .


luisalduucin:
I signed it!

.travis.yml
```diff
@@ -4,8 +4,11 @@ language: java
 
 jdk:
   - oraclejdk8
-  - oraclejdk7
-  - openjdk7
+
+addons:
+  apt:
+    packages:
+      - oracle-java8-installer
```
sim642: Since the current master already contains this change it'd be nicer (and less conflicting) to rebase your changes there, instead of imitating changes in your own branch.

guava/src/com/google/common/primitives/Bytes.java
```diff
@@ -152,6 +152,43 @@ private static int lastIndexOf(byte[] array, byte target, int start, int end) {
...
+    checkNotNull(target, "target");
+    if (target.length == 0) {
+      return array.length - 1;
+    } else if (target.length > array.length) {
+      return -1;
+    }
+
+    int lastIndexOf = -1;
+    boolean differentValue;
+    for (int i = 0; i <= array.length - target.length; i++) {
```
sim642: The way this search is implemented here, it is highly inefficient because it searches through every single match from the beginning of the array and keeps the last one. Instead it'd be wiser to do the `i` indexing in reverse like `Bytes.lastIndexOf(byte[], byte)` does it. Such approach also avoids the need for helper variables which you've used here.

guava/src/com/google/common/primitives/Bytes.java
```diff
@@ -152,6 +152,43 @@ private static int lastIndexOf(byte[] array, byte target, int start, int end) {
...
+   *
+   * @param array the array to search for the sequence {@code target}
+   * @param target the array to search for as a sub-sequence of {@code array}
+   */
+  public static int lastIndexOf(byte[] array, byte[] target) {
+    checkNotNull(array, "array");
+    checkNotNull(target, "target");
+    if (target.length == 0) {
+      return array.length - 1;
+    } else if (target.length > array.length) {
```
sim642: This check is unnecessary because the following outer `for` loop would immediately terminate anyway, were this the case.

guava/src/com/google/common/primitives/Bytes.java
```diff
@@ -152,6 +152,43 @@ private static int lastIndexOf(byte[] array, byte target, int start, int end) {
...
+   * {@code target}.
+   *
+   * @param array the array to search for the sequence {@code target}
+   * @param target the array to search for as a sub-sequence of {@code array}
+   */
+  public static int lastIndexOf(byte[] array, byte[] target) {
+    checkNotNull(array, "array");
+    checkNotNull(target, "target");
+    if (target.length == 0) {
+      return array.length - 1;
```
sim642: Why has this chosen to return `array.length - 1` and not `array.length`?

The former essentially represents the 0-length subarray before the last element, not after it. The latter makes more semantical sense to me for that reason.

guava-tests/test/com/google/common/primitives/BytesTest.java
```diff
@@ -119,6 +119,39 @@ public void testLastIndexOf() {
         (byte) 3));
   }
 
+  public void testLastIndexOf_arrayTarget() {
+    assertEquals(-1, Bytes.lastIndexOf(EMPTY, EMPTY));
```
sim642: `Bytes.indexOf(byte[], byte[])` finds an empty array in an empty array. For symmetry it would make sense for the same to happen when searching from the other direction, i.e. this should be `0` not `-1`.

luisalduucin: Sure, it makes sense ! Thanks !

guava/src/com/google/common/primitives/Bytes.java
luisalduucin: Yep, once I coded it I noticed that.

guava/src/com/google/common/primitives/Bytes.java
luisalduucin: I do not get your point. Could you explain it in another way ?

.travis.yml
luisalduucin: Yes ! thanks !

liach:
Well this way of approach is not fast enough. I suggest you to look at [KMP algorithm] (https://en.wikipedia.org/wiki/Knuth%E2%80%93Morris%E2%80%93Pratt_algorithm) for such a matching.

.travis.yml
```diff
@@ -5,6 +5,7 @@ language: java
 jdk:
   - oraclejdk8
 
+
```
liach: This change is not necessary.

liach:
@luisalduucin You can add these to all the primitive classes.

luisalduucin:
@liach Yes, sure I will, but before I would like that a Google member or guava contributor check if my approach is correct in order to proceed. Thanks.

guava/src/com/google/common/primitives/Bytes.java
```diff
@@ -152,6 +152,36 @@ private static int lastIndexOf(byte[] array, byte target, int start, int end) {
...
+   * {@code target}.
+   *
+   * @param array the array to search for the sequence {@code target}
+   * @param target the array to search for as a sub-sequence of {@code array}
+   */
+  public static int lastIndexOf(byte[] array, byte[] target) {
+    checkNotNull(array, "array");
+    checkNotNull(target, "target");
+    if (target.length == 0) {
+      return array.length;
```
liach: Or
```java
return array.length - 1;
```

liach: Yes you are right, `array.length` can be used for `Arrays.copyOfRange`

liach:
Ok... been dead for around 15 months.

PR #2634, opened 2016-11-08
Improved the performance of indexOf methods of primitives

liach:
Improved the performance of indexOf methods of primitives.
The original time efficiency was O(nm), where n is the size of the array
and m is the size of target. The new time efficiency is O(n).
The algorithm used is Sunday algorithm, a string matching algorithm that can be used on arrays.
All tests passed locally.

### DIFF
--- a/guava/src/com/google/common/primitives/Booleans.java
+++ b/guava/src/com/google/common/primitives/Booleans.java
@@ -138,15 +138,23 @@ public final class Booleans {
     if (target.length == 0) {
       return 0;
     }
-
-    outer:
-    for (int i = 0; i < array.length - target.length + 1; i++) {
-      for (int j = 0; j < target.length; j++) {
-        if (array[i + j] != target[j]) {
-          continue outer;
+    int i = 0;
+    while (i < array.length) {
+      int j = 0;
+      while (j < target.length && i + j < array.length && array[i + j] == target[j]) {
+        j++;
+      }
+      if (j == target.length) {
+        return i;
+      }
+      if (i + target.length < array.length) {
+        for (j = target.length - 1; j >= 0; j--) {
+          if (array[i + target.length] == target[j]) {
+            break;
+          }
         }
       }
-      return i;
+      i += target.length - j;
     }
     return -1;
   }
--- a/guava/src/com/google/common/primitives/Bytes.java
+++ b/guava/src/com/google/common/primitives/Bytes.java
@@ -116,15 +116,23 @@ public final class Bytes {
     if (target.length == 0) {
       return 0;
     }
-
-    outer:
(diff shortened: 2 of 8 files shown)


lowasser:
Though it has a better worst case, it's not clear to me that this is actually likely to be a win in practice.  For example, Java's built-in String.indexOf has long stuck to an O(mn) algorithm just because its constant factors in practice tend to be so dramatically better.  How much does this change affect the constant factor of the "good" case, as opposed to the worst case?

liach:
For now I do not really have good data to test these methods. Could you provide some big data if possible?

jbduncan:
Hi @liach. From what I can tell, I'm not sure if you actually need big data to test the performance of your new implementations of the methods.

In situations like this, people typically write micro-benchmarks with a framework like [JMH](http://openjdk.java.net/projects/code-tools/jmh/) to show how their solutions compare to existing ones speed-wise.

So my personal suggestion would be to investigate JMH if you're not already familiar with it, and then write a micro-benchmark between the current impls and yours, show us the results, and provide us with the source code so we can help you improve the benchmark in case it doesn't follow best practices. (But I admit that I wouldn't be the best person to ask about best practices, as I hardly know how to even write benchmarks with JMH or other frameworks.)

(I know that Google use Caliper instead as their framework for micro-benchmarking, but some people think it's a bit flawed and that JMH produces more accurate results, so I'd personally be inclined to use JMH for my own benchmarks. But that's a story for another time and for someone else to talk about, as I am not familiar with the details. :smile:)

liach:
Actually are benchmarks like [these](https://github.com/google/guava/blob/master/guava-tests/benchmark/com/google/common/primitives/UnsignedBytesBenchmark.java)?

jbduncan:
Yes, that's right. What you've shown is a link to one of Guava's Caliper benchmarks. However, I'd strongly recommend looking into **OpenJDK JMH** instead, because it's supposed to be much more reliable.

By my understanding, one of the reasons for this is that `for` loops are easy for the Oracle Hotspot JVM (and presumably other JVMs) to optimise away in non-obvious situations, and Caliper encourages its users to use `for` loops to benchmark code many times (this is done differently in JMH, which uses annotations and special optimization-defeating trickery), so Caliper benchmarks are more likely to report incorrect timings and thus are less reliable.

See https://groups.google.com/forum/#!msg/mechanical-sympathy/m4opvy4xq3U/7lY8x8SvHgwJ for more information.

liach:
Do I do in a separate project? I am not really clear about this.

jbduncan:
I think so? That's what the [front page](http://openjdk.java.net/projects/code-tools/jmh/) seems to say. But I'm really not the right person to ask, as I've never written a benchmark before.

PR #2637, opened 2016-11-11
specify executor

teaey:
Some times, we need events to be executed by specified executor.

### DIFF
--- a/guava/src/com/google/common/eventbus/SubscriberRegistry.java
+++ b/guava/src/com/google/common/eventbus/SubscriberRegistry.java
@@ -44,6 +44,8 @@ import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.concurrent.Executor;
+import javax.annotation.Nullable;
 import org.checkerframework.checker.nullness.compatqual.NullableDecl;
 
 /**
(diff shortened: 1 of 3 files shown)


guava/src/com/google/common/eventbus/SubscriberRegistry.java
```diff
@@ -174,6 +199,20 @@ void unregister(Object listener) {
     return methodsInListener;
   }
 
+  /**
+   * Returns all subscribers for the given listener grouped by the type of event they subscribe to.
+   */
+  private Multimap<Class<?>, Subscriber> findAllSubscribers(Object listener, Executor executor) {
```
liach: This can be just a `Map` since you do not need `Multimap` functionalities.

teaey:
@cpovirk @kluever @lowasser 
Can anybody accept this PR?

guava/src/com/google/common/eventbus/SubscriberRegistry.java
teaey: Keep the same semantics as the original method, eg:
```java
private Multimap<Class<?>, Subscriber> findAllSubscribers(Object listener)
```

teaey: @liach It's okay?

liach: Yeah, you can ignore my old comment.

teaey:
@liach 👍  Can you accept my PR now？

jbduncan:
@teaey You'll have to wait for someone from the Guava team to have a look at this PR first, as neither @liach nor myself are on the Guava team nor employees of Google ourselves.

Did you read the [API](https://github.com/google/guava/blob/master/CONTRIBUTING.md#api-changes) section of the contribution guidelines? :)

jbduncan:
@teaey That's correct.

It also means that Google employees have very high standards for the APIs they accept into Guava and follow a rigorous internal review process, so this PR may be rejected initially and they'll ask you to rewrite the code, or they may decide that the API doesn't really add anything and outright reject it.

But if either of those things do happen to you, try not to be discouraged, for it would actually be a very good thing as it would mean that Guava's API remains relatively easy to use and of high quality. :)

liach:
google doesn't really like additions. In comparison, bug fixes get accepted more easily.

nglorioso:
Realistically, we're unlikely to add functionality to EventBus at this point.

PR #2642, opened 2016-11-13
Added (last)indexOf methods to ObjectArrays

liach:


liach:
Do I need to add `@since` tags in the Javadoc?

cgdecker:
I don't think we want to add this to `ObjectArrays` because it's something you can do using `Arrays.asList(objectArray).(last)indexOf()`, and that's what we'd prefer you do.

liach:
Thanks for handling this pr!

PR #2652, opened 2016-11-19
Now that the CanIgnoreReturnValue annotation is used throughout the code the dependency is no longer optional

cheister:
In version 20.0 the errorprone annotations are used throughout the codebase so the error_prone_annotations dependency is not optional anymore.

### DIFF
--- a/guava/pom.xml
+++ b/guava/pom.xml
@@ -27,7 +27,7 @@
     <dependency>
       <groupId>com.google.errorprone</groupId>
       <artifactId>error_prone_annotations</artifactId>
-      <optional>true</optional><!-- needed only for annotations -->
+      <optional>false</optional>
     </dependency>
     <dependency>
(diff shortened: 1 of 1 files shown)


cheister:
I signed It!

cpovirk:
It should still be "optional" in the Maven sense (meaning that it needs to be present when compiling Guava but not when compiling projects that use Guava. What problem are you seeing?

cheister:
Ok.  I'm closing this PR so I can do more investigation.  When we updated to version 20.0 we started getting
```
Error:java: cannot access com.google.errorprone.annotations.CanIgnoreReturnValue
  class file for com.google.errorprone.annotations.CanIgnoreReturnValue not found
```
when compiling the project in IntelliJ.  The workaround is to explicitly add the error_prone_annotations dependency to our app but I'm not sure what part of our code is trying to use that annotation.

PR #2655, opened 2016-11-21
Fix for Optional(Present) equals/hashcode, when the reference is array.

StefanDekanski:


cpovirk:
I think we want to be consistent with normal `Object.equals` and `hashCode` here. I recognize that that's inconvenient in cases with arrays, but arrays are inconvenient in a lot of ways (including direct calls to `equals`), so the best course of action is typically to wrap them in lists as soon as possible.

In particular, the JDK's `Optional` behaves like `Object.equals` and `hashCode`, and we don't want to interfere with migrations to that type (or with migrations from nullable references to `Optional`).

StefanDekanski:
Fair points. Sometimes I really want to use arrays, to save some space (avoid Autoboxing). I guess I'll create a wrapper class for my arrays.

PR #2663, opened 2016-11-30
Fix #1900

### LINKED ISSUE #1900: MapSplitter should ignore key-value separator after first
Discussion here: https://groups.google.com/d/topic/guava-discuss/FuyKu3oF_PQ/discussion

Basically, after encountering the first key-value separator, it may be reasonable to just include any other separators that appear before the next entry separator in the value rather than throwing `IllegalArgumentException` as we do now.

Example:

``` java
MapSplitter splitter = Splitter.on(";").withKeyValueSeparator("=");

splitter.split("key=value;key2=value"); // normal
splitter.split("key=value;key2=value="); // currently throws IAE
// proposed change would make the second entry split to "key2" -> "value=" instead
```

jiafu1115:
it is one important issue to support most cases which value may contain splitter.


jiafu1115:
I signed it!

jiafu1115:
anyone can take a look?

liach:
@jiafu1115 Should this regard different types of splitters? Or add some method like `ignoreExtraSeparator()`?

jiafu1115:
@liach no. only change the inner behavior for map splitter.This is the simple way to meet requirement. WDYT？

liach:
There must be a reason why it throws an exception when detecting more than one separators. You should ping a project manager.

liach:
I think we need to ping the authors of `Splitter`. @lowasser

jiafu1115:
@liach Thanks for your reminder. We may need to wait and wait and wait......

jiafu1115:
@eamonnmcmanus Thanks for your change for title, it is more clear now.

jiafu1115:
@raghsriniv. can you help to review it? thanks

jiafu1115:
ask review

Issue #2667, opened 2016-12-01
ImmutableEnum{Set,Map} Collectors


michaelhixson:
Add collectors for `enum`-specialized `ImmutableSet` and `ImmutableMap`:

- `toImmutableEnumSet()`
- `toImmutableEnumMap(keyFunction, valueFunction)`
- `toImmutableEnumMap(keyFunction, valueFunction, mergeFunction)`

Providing these collectors in Guava would be a matter of convenience.  Users could implement functionally-equivalent collectors themselves, but I think most users would not do so.  I think `enum`-specialized collections are underused and that it is worthwhile to promote their usage and remove friction / barriers to entry.

For example, if a user has code like this:

```java
ImmutableSet<Month> months = stream.collect(ImmutableSet.toImmutableSet());
```

... then it would be cool if they could optimize their code like this:

```java
ImmutableSet<Month> months = stream.collect(ImmutableSet.toImmutableEnumSet());
```

... and it would be lame if they had to do this (it would be so lame that they wouldn't bother):

```java
ImmutableSet<Month> months = stream.collect(SomeHelper.toImmutableEnumSet(Month.class));

// Elsewhere... javadoc and tests omitted...
public static <E extends Enum<E>> Collector<E, ?, ImmutableSet<E>>
toImmutableEnumSet(Class<E> elementType) {
  Objects.requireNonNull(elementType);
  return Collectors.collectingAndThen(
      Collectors.toCollection(() -> EnumSet.noneOf(elementType)),
      Sets::immutableEnumSet);
}
```

tli2:
Hi: I have been working on this issue and discovered that the fauxverride tests are failing because the original equals code no longer works. Specifically, `K extends Enum<K>` as a type literal's equals method depends on the place where it is declared and its name. (The code can be found under sun.reflect.generics.reflectiveObjects.TypeVariableImpl's equals method).

I was able to fix the tests by simply comparing the result of `Type::getTypeName`, however, this is not correct as it is also sensible to the specific name of the generics declared.

Maybe this should be its own issue? I am not sure what the correct fix would be. Any input would be much appreciated!

michaelhixson:
@tli2 Oh, gross!  I hadn't considered that `ImmutableSortedSet` and friends might "inherit" these methods.

[...]

PR #2668, opened 2016-12-03
Code and test for #2667

tli2:
I came across this issue while browsing and thought I could take a shot at it. The collectors are implemented as wrappers around existing ones and I have written tests following the format of existing collector tests.

### DIFF
--- a/guava-tests/test/com/google/common/collect/ImmutableMapTest.java
+++ b/guava-tests/test/com/google/common/collect/ImmutableMapTest.java
@@ -646,6 +646,47 @@ public class ImmutableMapTest extends TestCase {
               mapEntry("three", 3),
               mapEntry("two", 2));
     }
+
+    public void testToImmutableEnumMap() {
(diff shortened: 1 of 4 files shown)


tli2:
I am stuck on a bug about Fauxverides for a while now (did not notice before I send the request). I have added the new methods to the `FauxverideShims` of the affected classes, and a manual test suggest that I have indeed deprecated the said methods, but I still cannot pass the tests. A debugging session suggests that the tests were able to find the fauxveridden methods but unable to recognize that the two are actually the same. `E extends Enum<E>` seems to be messing with the test's ability to compare. Specifically, the `Type` object under `TypeParameterSignature` seems to be an anonymous class when we do not extend `Enum<E>`, but would become `ParametrizedTypeImpl` when we do. A closer examination reveals that the `actualTypeArguments` field of `ParametrizedTypeImpl` is different, probably causing this problem. Here are screenshots from the debugger:

screen shot 2016-12-03 at 5 29 43 pm
screen shot 2016-12-03 at 5 30 33 pm

PR #2677, opened 2016-12-07
Add a MIME type MPEG-DASH (mpd files)

joankaradimov:
This format is becoming a thing. I couldn't find it in Guava's media types.


joankaradimov:
I signed it!

Issue #1563, opened 2014-10-31
Inconsistency between Ints and Longs in tryParse


gissuebot:
_[Original issue](https://code.google.com/p/guava-libraries/issues/detail?id=1563) created by **Maaartinus** on 2013-10-28 at 05:41 AM_

---

As requested in
https://mail.google.com/mail/ca/u/0/#label/8-java%2Fguava/141e7f32e0596002
I'm filing this as bug.

Currently, Ints.tryParse accepts non-ASCII digits, while Long.tryParse does not. This is both undocumented and inconsistent. The questions are
- Should non-ascii digits be accepted?
- If so, should even mixed numbers like DEVANAGARI_DIGIT_ONE + "" + BENGALI_DIGIT_TWO be accepted?

Note also that currently the fastest way to parse an Integer is via Longs.tryParse (this doesn't necessarily mean that accepting non-ASCII must cause a slowdown).

https://microbenchmarks.appspot.com/runs/f32a811c-aeb5-40ac-ae48-07b60256838f

pavlajak:
Hi,
I see no defect here. I checked and only Integer.parseInt(String s), or Long.parseLong(String s) can parse characters like DEVANAGARI_DIGIT_ONE. These methods can also parse a combination of several mixed non-ascii numbers.

None of the guava methods, like Ints.tryParse nor Longs.tryParse cannot parse any non-ascii characters. There is no inconsistency here either, since Ints.tryParse in the end calls Longs.tryParse.

So as far as I can tell, not a bug.

cpovirk:
This probably was broken back in 2013. We made some changes to at least one of these methods, probably within the past few years. It does seem to be fine now, as demonstrated by tests that we have:

    assertNull(Ints.tryParse("\u0662\u06f3"))
    assertNull(Longs.tryParse("\u0662\u06f3"))

PR #2680, opened 2016-12-12
Test case for #1563

pavlajak:
Added test cases for bug regarding incosistency between Ints.tryParse
and Longs.tryParse to show there is no defect.


pavlajak:
I signed it!

cpovirk:
Thanks. I think we're satisfied with the tests that we have:

    assertNull(Ints.tryParse("\u0662\u06f3"));
    assertNull(Longs.tryParse("\u0662\u06f3"));

I've gone back and closed the old bug. Thanks for checking into this.

Issue #1597, opened 2014-10-31
add a GWT serializer for the Range class


gissuebot:
_[Original issue](https://code.google.com/p/guava-libraries/issues/detail?id=1597) created by **ian.springer** on 2013-11-30 at 02:25 AM_

---

Add a GWT RPC serializer/deserializer for Ranges.

kluever:
@MyRealityCoding Are you still working on this?

MyRealityCoding:
I worked on it but it never went into a pull request :(

kluever:
We are dropping GWT support in ~6 months: https://groups.google.com/forum/#!topic/guava-discuss/ZLpBaYAtUl4

PR #2687, opened 2016-12-17
Add a custom gwt serializer for the Range object

ddebree:
Fixes issue #1597


cpovirk:
This is tough because GWT doesn't recognize, for a `Range<Foo>`, that it needs to know how to serialize a `Foo`. The fix is to introduce a dummy field into `Range` of type `C`. Possibly we could get away with doing this only in the GWT version -- maybe introducing a superclass of `Range` for just that purpose. I am hesitant to invest a lot of time in the deprecated GWT-RPC at this point, but I can probably make an exception. I'll see what I can do, though possibly not until the new year.

cpovirk:
(I think I have it working.)

Issue #2113, opened 2015-07-21
Random string generation API


lvxiang:
Hi, I’ve searched through Guava API, looking for a random string generator, but found none. And I’ve looked at [this thread](http://stackoverflow.com/questions/20782919/does-guava-have-a-method-to-generate-random-strings) which suggested using the <code>BaseEncoding</code> in some way for your purposes. I’m not a big fan of this solution. For one thing you have to know BASE64 very well so as not to make any mistakes, such as forgetting to omit paddings. Moreover, it’s not an instant solution as most APIs usually do, you have to think twice to come up with the idea. On the other hand, the solution is too detailed to allow flexibility, consider the following cases:
 * What if some chars are not supposed to appear in the string. I might want to ignore ‘I’ and ‘l’ for they look very alike.
 * There is no way to generate a unicode string.
 * What if I want all letters in their capital forms?

You have to write your own code to cover the cases above. There are other issues to consider:
 * There’s definitely a better solution with better performance than using BaseEncoding.
 * There’s no formal proof to the randomness of strings generated.

Personally, I often come across requirements for random strings in various formats, and I see a good reason for Guava to provide dedicated APIs for generating random strings. The fluid-style can perfectly fit into this situation. Please let me know your thoughts on this issue.

thiagokronig:
See
http://stackoverflow.com/questions/41107/how-to-generate-a-random-alpha-numeric-string

Construct a BigInteger with N bits randomly obtained from a Random source
and encode it in Base-32.

Cheers,

Thiago Kronig

lvxiang:
@thiagokronig checked that before, still not solving all the problems

Maaartinus:
[This answer](http://stackoverflow.com/a/157202/581205) uses no `BigInteger` and is pretty general. It also seems to be optimal (except for using `StringBuilder` where `char[]` would do).

lvxiang:
@Maaartinus I don't think it's general enough as you have to redefine "AB" constantly. Except for StringBuilder, it might not be optimal in some cases, i.e, if all you want is a string of decimal digits, the following code might be faster:

    Random rng = new Random();
    char[] str = new char[length];
    for(int i = 0; i < length; i ++)
        str[i] = (char) ('0' + rng.nextInt(10));
    return new String(str);

ogregoire:
I agree with @lvxiang in the sense that usually people have a set of constraints and want to generate strings according to those sets of constraints, without having to redefine them everywhere.

[...]

PR #2707, opened 2017-01-10
Implementation of #2113

ogregoire:
This is a standard and clean implementation for a random string generation API.

    RandomStringGenerator generator = new RandomStringGenerator.Builder()
      .withLowercaseAsciiLetters()
      .withAsciiDigits(1)
      .withCharacters("!@#")
      .build(10);
    for (int i = 0; i < 100; i++) {
      String generated = generator.next();
      // generated is guaranteed to have 10 characters of which at least one digit
    }

There are a few choice that guided my implementation:

1. It must implement `Supplier`.
2. The configuration must be in the builder to allow a clean and simple API. The builder is necessary instead of a `Splitter`-like or `CharMatcher`-like API because we don't know which characters to provide by default and providing those methods at starting points make it overly complex.
3. A user-provided random had to be providable.
4. The implementation had to be thread-safe.
5. I discarded the idea of a variable length String generator because that would have been very difficult to implement, and is overkill. People usually want to generate `String`s of the same length. If not, one can still keep the `Builder` and create several generators.
6. There had to be a "similar characters" discarder.
7. `"aaab"` would have to give 3 times more `a` than `b` by default, and a method had to exist to leave the possibility to give each char the same weight.

### DIFF
--- a/guava/src/com/google/common/base/RandomStringGenerator.java
+++ b/guava/src/com/google/common/base/RandomStringGenerator.java
@@ -1,0 +1,433 @@
+/*
+ * Copyright 2017 The Guava Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.google.common.base;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+
+import com.google.common.collect.ImmutableMultiset;
+import com.google.common.collect.Multiset;
+import com.google.common.primitives.Chars;
+
+import java.util.HashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Random;
+import java.util.Set;
+
+import javax.annotation.concurrent.ThreadSafe;
+
+/**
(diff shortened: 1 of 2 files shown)


guava/src/com/google/common/base/RandomStringGenerator.java
```diff
@@ -0,0 +1,433 @@
...
+ *       .build();
+ *   for (int i = 0; i &lt; 10; i++) {
+ *     System.out.println(generator.next());
+ *   }
+ * }
+ * </pre>
+ *
+ * <p>
+ * By default, no check is done on the quantity a character is present, so if a generator has to
+ * look through {@code "aaab"}, {@code 'a'} has three times the chances of {@code 'b'} to appear. If
```
HaasJona: To me, it's not completely clear what "look through aaab" means.

guava/src/com/google/common/base/RandomStringGenerator.java
```diff
@@ -0,0 +1,433 @@
...
+      for (int i = 0; i < range.length; i++, from++) {
+        range[i] = from;
+      }
+      return new String(range);
+    }
+
+    private Builder addCharacters(String characters, Integer minimumQuantity) {
+      checkArgument(minimumQuantity == null || minimumQuantity > 0,
+          "minimumQuantity (%s) must be strictly positive", minimumQuantity);
+      chars.put(characters, minimumQuantity);
```
HaasJona: What happens if the map already contains the characters String?

guava/src/com/google/common/base/RandomStringGenerator.java
```diff
@@ -0,0 +1,433 @@
...
+    private void checkBuild(int size) {
+      checkArgument(size >= countRequiredCharacters,
+          "size (%s) is too low compared to the minimum number of characters (%s)",
+          size, countRequiredCharacters);
+      checkState(!chars.isEmpty(), "No character definition was added to this builder");
+
+      if (avoidSimilarAsciiCharacters) {
+        for (String key : chars.keySet()) {
+          checkState(!SIMILAR_ASCII_CHARS.removeFrom(key).isEmpty(),
+              "withCharacters(%s) and avoidSimilarAsciiCharacters() cancel each other and may not be called together");
```
HaasJona: %s?

Issue #2714, opened 2017-01-15
ReverseConverter, ConverterComposition, & FunctionBasedConverter classes implement Serializable without transient/serialized member variables


petermalone:
For example, line 146 of com.google.common.base.Converter:

final Converter<A, B> original;

Should be:

final transient Converter<A, B> original;

petermalone:
Not an issue.

PR #2715, opened 2017-01-15
Fix for Issue 2714: https://github.com/google/guava/issues/2714

petermalone:
ReverseConverter, ConverterComposition, & FunctionBasedConverter classes implement Serializable without transient/serialized member variables

### DIFF
--- a/guava/src/com/google/common/base/Converter.java
+++ b/guava/src/com/google/common/base/Converter.java
@@ -241,7 +241,7 @@ public abstract class Converter<A, B> implements Function<A, B> {
 
   private static final class ReverseConverter<A, B> extends Converter<B, A>
       implements Serializable {
-    final Converter<A, B> original;
+    final transient Converter<A, B> original;
 
     ReverseConverter(Converter<A, B> original) {
       this.original = original;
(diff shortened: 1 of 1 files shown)


lowasser:
I'm not sure I follow what you're going for with this.  Transient variables aren't serialized; these fields appear to need to be serialized...?

petermalone:
I signed the CLA.

lowasser - it's in response to this: http://cwe.mitre.org/data/definitions/594.html & squid:S1948

"Fields in a Serializable class must themselves be either Serializable or transient even if the class is never explicitly serialized or deserialized. That's because under load, most J2EE application frameworks flush objects to disk, and an allegedly Serializable object with non-transient, non-serializable data members could cause program crashes, and open the door to attackers."

petermalone:
Contributor License Agreement (CLA),I signed it!

lowasser:
That, too, seems working as intended to me.  The documented behavior is that the generated Converters are serializable if the original converter is serializable, which is accomplished with the status quo: if the underlying Converter is Serializable the serialization will work, otherwise you will get an exception when serialization is attempted.

Marking these fields transient will cause a NullPointerException if the object is serialized and deserialized, even if the original object was serializable.  (If our tests pass with your change, that probably reflects a problem with the tests?)

lowasser:
Or, rather, to be specific: the converter, if serialized and deserialized while the fields are transient, will throw an NPE on an attempt to use them for conversions.

petermalone:
Understood - if you're fine with this, let me close the pull request.  

Thanks so much for the speedy response.

PR #2734, opened 2017-02-06
Introduce ImmutableMap#mergeMapsCollector

numero-six:
Introduce a Collector to ease merging of Maps with java stream API.

Given (random Maps implementations used to ensure collector is generic):
```java
HashMap<String, Integer> foo = new LinkedHashMap<>();
foo.put("one", 1);
foo.put("two", 2);
NavigableMap<String, Integer> bar = new TreeMap<>();
bar.put("three", 3);
bar.put("four", 4);
Stream<Map<String, Integer>> stream = ImmutableSet.of(foo, bar).stream();
```

This code:
```java
ImmutableMap<String, Integer> collectedTheOldWay = stream
        .map(Map::entrySet)
        .flatMap(Set::stream)
        .collect(ImmutableMap.toImmutableMap(Map.Entry::getKey, Map.Entry::getValue));
```
can now be done with:
```java
ImmutableMap<String, Integer> collected = stream.collect(ImmutableMap.mergeMapsCollector());
```

### DIFF
--- a/guava-tests/test/com/google/common/collect/ImmutableMapTest.java
+++ b/guava-tests/test/com/google/common/collect/ImmutableMapTest.java
@@ -646,6 +646,21 @@ public class ImmutableMapTest extends TestCase {
               mapEntry("three", 3),
               mapEntry("two", 2));
     }
+
+    public void testMergeMapsCollector() {
+      Collector<? super Map<String, Integer>, ?, ImmutableMap<String, Integer>> collector =
+          ImmutableMap.mergeMapsCollector();
+      Equivalence<ImmutableMap<String, Integer>> equivalence =
+          Equivalence.equals()
+              .<Entry<String, Integer>>pairwise()
+              .onResultOf(ImmutableMap::entrySet);
+      CollectorTester.of(collector, equivalence)
+          .expectCollects(
+              ImmutableMap.of("one", 1, "two", 2, "three", 3, "four", 4),
+              ImmutableMap.of("one", 1, "two", 2),
+              ImmutableMap.of("three", 3, "four", 4)
+          );
+    }
   }
 
   public void testNullGet() {
--- a/guava/src/com/google/common/collect/ImmutableMap.java
+++ b/guava/src/com/google/common/collect/ImmutableMap.java
@@ -103,6 +103,18 @@ public abstract class ImmutableMap<K, V> implements Map<K, V>, Serializable {
   }
 
   /**
+   * Returns a {@link Collector} that accumulates elements into an {@link ImmutableMap} taking
+   * all the entries of all the encountered {@link Map}.
+   */
+  public static <K, V> Collector<? super Map<K, V>, ?, ImmutableMap<K, V>> mergeMapsCollector() {
+    return Collector.of(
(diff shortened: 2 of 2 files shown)


lowasser:
Hi!

As this is a new API, it's going to have to go through a process and some internal review before we even start thinking about the code.  (There's some explanation of this [here](https://github.com/google/guava/wiki/HowToContribute#code-contributions).)

As it stands, there's a few things we'd need to get together before we approved this feature:
- Some research on how common the need for this API is, ideally including several different places in real existing code that this API would have been helpful.  (This will probably end up needing to be done by a Googler, just because we have a large, easily-searchable codebase.)
- For each of those real uses, a discussion of whether or not this API is really the best way to help those users.  For this feature, I'd wonder why the maps were created separately in the first place?  Do people actually have an unbounded number of maps, or do they have a fixed number of maps that need to be combined?  For example, I'd tend to rewrite the example you gave to the stream-free `new ImmutableMap.Builder<>().putAll(foo).putAll(bar).build()`, which is very short and simple.

numero-six:
Hi,

> For this feature, I'd wonder why the maps were created separately in the first place?

That is just for the example.

> Do people actually have an unbounded number of maps, or do they have a fixed number of maps that need to be combined?

Of course, the actual purpose is for an unbounded number of maps. For a fixed number of maps, your code is better.

My cases are usually like the Map is generated by a `Stream#map` call. Let's take a new example (imperative/no-stream-style first):

```java
ImmutableMap.Builder<MyKey, MyValue> allKnownValues = ImmutableMap.builder();
for (Foo foo : foos) {
    Bar bar = this.convertFooToBar(foo);
    ImmutableMap<MyKey, MyValue> mapForFoo = bar.getTheMap();
    allKnownValues.putAll(mapForFoo);
}
return allKnownValues.build();
```

With the stream API, should be:

```java
return foos.stream()
        .map(this::convertFooToBar)
        .map(Bar::getTheMap) // gives the Stream<Map<MyKey, MyValue>>
        .map(Map::entrySet)
        .flatMap(Set::stream)
        .collect(ImmutableMap.toImmutableMap(Map.Entry::getKey, Map.Entry::getValue));
```

With the suggested new API:

```java
return foos.stream()
        .map(this::convertFooToBar)
        .map(Bar::getTheMap)
        .collect(ImmutableMap.mergeMapsCollector());
```

I think this kind of creating a Map by dynamically filling it with dynamically computed maps is frequent (hence, the `ImmutableMap.Builder#putAll(Map)`). I won't discuss the need for this API because I think Guava maintainers have a much better point of view/experience for that than mine. My experience is just that people will favor 1st code (imperative style) to 2nd code which is too verbose and quite cryptic for the non-stream-API-fluent-developer and I think the 3rd one is much more functional-style (ie: declarative about the expected result), but it's just based on nothing more than my day-to-day experience.

Also, maybe `mergeMaps` is not the suitable term but I'm not a native english-speaker) and a `mergeOperator` could be passed.

cpovirk:
Mass-closing PRs without an accepted [CLA](https://cla.developers.google.com/about). You're welcome to reopen if you get the check to pass. Alternatively, feel free to open a plain issue to describe the change you'd like to see made.

PR #2761, opened 2017-03-13
Making the BloomFilter thread-safe & lock-free

### LINKED ISSUE #2748: Any interest in a PR making BloomFilter lock-free?
As part of my work at LinkedIn, I've created a lock-free version of Guava's `BloomFilter` class (technically, a lock-free version of the `BitArray` class inside `BloomFilterStrategies`). **Would Guava upstream be interested in seeing a PR for this change?** The PR would naturally come with extra tests.

### Short overview

Our use-case involves a hundred+ threads doing a total of millions of read QPS on the bloom filter with ~20 QPS of writes. So I wrapped the bloom filter with a read-write lock. This wasn't performant enough though on the read side; after lots of benchmarks and analysis, the root cause was the write lock; even removing the read locks (which of course introduces race conditions, but was done for the sake of a benchmark) made no difference for read path perf... but removing both read _and_ write locks improved read perf greatly (but it would of course be terribly unsafe). My guess is that it was exclusive access to the cache line that stored the lock data that caused excessive CPU cache bus traffic, but I don't have data to back that up.

Replacing the internals of the BitArray so that it used an `AtomicLongArray` for the `data` member along with compare-and-swap usage (plus an `AtomicLong` for the `bitcount` member) improved perf substantially.

This is all on a modern server CPU with 24 logical CPUs (12 cores) + HotSpot Java 8 JVM.

Valloric:
Previously, the BloomFilter wasn't thread-safe and required external locking to ensure safety. Now, it's thread-safe and lock-free through the use of atomics and compare-and-swap.

This PR introduces **no** API changes beyond an extra `@ThreadSafe` annotation on the BloomFilter class. It should also be entirely backwards (and forwards) compatible with the serialization format because that too isn't being changed. 

Please extend extra scrutiny to the `LockFreeBitArray.putAll()` method because it's not present in our internal fork of the BloomFilter class and has thus not gone through our integ tests or has seen prod (I wrote it for this PR).

Fixes #2748.

### DIFF
--- a/guava/src/com/google/common/hash/BloomFilterStrategies.java
+++ b/guava/src/com/google/common/hash/BloomFilterStrategies.java
@@ -21,6 +21,8 @@ import com.google.common.primitives.Ints;
 import com.google.common.primitives.Longs;
 import java.math.RoundingMode;
 import java.util.Arrays;
+import java.util.concurrent.atomic.AtomicLongArray;
+import java.util.concurrent.atomic.LongAdder;
 import javax.annotation.Nullable;
 
 /**
@@ -44,7 +46,7 @@ enum BloomFilterStrategies implements BloomFilter.Strategy {
   MURMUR128_MITZ_32() {
     @Override
     public <T> boolean put(
-        T object, Funnel<? super T> funnel, int numHashFunctions, BitArray bits) {
+        T object, Funnel<? super T> funnel, int numHashFunctions, LockFreeBitArray bits) {
       long bitSize = bits.bitSize();
       long hash64 = Hashing.murmur3_128().hashObject(object, funnel).asLong();
       int hash1 = (int) hash64;
@@ -64,7 +66,7 @@ enum BloomFilterStrategies implements BloomFilter.Strategy {
 
     @Override
     public <T> boolean mightContain(
-        T object, Funnel<? super T> funnel, int numHashFunctions, BitArray bits) {
+        T object, Funnel<? super T> funnel, int numHashFunctions, LockFreeBitArray bits) {
       long bitSize = bits.bitSize();
       long hash64 = Hashing.murmur3_128().hashObject(object, funnel).asLong();
       int hash1 = (int) hash64;
@@ -92,7 +94,7 @@ enum BloomFilterStrategies implements BloomFilter.Strategy {
   MURMUR128_MITZ_64() {
     @Override
     public <T> boolean put(
-        T object, Funnel<? super T> funnel, int numHashFunctions, BitArray bits) {
+        T object, Funnel<? super T> funnel, int numHashFunctions, LockFreeBitArray bits) {
       long bitSize = bits.bitSize();
       byte[] bytes = Hashing.murmur3_128().hashObject(object, funnel).getBytesInternal();
       long hash1 = lowerEight(bytes);
@@ -110,7 +112,7 @@ enum BloomFilterStrategies implements BloomFilter.Strategy {
 
     @Override
     public <T> boolean mightContain(
-        T object, Funnel<? super T> funnel, int numHashFunctions, BitArray bits) {
+        T object, Funnel<? super T> funnel, int numHashFunctions, LockFreeBitArray bits) {
       long bitSize = bits.bitSize();
       byte[] bytes = Hashing.murmur3_128().hashObject(object, funnel).getBytesInternal();
       long hash1 = lowerEight(bytes);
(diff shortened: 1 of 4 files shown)


guava/src/com/google/common/hash/BloomFilterStrategies.java
```diff
@@ -138,80 +140,149 @@
...
+        //
+        // If the promoted type of the left-hand operand is long, then only the
+        // six lowest-order bits of the right-hand operand are used as the shift
+        // distance. It is as if the right-hand operand were subjected to a
+        // bitwise logical AND operator & (§15.22.1) with the mask value 0x3f
+        // (0b111111). The shift distance actually used is therefore always in the
+        // range 0 to 63, inclusive.
+        // More details:
+        //   http://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.19
+        long oldValue = data.getAndUpdate(longIndex, x -> x | (1L << bitIndex));
```
ben-manes: This creates garbage (lambda captures `bitIndex`). I prefer minimizing allocations in data structures.

guava/src/com/google/common/hash/BloomFilterStrategies.java
```diff
@@ -138,80 +140,149 @@
...
+
+        // Only if we ended up setting the bit do we run
+        // bitCount.incrementAndGet(). Note that the first get() in the
+        // function could have returned false, then another thread could have
+        // set the bit before we ran getAndUpdate. That's why we need to check
+        // what the oldValue was for our compare-and-swap operation.
+        if ((oldValue & (1L << bitIndex)) == 0) {
```
ben-manes: I'd pull `1L << bitIndex` into a local field called `mask` for both usages.

guava/src/com/google/common/hash/BloomFilterStrategies.java
```diff
@@ -138,80 +140,149 @@
...
+  /**
+   * Models a lock-free array of bits.
+   *
+   * We use this instead of java.util.BitSet because we need access to the
+   * array of longs and we need compare-and-swap.
+   */
+  static final class LockFreeBitArray {
+    private static final int LONG_ADDRESSABLE_BITS = 6;
+    final AtomicLongArray data;
+    private AtomicLong bitCount;
```
ben-manes: Arguably a `LongAdder` to avoid a hotspot.

Valloric: Not a bad idea! I'll try that, but I'll have to benchmark it first.

guava/src/com/google/common/hash/BloomFilterStrategies.java
Valloric: I'll benchmark a version with a separate non-lambda function.

guava/src/com/google/common/hash/BloomFilterStrategies.java
Valloric: Makes sense, will do.

Valloric:
>To strengthen your argument, you should probably add a benchmark to show the performance impact for reads and writes in a single threaded case. See Guava's Caliper benchmarks.

I have JMH benchmarks (see this email for [Caliper vs JMH](https://groups.google.com/forum/#!msg/mechanical-sympathy/m4opvy4xq3U/7lY8x8SvHgwJ)) set up for our internal fork which is largely the same except the serialization parts have been ripped out, other hash strategies have been added etc. Perf using the standard murmur3 strategy should be identical. 

I can update that and post numbers. It's not ideal, but it's the best I have without going through the massive hassle of supporting JMH benchmarks for guava (I wouldn't trust Caliper).

Valloric:
@ben-manes 
OK, I'm doing benchmarks first for the suggestions you've made. [Here's my benchmark code](https://gist.github.com/Valloric/f57aa7726947250065dc1a878fc33888), for reference only.

Benchmark results I currently have:

```
NOTE: "Batch" means 10k puts, NOT one put!
NOTE: The units aren't the same across benchmarks; some use nanos, some micros.

Original PR

Benchmark                                                        Mode  Cnt      Score       Error  Units
LockFreeBloomFilterBenchmark.measureContainMultiThread           avgt    4    511.674 ±    17.661  ns/op
LockFreeBloomFilterBenchmark.measureContainSingleThread          avgt    4    490.260 ±     4.765  ns/op
LockFreeBloomFilterBenchmark.measurePutBatchRandomsMultiThread   avgt    4  15080.691 ± 11069.316  us/op
LockFreeBloomFilterBenchmark.measurePutBatchRandomsSingleThread  avgt    4  10408.925 ±   203.714  us/op
LockFreeBloomFilterBenchmark.measurePutSameSingle                avgt    4    624.769 ±    39.450  ns/op

Original PR + LongAdder

Benchmark                                                        Mode  Cnt      Score     Error  Units
LockFreeBloomFilterBenchmark.measureContainMultiThread           avgt    4    494.897 ±  21.302  ns/op
LockFreeBloomFilterBenchmark.measureContainSingleThread          avgt    4    489.003 ±  14.366  ns/op
LockFreeBloomFilterBenchmark.measurePutBatchRandomsMultiThread   avgt    4  12040.170 ± 312.177  us/op
LockFreeBloomFilterBenchmark.measurePutBatchRandomsSingleThread  avgt    4  10927.175 ± 381.793  us/op
LockFreeBloomFilterBenchmark.measurePutSameSingle                avgt    4    634.361 ±  75.430  ns/op
```

As expected, LongAdder helps in the multi-threaded Put scenario. So I'll make that change. I'll run a similar benchmark for the lambda change.

guava/src/com/google/common/hash/BloomFilterStrategies.java
Valloric: Done.

guava/src/com/google/common/hash/BloomFilterStrategies.java
Valloric: Doing some research it seems the JVM will inline small lambdas with simple capture state (and a `long` field is about as simple as it comes) starting with Java 8 which has better escape analysis for this very use-case.

I can also confirm that in prod for us this code gets called half a million (or more) times a second and we saw no increase in GC pauses or GC frequency. 

So I personally wouldn't worry about it. I did benchmark a version where instead of using `getAndUpdate` and passing a lambda, I inline the loop from `getAndUpdate` and using a raw `compareAndSwap` call. The results are the same. This of course doesn't confirm that garbage is/isn't being created though, just that the time perf stays the same.

Given the above info, how strongly do you feel about the lambda vs inlined CAS call? The lambda approach is much more readable code, so I'd prefer to stick with it.

ben-manes: That's nifty! I try to avoid relying on escape analysis, but I have optimized to make it easier to be taken. This isn't critical to me, especially since the hashing creates garbage too. Its nice when ES works and simplifies code.

But any of these recommendations are strictly my own. I have no influence on the Guava team or having this merged.

Valloric:
And here are the benchmarks for the original code, without any lock-free stuff. Note that the benchmark names for the original code still have "lockfree" in them but the code is not actually lock-free.

```
Original PR + LongAdder

Benchmark                                                        Mode  Cnt      Score     Error  Units
LockFreeBloomFilterBenchmark.measureContainSingleThread          avgt    4    489.003 ±  14.366  ns/op
LockFreeBloomFilterBenchmark.measurePutBatchRandomsSingleThread  avgt    4  10927.175 ± 381.793  us/op
LockFreeBloomFilterBenchmark.measurePutSameSingle                avgt    4    634.361 ±  75.430  ns/op

Without PR (no atomics or locking, original code)

Benchmark                                                        Mode  Cnt      Score      Error  Units
LockFreeBloomFilterBenchmark.measureContainSingleThread          avgt    4    474.787 ±   10.863  ns/op
LockFreeBloomFilterBenchmark.measurePutBatchRandomsSingleThread  avgt    4   8958.676 ±   56.457  us/op
LockFreeBloomFilterBenchmark.measurePutSameSingle                avgt    4    624.619 ±   37.565  ns/op
```

As expected, the puts are slightly slower (about 20%) in the lock-free version. The read perf is the same because atomic reads are free on x86 (and x86-64). It's hard for me to imagine a realistic single-threaded use-case for a bloom filter; even then, since the read path is what almost always matters in a bloom filter and that sees no slowdown, I think a 20% hit on the single-threaded put path is acceptable.

Valloric:
@lowasser Any thoughts on this PR so far?

guava/src/com/google/common/hash/BloomFilterStrategies.java
```diff
@@ -138,80 +140,151 @@
...
+        // (0b111111). The shift distance actually used is therefore always in the
+        // range 0 to 63, inclusive.
+        // More details:
+        //   http://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.19
+        long mask = 1L << bitIndex;
+
+        // This performs a compare-and-swap operation using the corresponding
+        // platform CPU instruction. More details:
+        //     https://en.wikipedia.org/wiki/Compare-and-swap
+        long oldValue = data.getAndUpdate(longIndex, x -> x | mask);
```
lowasser: I think I would prefer this to be a Java 7 compatible implementation, so we don't need gratuitous diffs between the Java 7 branch and the Java 8 branch.

Valloric: Excellent point I hadn't considered; removed the lambda by inlining `getAndUpdate`.

Valloric:
@lowasser @ben-manes The latest code addresses all issues that have been brought up so far.

guava/src/com/google/common/hash/BloomFilterStrategies.java
```diff
@@ -138,80 +140,159 @@
...
+        // range 0 to 63, inclusive.
+        // More details:
+        //   http://docs.oracle.com/javase/specs/jls/se7/html/jls-15.html#jls-15.19
+        long mask = 1L << bitIndex;
+
+        long oldValue, newValue;
+        do {
+          oldValue = data.get(longIndex);
+          newValue = oldValue | mask;
+        // This performs a compare-and-swap operation using the corresponding
```
lowasser: I'd be kind of tempted to check if (oldValue == newValue) here instead of CASing a variable to itself and then doing the check after?

Valloric: Hm, we could do that, but note that it's a trade-off. It would optimize the uncontended path while making the contended path slower because of the extra branch (in theory; branch prediction yadda yadda). That does sound like a reasonable trade-off to me though, so I'll do that and run a benchmark.

Valloric:
@lowasser @ben-manes 

OK, benchmarks with the extra IF branch. Note that I'm including another set of benchmarks for the same code pairs, but the second set is with a much smaller bloom filter: 10k puts on a BF with 50k expected insertions vs the original 10k puts on a BF with 150M expected insertions. I added this to look at cases with more thread contention.

```
Original PR + LongAdder + no-lambda

Benchmark                                                        Mode  Cnt      Score     Error  Units
LockFreeBloomFilterBenchmark.measureContainMultiThread           avgt    4    487.889 ±  14.100  ns/op
LockFreeBloomFilterBenchmark.measureContainSingleThread          avgt    4    489.317 ±   4.928  ns/op
LockFreeBloomFilterBenchmark.measurePutBatchRandomsMultiThread   avgt    4  11804.317 ±  39.740  us/op
LockFreeBloomFilterBenchmark.measurePutBatchRandomsSingleThread  avgt    4  11673.301 ± 407.479  us/op
LockFreeBloomFilterBenchmark.measurePutSameSingle                avgt    4    618.878 ±  38.190  ns/op

Original PR + LongAdder + no-lambda + extra IF in retry loop

Benchmark                                                        Mode  Cnt      Score     Error  Units
LockFreeBloomFilterBenchmark.measureContainMultiThread           avgt    4    495.849 ±  72.557  ns/op
LockFreeBloomFilterBenchmark.measureContainSingleThread          avgt    4    490.722 ±   4.135  ns/op
LockFreeBloomFilterBenchmark.measurePutBatchRandomsMultiThread   avgt    4  11860.600 ± 142.482  us/op
LockFreeBloomFilterBenchmark.measurePutBatchRandomsSingleThread  avgt    4  11234.194 ± 257.156  us/op
LockFreeBloomFilterBenchmark.measurePutSameSingle                avgt    4    630.437 ±  51.002  ns/op

=== SMALLER BENCH BLOOM FILTER! ===

Original PR + LongAdder + no-lambda

Benchmark                                                        Mode  Cnt     Score     Error  Units
LockFreeBloomFilterBenchmark.measureContainMultiThread           avgt    4   523.716 ±  76.739  ns/op
LockFreeBloomFilterBenchmark.measureContainSingleThread          avgt    4   494.077 ±   6.449  ns/op
LockFreeBloomFilterBenchmark.measurePutBatchRandomsMultiThread   avgt    4  6541.044 ± 112.961  us/op
LockFreeBloomFilterBenchmark.measurePutBatchRandomsSingleThread  avgt    4  6318.562 ± 179.519  us/op
LockFreeBloomFilterBenchmark.measurePutSameSingle                avgt    4   616.819 ±   4.884  ns/op

Original PR + LongAdder + no-lambda + extra IF in retry loop

Benchmark                                                        Mode  Cnt     Score      Error  Units
LockFreeBloomFilterBenchmark.measureContainMultiThread           avgt    4   529.688 ±   78.308  ns/op
LockFreeBloomFilterBenchmark.measureContainSingleThread          avgt    4   501.248 ±    5.284  ns/op
LockFreeBloomFilterBenchmark.measurePutBatchRandomsMultiThread   avgt    4  6574.708 ± 1121.399  us/op
LockFreeBloomFilterBenchmark.measurePutBatchRandomsSingleThread  avgt    4  6241.094 ±  275.969  us/op
LockFreeBloomFilterBenchmark.measurePutSameSingle                avgt    4   618.840 ±   20.353  ns/op
```

The results are the same with and without the extra IF. Should I add it then or not? I'm leaning "no" because it doesn't seem to be paying off.

Valloric:
Also, the extra IF seems to be introducing extra variance in the multi-threaded runs, probably because of branch prediction failures.

ben-manes:
Note that you are using a uniform distribution. The `if` check would be helpful if the entry had already been added to avoid unnecessary CASes. A Zipf distribution ([example](https://github.com/ben-manes/caffeine/blob/master/caffeine/src/jmh/java/com/github/benmanes/caffeine/cache/GetPutBenchmark.java)) might show beneficial results.

Valloric:
@ben-manes That's convincing enough for me. Perf is flat on a uniform distribution and the difference in variance is likely to be noise, and even if it isn't, it probably doesn't matter.

Added the IF.

guava/src/com/google/common/hash/BloomFilterStrategies.java
```diff
@@ -138,80 +140,165 @@
...
+          other.data.length());
+      for (int i = 0; i < data.length(); i++) {
+        long otherLong = other.data.get(i);
+
+        long ourLongOld, ourLongNew;
+        do {
+          ourLongOld = data.get(i);
+          ourLongNew = ourLongOld | otherLong;
+          if (ourLongOld == ourLongNew) {
+            continue;
```
ben-manes: isn't this supposed to be a `break`?

Valloric: Oops. :) Good catch!

Valloric: What I wanted was to continue from the outer loop, but somehow forgot I was in the inner one. We still perform the bitcount ops unnecessarily though. I could use a goto to avoid this, but I'd rather not.

Valloric: Added an `earlyEscape` boolean so we can avoid the bitcounts.

Valloric:
More data:

```
Original PR + LongAdder + no-lambda + extra IF in retry loop

Benchmark                                                        Mode  Cnt     Score     Error  Units
LockFreeBloomFilterBenchmark.measureContainMultiThread           avgt    4   500.731 ±  68.968  ns/op
LockFreeBloomFilterBenchmark.measureContainSingleThread          avgt    4   490.522 ±   5.937  ns/op
LockFreeBloomFilterBenchmark.measurePutBatchRandomsMultiThread   avgt    4  6530.204 ± 114.009  us/op
LockFreeBloomFilterBenchmark.measurePutBatchRandomsSingleThread  avgt    4  6267.970 ± 176.621  us/op
LockFreeBloomFilterBenchmark.measurePutSameSingle                avgt    4   615.487 ±   9.207  ns/op
```

I did another run of the benchmark with the same smaller BF config and the large variance from the previous run went away. So it's noise, making the IF a safe pick. It's either better with a zipfian distribution, or no worse.

Valloric:
Note that with the new IF in the retry loop, we can (and I have) now remove this IF check:

```java
if ((oldValue & mask) == 0) {
  bitCount.increment();
  return true;
 }
```

We just always run the body now.

Valloric:
@lowasser @ben-manes 

Code updated to address all issues raised so far.

Valloric:
> Note that my approval carries no weight towards integration.

I know, but it carries weight with me in that I'm more confident this is working well. :)

Valloric:
@lowasser Friendly ping. :)

Valloric:
@lowasser One more ping...

lowasser:
LGTM, I'll try to get this merged internally.

Valloric:
> LGTM, I'll try to get this merged internally.

Thank you! :)

Valloric:
Resolved recently introduced merge conflicts.

Valloric:
@lowasser Any updates? It's been 2 weeks since this PR reached a stable state...

lowasser:
We ran into the trouble that LongAdder isn't present in Java 7.  We have a
fork of it in common.cache, we just ran into the issue that we'll need
another copy for the Java 7 branch in common.hash.

Valloric:
@lowasser Can I help in any way? If I can make your life easier by changing this PR, I'd be happy to do so.

Valloric:
@lowasser @artityod Any updates? This has been ready to merge for a month and a half now...

lowasser:
It's in the review process internally; we've been working on the comments significantly.

Valloric:
@lowasser Thanks for the update! Glad to see this hasn't fallen by the wayside. :)

cpovirk:
https://github.com/google/guava/pull/2806

adityamohan93:
Hello @Valloric , @ben-manes and @lowasser 
I needed a quick suggestion on a use case for faster writes to this bloom filter implementation. I have 10 sets each with around 100,000 Long integers and I want to add them to a bloom filter in the fastest way possible. As this filter is thread safe, can writing all 10 sets in parallel cause any issue in terms of data corruption or false positive rate?

ben-manes:
It should be safe since the api only allows accumulation, so you are only subject to ensuring a initial good sizing to reach the desired fp rate. In some cases it could be faster to union the bloom filters, but this heavily depending on the table vs update sizes, thread counts, etc. For example you might use `putAll(bloomFilter)` instead of individual `put(long)` or you might query N filters so that the writes are independent. "fastest way possible" is very use-case specific

adityamohan93:
Thanks @ben-manes for the prompt response. I will go ahead with `put(long)` operation as the inputs are from hash sets whose sizes are known. I will also check if pre-computing bloom filters from the hash sets and then merging them using `putAll(bloomFilter)` would be faster. My concern with merging bloom filters is the unpredictability for the merged fp rate.

PR #2776, opened 2017-04-04
Promote HostAndPort out of Beta.

swankjesse:
It's all grown up.

### DIFF
--- a/guava/src/com/google/common/net/HostAndPort.java
+++ b/guava/src/com/google/common/net/HostAndPort.java
@@ -18,7 +18,6 @@ import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkState;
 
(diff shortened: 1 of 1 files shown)


cpovirk:
Evidently we have a competing internal class that we've been intending to reconcile this with. Possibly we should just admit that we'll never get around to it :(

kashike:
Changed your mind?

swankjesse:
Stopped holding my breath!

mikehearn:
This is kind of annoying. It's a useful utility class but I can't use it without hitting various warnings, IDE highlightings etc.

If you'll never get around to it, why not apply this PR? The class is obviously stable and hasn't changed for many years. It confuses and dilutes the meaning of `@Beta` for it to be stuck in limbo like this.

AParise1217:
@cpovirk  Has this competing internal class been reconciled with `HostAndPort` yet? It looks like this class has yet be change since this initial PR was opened.

PR #2789, opened 2017-04-13
Add Throwables.propagateIf

rotilho:
In a normal application is common to throw an exception after eval an expression like

```
if (balance < amount) {
   throw new InsufficientFundsException();
}
```

```
if (!optional.isPresent()) {
  throw new NotFoundException();
}
```

This PR aims to get rid of those boilerplate code with propagateIf

```
Throwables.propagateIf(balance < amount, () -> new InsufficientFundsException());
Throwables.propagateIf(!optional.isPresent(), () -> new NotFoundException());
```

### DIFF
--- a/guava-tests/test/com/google/common/base/ThrowablesTest.java
+++ b/guava-tests/test/com/google/common/base/ThrowablesTest.java
@@ -323,6 +323,29 @@ public class ThrowablesTest extends TestCase {
         SomeUncheckedException.class);
   }
 
+  @GwtIncompatible // propagateIfPossible(Expression, Supplier)
+  public void testPropageIf_null() throws SomeCheckedException {
+    try {
+      Throwables.propagateIf(false, null);
+      fail();
+    } catch (NullPointerException e){
+    }
+  }
+
+  @GwtIncompatible // propagateIfPossible(Expression, Supplier)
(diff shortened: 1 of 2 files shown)


rotilho:
I signed it!

rotilho:
I signed it!

lowasser:
I'm not sure I follow what aspect of this new method is superior to the original.  The "boilerplate code" you're eliminating seems to be getting replaced with roughly the same number of characters in the method call.  The first version is utterly normal, predictable Java that you don't have to know about a special library method to interpret.

rotilho:
I understand your point however Guava provides Preconditions which have similar idea. Look at the  throwIfInstanceOf:

```
Throwables.throwIfInstanceOf(e, MyException.class)
```
The same could be achieved with
```
if (e instanceof MyException) {
  throw (MyException) e;
}
```

Both have similar amount of caracteres. The same applies for Preconditions so I assumed that amount of caracteres is not the main factor here and reduce the boilerplate code and unneeded complexity it was the main goal.

zackehh:
Plus, you can eliminate characters:

```
Throwables.propagateIf(balance < amount, InsufficientFundsException::new);
```

PR #2791, opened 2017-04-15
Update InternetDomainName docs to reflect the most recent public suffix list.

dave-r12:
It turns out `google` is now a public suffix!

### DIFF
--- a/guava/src/com/google/common/net/InternetDomainName.java
+++ b/guava/src/com/google/common/net/InternetDomainName.java
@@ -294,7 +294,7 @@ public final class InternetDomainName {
    * Foundation's <a href="http://publicsuffix.org/">Public Suffix List</a> (PSL). A public suffix
    * is one under which Internet users can directly register names, such as {@code com},
    * {@code co.uk} or {@code pvt.k12.wy.us}. Examples of domain names that are <i>not</i> public
-   * suffixes include {@code google}, {@code google.com} and {@code foo.co.uk}.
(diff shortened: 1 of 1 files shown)


dave-r12:
I signed it!

guava/src/com/google/common/net/InternetDomainName.java
```diff
@@ -306,7 +306,7 @@ public boolean isPublicSuffix() {
   /**
    * Indicates whether this domain name ends in a {@linkplain #isPublicSuffix() public suffix},
    * including if it is a public suffix itself. For example, returns {@code true} for
-   * {@code www.google.com}, {@code foo.co.uk} and {@code com}, but not for {@code google} or
+   * {@code www.google.com}, {@code foo.co.uk} and {@code com}, but not for {@code thegoogle} or
    * {@code google.foo}. This is the recommended method for determining whether a domain is
```
cpovirk: Apparently `foo` is a public suffix nowadays, too. Want to fix that one, too?

dave-r12: Oh, right - good catch. Any strong preference on naming these things? Maybe `foobar`?

cpovirk: `foobar` is fine with me.

cpovirk:
Sorry, I missed the further update here :( Someone ended up fixing this internally in October.

Issue #2743, opened 2017-02-17
LocalManualCache deadlock in guava 21


epavlov1:
Unit test testCache works with guava 20, locks up in 21 (LocalCache.java,line 2392):

    @Test
    public void testCache()
    {
        String key = "test";
        int count  = 1000;
        Cache<String, String> cache = CacheBuilder.newBuilder().expireAfterAccess(500000, TimeUnit.MILLISECONDS).maximumSize(1000).build();

        IntStream.range(0, count).parallel().forEach((n) -> {
            String val = cache.asMap().computeIfAbsent(key, k->"testv" + n);
            logger.info("Val=" + val);

        });
    }
    @Test
    public void testConcurrentMap()
    {
        String key = "test";
        int count  = 1000;
        Map<String, String> cache = new ConcurrentHashMap<>();

        IntStream.range(0, count).parallel().forEach((n) -> {
            String val = cache.computeIfAbsent(key, k->"testv" + n);
            logger.info("Val:", val );

        });
    }

ben-manes:
Reproduced. This is due to mixing synchronization primitives and non-deterministic ordering for lock acquisition. 

```java
lock {
  // note valueReference can be an existing value or even itself another loading value if
  // the value for the key is already being computed.
  loadingValueReference = new LoadingValueReference<K, V>(valueReference);

  if (e == null) {
    // create entry...
  }
  e.setValueReference(loadingValueReference);
}
...
synchronized (e) {
  V newValue = loadingValueReference.compute(key, function);
  ...
}
```

This creates a computation chain where each computer waits for the preceding one to complete. The order of acquiring the intrinsic lock on `e` is non-deterministic, so T3 could acquire it before T2. Then T3 blocks on `waitForValue()`, which is waiting for T2 to notify its completion. However, that does not release the lock on `e` which T3 owns. This results in a deadlock between T2 and T3.

The assumption was probably that, like `wait()` and `notify()`, the intrinsic lock would be released when waiting on the `SettableFuture`. But that's not the case, as it uses `LockSupport.park`. The mixing of different locking infrastructure tends to not play well. Ideally the lock on `e` would be acquired inside the segment lock and released outside of it, allowing for ensuring the order. But that's not possible with Java's language constraints.

(Verified Caffeine passes this test case)

ben-manes:
[...]

PR #2799, opened 2017-04-25
Fix deadlock if compute function throw a RuntimeException

imario42:
If the computeIfAbsent methods fails with a runtime exception, successive calls to computeIfAbsent lock due to a call to oldValue.waitForValue which never will receive a value then.

A test case to reproduce this behavior - the second computeIfAbsent will never return and your app is frozen.
The pull request restores the previous state of the cache on RuntimeException and the test will finish correctly.

`public class TestGuava extends TestCase
{
    private Cache<String, Boolean> cache = CacheBuilder.newBuilder()
            .maximumSize(10000)
            .expireAfterAccess(30, TimeUnit.SECONDS)
            .build();
    private Map<String, Boolean> cacheMap = cache.asMap();

    private final static Boolean TEST_BOOLEAN = new Boolean(true);
    public void testComputeException()
    {
        try
        {
            cacheMap.computeIfAbsent("a", key ->
            {
                throw new IllegalArgumentException();
            });

            fail();
        }
        catch (Exception e)
        {
            // expected
        }

        Boolean ret = cacheMap.computeIfAbsent("a", key -> {
            return TEST_BOOLEAN;
        });

        assertEquals(Boolean.TRUE, ret);
        assertSame(TEST_BOOLEAN, ret);

        ret = cacheMap.computeIfAbsent("a", key -> {
            return new Boolean(true); // should not create a new entry
        });

        assertEquals(Boolean.TRUE, ret);
        assertSame(TEST_BOOLEAN, ret);
    }
}`


imario42:
I signed it!

ben-manes:
This should rethrow Throwable in case of an Error.

imario42:
@ben-manes : You mean instead of catching "just" the RuntimeException (which actually is rethrown) to catch a Throwable, right? Else, I don't know which "Error" you mean.

ben-manes:
Yes, use Throwable. Otherwise a [java.lang.Error](https://docs.oracle.com/javase/8/docs/api/java/lang/Error.html) or sneakily thrown checked exception will not be handled.

imario42:
Well, thanks to Java 8 that one was easy to fix. Done!

imario42:
I'd like to ask if there is something else I can do that this fix gets merged?

schlosna:
I would also like to see this fix merged and released soon as I've seen a couple separate projects hit deadlocks on cache `compute` and the fix for https://github.com/google/guava/issues/2743 did not fully handle the above mentioned failure cases.

Thanks!

dawid-stosur:
We also experienced this issue on PROD so also would like to see it merged ASAP as we use caches extensively throughout the whole project.

imario42:
Maybe have a look at https://github.com/ben-manes/caffeine which does not suffer from this bug?

ben-manes:
I think this was fixed in https://github.com/google/guava/commit/821082833783f76a1b384924f904ecfd0281ae63 (25.1)

netdpb:
Closing this PR as it looks like the bug was fixed.

PR #2802, opened 2017-04-27
DO NOT SUBMIT - testing travis

ronshapiro:


cpovirk:
I suspect that we want:

    ( cd android && ... )

Instead of:

    cd android && ...

So that we restore the original directory for the later commands. Currently, I think that we're changing directory for `install` and then ending up in the wrong place for `verify`.

ronshapiro:
Yup, that's exactly what we want. d'oh

Issue #1259, opened 2014-10-31
Document that Ints.tryParse(String st) throws NullPointerException


gissuebot:
_[Original issue](https://code.google.com/p/guava-libraries/issues/detail?id=1259) created by **yurachud** on 2013-01-16 at 04:20 PM_

---

If we pass null to Ints.tryParse it throws NullPointerException.
Method uses AndroidInteger.tryParse(), but the first line there is checkNotNull() method.

TarasMazepa:
And do the same for Longs

cgdecker:
I'm adding `@throws NullPointerException if {@code string} is {@code null}` to all the `tryParse` methods even though it's always the case for non-`@Nullable` parameters in Guava, to hopefully reduce the chance of confusion. It's still true that the methods do not throw if _parsing_ fails, but `null` isn't even something you can try to parse.

The change should be synced out soon.

PR #2806, opened 2017-05-02
Issue #1259: Fix misleading documentation on behaviour of Ints.tryParse with null input

deepaks4077:
Referencing #1259.

The documentation for Ints.trParse explicitly says that a NullPointerException is not thrown if the input is null. However, this is not true and the function does indeed throw a NPE.

I have simply removed the misleading lines from the doc.


deepaks4077:
@googlebot I signed it!

cgdecker:
Thanks for the PR. I'm inclined to leave those lines because they do explain the difference between the methods and the equivalent non-"try" method on Integer itself. It's true that the method doesn't throw if parsing fails, but parsing failing and the input being null are different things. I do see how it could be confusing, so I'm inclined to add explicit `@throws NullPointerException` to the Javadoc instead (even though that's not a thing we normally do). We've already done that in a similar cases, as pointed out in #1259.

PR #2807, opened 2017-05-03
ImmediateEventBus that uses Dispatcher.IMMEDIATE

DavesMan:
New child of EventBus that uses immediate Dispatcher, which isn't used in any impl as of now.


DavesMan:
I signed it!

PR #2825, opened 2017-05-23
Added some Unit Tests

TheRealHaui:
Added some Unit Tests that cover uncovered code until yet.

### DIFF
--- a/guava-tests/test/com/google/common/eventbus/EventBusTest.java
+++ b/guava-tests/test/com/google/common/eventbus/EventBusTest.java
@@ -16,13 +16,14 @@
 
 package com.google.common.eventbus;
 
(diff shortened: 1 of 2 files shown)


TheRealHaui:
@cpovirk 
Would you mind approving my contribution?

cpovirk:
Merged, thanks. I ended up tweaking some things:

- using https://github.com/google/google-java-format to match https://google.github.io/styleguide/javaguide.html
- removing `@author` (we're trying to phase that out but haven't done so yet)
- testing parsing zero with various radixes
- changed the empty-string test to test with a valid radix. In your version, it's hard to see that it's failing because of the empty string and not because of the invalid radix

PR #2848, opened 2017-06-16
Support relaxed children() signature in TreeTraverser

findepi:
Previously, the TreeTraverser required that `children()` extended
`Iterable<T>`. This commit adds support for trees where children are
returned as `Iterable<? extends T>`.

Use case: support for https://github.com/prestodb/presto/blob/master/presto-parser/src/main/java/com/facebook/presto/sql/tree/Node.java#L43 without need to copy children lists during traversal.

### DIFF
--- a/guava-tests/test/com/google/common/collect/TreeTraverserTest.java
+++ b/guava-tests/test/com/google/common/collect/TreeTraverserTest.java
@@ -159,6 +159,18 @@ public class TreeTraverserTest extends TestCase {
     assertThat(iterationOrder(ADAPTER_USING_USING.preOrderTraversal(h))).isEqualTo("hdabcegf");
   }
 
+  public void testUsingWithRelaxedChildFunction() {
+    FluentIterable<Tree> travesal = TreeTraverser.using(
+        new Function<Tree, List<? extends Tree>>() {
+          @Override
+          public List<? extends Tree> apply(Tree node) {
+            return node.children;
+          }
+        })
+        .preOrderTraversal(h);
+    assertThat(iterationOrder(travesal)).isEqualTo("hdabcegf");
+  }
+
   @GwtIncompatible // NullPointerTester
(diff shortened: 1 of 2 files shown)


findepi:
e.g. `I signed it!` ?

lowasser:
Why copy when you can use Iterables.unmodifiableIterable?  That'll let you safely treat an `Iterable<? extends T>` as an `Iterable<T>`.

jrtom:
We're still working out the details, but we are working on integrating the capabilities of TreeTraverser into the new-ish package common.graph.  Specifically, I expect that we will _probably_ be retiring TreeTraverser in favor of capabilities based on the new [SuccessorsFunction](https://google.github.io/guava/releases/snapshot/api/docs/com/google/common/graph/SuccessorsFunction.html) and PredecessorsFunction types.

And to come back to your original concern: those types _do_ return `Iterable<? extends N>`.

findepi:
@lowasser , good point, thanks, but it just doesn't look that good.

@jrtom , this is even better since signature `<T> preOrder(T x, Function<T, ? extends Iterable<? extends T>>)` will compile nicely with method references. `TreeTraversal.using(func)` can't be made to work with method references in the "relaxed" case unless explicit type for method generic is provided.

That said, is there going to be  `preOrder()` operating on `SuccessorsFunction`-s? I looked into sources, but couldn't find anything besides the interface declaration.

Last but not least, does the _probable_ retirement of `TreeTraverser` say anything in regards on this PR? :)

jrtom:
Yes, there are going to be various traversal options including preorder.  Wouldn't be much of a replacement otherwise.  :) The details are still pending review but there should be something soon, probably in a few weeks.

The fact that TreeTraverser is expected to be replaced suggests to me that we're unlikely to change it at this point, but that's not my call.

Hope this helps.

PR #2856, opened 2017-06-26
Add EquivalenceRelation and corresponding tests

### LINKED ISSUE #2851: Add Equivalence Relation
When we have some reflexive symmetric transitive relation between two instances of one class then an equivalence relation is a helpful assistant for check that this relation was accounted in code earlier.
For example we processed a relation between two people (p1, p2) and we don't want to process a relation (p2,p1) because it is the same. We put (p1, p2) to a set and when we invoke 'contains' method for (p2, p1) on the set we will get true.

[More about Equivalence Relation](https://en.wikipedia.org/wiki/Equivalence_relation)
[Example of implementation](https://github.com/ivan-osipov/guava/commit/007e471f459085edea9a04a13e08c40af4659911)

ivan-osipov:
It resolves #2851

### DIFF
--- a/guava-tests/test/com/google/common/base/EquivalenceRelationTest.java
+++ b/guava-tests/test/com/google/common/base/EquivalenceRelationTest.java
@@ -1,0 +1,82 @@
+package com.google.common.base;
+
+import static com.google.common.truth.Truth.assertThat;
+
+import junit.framework.TestCase;
+
+/**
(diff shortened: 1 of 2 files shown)


ivan-osipov:
I signed it!

raghsriniv:
https://github.com/google/guava/issues/2851 was closed.

Issue #2797, opened 2017-04-21
Remove @Beta from com.google.common.util.concurrent.RateLimiter


basilv:
I'd like RateLimiter to come out of Beta. It has been around since version 13.0

cpovirk:
We'll take a look. There are a number of things about it that we've hoped to eventually change, but it's hard to see when we would ever get around to them, so perhaps we should just finalize it.

maiph:
There is a TODO over at com.google.common.util.concurrent.RateLimiter.create

```java
/*
 * TODO(cpovirk): make SleepingStopwatch the last parameter throughout the class so that the
 * overloads follow the usual convention: Foo(int), Foo(int, SleepingStopwatch)
 */
@VisibleForTesting
static RateLimiter create(SleepingStopwatch stopwatch, double permitsPerSecond) {
  RateLimiter rateLimiter = new SmoothBursty(stopwatch, 1.0 /* maxBurstSeconds */);
  rateLimiter.setRate(permitsPerSecond);
  return rateLimiter;
}
```

Would you want a pull request to refactor this?

cpovirk:
Sure, thanks.

stevedes77:
I would also like this API to come out of Beta. Seeing that there hasn't been much activity in this API since this issue was filed, what is the path to getting the annotation removed?

djagielo:
I'd also be glad to see that out of Beta. Are there any still blocking TODOs? What are the plans?

vpriesner:
Any updates?
There seems to be one TODO left:

```
 * @author Dimitris Andreou
 * @since 13.0
 */
// TODO(user): switch to nano precision. A natural unit of cost is "bytes", and a micro precision
// would mean a maximum rate of "1MB/s", which might be small in some cases.
```

joaquinfernandez131:
Any updates on this?

eamonnmcmanus:
Duplicate of #5612. I'm closing in favour of that one since it has an assignee and this one doesn't.

PR #2869, opened 2017-07-10
Refactored RateLimiter#create signature

maiph:
Referencing: #2797

### DIFF
--- a/guava/src/com/google/common/util/concurrent/RateLimiter.java
+++ b/guava/src/com/google/common/util/concurrent/RateLimiter.java
@@ -156,16 +156,16 @@ public abstract class RateLimiter {
   public static RateLimiter create(double permitsPerSecond, long warmupPeriod, TimeUnit unit) {
     checkArgument(warmupPeriod >= 0, "warmupPeriod must not be negative: %s", warmupPeriod);
     return create(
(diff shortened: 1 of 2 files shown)


guava/src/com/google/common/util/concurrent/RateLimiter.java
```diff
@@ -119,15 +119,11 @@ public static RateLimiter create(double permitsPerSecond) {
...
-    return create(SleepingStopwatch.createFromSystemTimer(), permitsPerSecond);
+    return create(permitsPerSecond, SleepingStopwatch.createFromSystemTimer());
   }
 
-  /*
-   * TODO(cpovirk): make SleepingStopwatch the last parameter throughout the class so that the
-   * overloads follow the usual convention: Foo(int), Foo(int, SleepingStopwatch)
-   */
   @VisibleForTesting
-  static RateLimiter create(SleepingStopwatch stopwatch, double permitsPerSecond) {
```
cpovirk: Can you also fix the other `create(..., SleepingStopwatch)` method below?

maiph: Sure, I missed that one 👍

PR #2875, opened 2017-07-20
(refactor) some opportunities to use diamond operator

desmorto:


jbduncan:
Hmm, I wonder if Android even supports the diamond operator...

JakeWharton:
Diamond is a compilation​ feature only. Has nothing to do with runtime.

desmorto:
well, travis had a android test as well...

jbduncan:
Yep, looks like it passes Travis (which indeed runs the Android tests), so I've no objections to this!

cpovirk:
I think I found a couple cases that didn't compile with `-source 7 -target 7` (which we use so that the Android version is usable under Java 7, too), thanks to difference in the compiler's type inference under those flags. But the rest of the changes are in. Thanks.

PR #2889, opened 2017-08-01
Add generics with Maps. fromProperties().

purplest:
Add generics with Maps. fromProperties().

It can use like this:
```java
Map<Integer, Integer> maps = fromProperties(prop, (k)->Integer.valueOf(k), (v)->Integer.valueOf(v));
```
It's very useful when we use it to convert properties to a generics map.

And in other way, we can implement with BiFunction<T, U, R> to expand this method.

### DIFF
--- a/guava/src/com/google/common/collect/Maps.java
+++ b/guava/src/com/google/common/collect/Maps.java
@@ -1382,10 +1382,12 @@ public final class Maps {
     }
   }
 
+  @FunctionalInterface
   public interface KConverter<K>{
     K convert(String key);
   };
 
+  @FunctionalInterface
   public interface VConverter<V>{
     V convert(String value);
   };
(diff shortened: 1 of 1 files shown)


akhil-am:
I signed it!

lowasser:
Can you explain your use case in more detail?  Why you're using `Properties` in the first place, given that it seems pretty ancient and rarely used?  I'm also not sure this adds all that much value over e.g. `properties.entrySet().stream().collect(toImmutableMap(...))`.

ronshapiro:
It also seems a bit odd to me to have full pojos encoded in a properties file.

nglorioso:
Unfortunately, it looks like we're not going to be able to accept this change. It sounds like there are alternatives provided (either working with `properties.entrySet()` or the `ImmutableMap<String, String>` returned currently and interpreting the correct values).

PR #2900, opened 2017-08-09
Handle exception in LocalCache.asMap().compute(...)

mmimica:
LocalCache.asMap().compute(...) deadlocks if its lambda throws an exception. Demonstrated by the test.
Fixed by properly handling the exception and realising the lock.


mmimica:
I signed it!

PR #2901, opened 2017-08-09
LocalCache.asMap().compute(...) deadlocks if its lambda throws an exception

mmimica:
LocalCache.asMap().compute(...) deadlocks if its lambda throws an exception. Demonstrated by the test. Fixed by properly handling the exception and realising the lock.

### DIFF
--- a/guava-tests/test/com/google/common/cache/LocalCacheMapComputeTest.java
+++ b/guava-tests/test/com/google/common/cache/LocalCacheMapComputeTest.java
@@ -82,4 +82,13 @@ public class LocalCacheMapComputeTest extends TestCase {
     });
     assertEquals(0, cache.size());
   }
+
+  public void testComputeExceptionally() {
+    try {
+      doParallelCacheOp(count, n -> {
+        cache.asMap().compute(key, (k, v) -> { throw new RuntimeException(); });
+      });
+    } catch (RuntimeException ex) {
+    }
+  }
 }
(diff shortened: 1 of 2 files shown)


ben-manes:
Also see https://github.com/google/guava/pull/2799.

mmimica:
Oh I see, a duplicate. Well, I think my fix is better because it localizes changes in `LoadingValueReference` where the actual problem is. An exception is kind of function's legitimate result, which should be caught there and set as a result.

Also there is a test. It could be written better. For some reason I cannot get `@Test(expected = RuntimeException.class)` to work.

ben-manes:
Does this remove the loading value? I don't think it does from a glance.

mmimica:
It doesn't. It assumes the rest of the code works as expected. I didn't manage to reproduce any problems. Even the test from the other PR passes.
But I'm not familiar with the inner workings of `LocalCache`.

ben-manes:
From looking at the `Map.compute` implementation, I think it doesn't handle the exception being thrown. Instead the entry is stuck and a subsequent `compute` replaces it because of the lines previous to your change regarding the previous. It is then stuck with a `LoadingValueReference`, which works as expected because a `getIfPresent` will read from the `oldValue`. But likely a `LoadingCache.get` will be stuck thinking there is a computing entry and that the loader will do the cleanup (`getAndRecordStats` vs `waitForLoadingValue`).

Ideally after a test case an internal integrity check would be run to find these conditions. Then any invisible or hard to query cases would be caught. I think `CacheTesting.checkEmpty` would do that here due to `keySet` likely having 1 entry. The `checkValidState` doesn't verify that all entries are fully loaded, so it wouldn't catch it.

Issue #2790, opened 2017-04-13
com.google.common.escape.Platform - ThreadLocal leak


sdavids:
org.apache.catalina.loader.WebappClassLoaderBase.checkThreadLocalMapForLeaks The web application [xxx] created a ThreadLocal with key of type [com.google.common.escape.Platform$1] (value [com.google.common.escape.Platform$1@614d8014]) and a value of type [char[]] (value [[C@1fb37a49]) but failed to remove it when the web application was stopped. Threads are going to be renewed over time to try and avoid a probable memory leak.

Maaartinus:
As I [wrote][1] long time ago, this one is trivially fixable. Using `ThreadLocal#initialValue` together with a JDK class saves a line or two at the cost of potential memory leak. No good deal, IMHO.

[1]: https://github.com/google/guava/pull/2112#issuecomment-129839694

cpovirk:
Thank you for reporting this. I'm sorry for letting it sit. Every time I go to look up the problem with `ThreadLocal` leaks, I find a bunch of incorrect information :( (In particular, I think that some of the Tomcat documentation is [wrong](http://java.jiderhamn.se/2012/01/29/classloader-leaks-iv-threadlocal-dangers-and-why-threadglobal-may-have-been-a-more-appropriate-name/).) Additionally, there seem to be several related problems, so it's wrong for me to look for "the \[one\] problem." My current understanding, which hopefully I have right and will be able to communicate well, is:

Generally speaking, if all the threads that have values go away, then nothing important is leaked. Perhaps there are cases in which that's not true, but I haven't looked into them, and that's not what this issue is about, so let's set that aside.

The main family of problems, then, comes when a `Thread` sticks around in a thread pool but the `ThreadLocal` that has a value for that `Thread` will never be used again. (I guess in theory there could be a problem in which the `ThreadLocal` _was_ going to be used again but we knew that it would never be used from that particular _`Thread`_. But I don't think that's the scenario anyone has in mind, either.)

[...]

Issue #1977, opened 2015-02-19
Possible deadlock in ImmutableList 


kronar:
ImmutableList contains following code since version 15.0

 private static final ImmutableList<Object> EMPTY =
      new RegularImmutableList<Object>(ObjectArrays.EMPTY_ARRAY);

cpovirk:
Can you elaborate on the deadlock? I guess the idea is that one thread starts to load ImmutableList while the other starts to load RegularImmutableList? That's probably not good, though we may have this problem in a lot of places, and I haven't heard reports of trouble... yet...

apangin:
Here is the minimal test case I could create. It hange intermittently.

    public static void main(String[] args) throws Exception {
        new Thread() {
            public void run() {
                ImmutableList.of();
            }
        }.start();

        Thread.sleep(1);

        Class.forName("com.google.common.collect.RegularImmutableList");
    }

cpovirk:
Sorry, I missed the last part of your message (about how you've hit that in prod). That makes more much more worried about this.

Part of the reason I'm worried is that we might have these cycles in lots of places. I'll fix this one, but we may have others. Maybe we should write a test that loads each of our classes in a new, instrumented `ClassLoader`. I'm always in over my head with `ClassLoader`, but I'm hoping that maybe we can override `loadClass` to record all classes loaded during the loading of the class. Then we can build a graph and look for cycles. If they're all over, we may just wait for reports to come in.... Hopefully there would be a manageable number.

cpovirk:
`ClassLoader` is probably the wrong approach. @lukesandberg suggests that ASM may help. For now, though, I'm dropping this.

(The specific problem of `RegularImmutableList` has been fixed internally, so it will be mirrored out soon.)

cpovirk:
Thought about this a little more today.

One subtlety here is how the app starts to initialize `RegularImmutableList` without first initializing `ImmutableList`. How does the app get at package-private `RegularImmutableList` except through `ImmutableList`? Well, it turns out that Guava calls directly into `RegularImmutableList` in places outside `ImmutableList`. For example, `ImmutableSortedMap.fromEntries` does this.

Until then, I had hoped that I could restrict my search to classes that involve `clinit` from at least 2 `public` (or `protected`) classes. (It turns out, as expected, that Guava has a lot of cycles that involve non-`public` classes.) But if I restrict myself to cases in which at least 2 of the classes involved are `public`, I miss the `ImmutableList` problem.

[...]

PR #2902, opened 2017-08-10
Static initializer references subclass

Moriadry:
Such references can cause JVM-level deadlocks in multithreaded environment, when one thread tries to load superclass and another thread tries to load subclass at the same time.

My refactor will not change the method provided to outer space.

### DIFF
--- a/guava/src/com/google/common/collect/ComparisonChain.java
+++ b/guava/src/com/google/common/collect/ComparisonChain.java
@@ -124,9 +124,9 @@ public abstract class ComparisonChain {
         }
       };
 
-  private static final ComparisonChain LESS = new InactiveComparisonChain(-1);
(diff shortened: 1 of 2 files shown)


cpovirk:
Is it possible for a thread to initialize, e.g., `InactiveComparisonChain` without first initializing `ComparisonChain`? If not, this may be benign. (See [a previous discussion of class-initialization deadlocks](https://github.com/google/guava/issues/1977#issuecomment-146659846).) And while eliminating benign cycles is still a good thing, I'm nervous about [using](https://github.com/google/guava/pull/2112#issuecomment-128842365) [`ThreadLocal`](https://github.com/google/guava/issues/2790) because it can produce leaks. There's probably a safe way to use `ThreadLocal`, but I'd prefer not to go down that road if we don't have to.

ben-manes:
Assuming there is a cycle, wouldn't the initialization-on-demand holder idiom work? That would be preferable to ThreadLocal.

PR #2903, opened 2017-08-10
Polishing

jbduncan:
This PR applies a good number of apparently applicable IntelliJ IDEA code inspections.


cpovirk:
GitHub ate my comment :( In short, I took most of this:
- not sure if we can do try-with-resources on old Android. ditto Integer.compare, etc.
- we aren't in the habit of omitting private from enum constructors (except on private enum classes)
- new instances in FreshValueGenerator is probably intentional
- double append() is harder to read in some cases, so string concat is nicer when performance doesn't matter
- MapsTest is specifically testing the newHashMap method
- not sure the Optional.map version is clearer than the original, but maybe I just need to get more used to functional programming in Java
- doesn't seem worth worrying about the semicolon after enum
- space in LittleEndianByteArray is from a google-internal comment
- AtomicDouble we're trying to leave as close to upstream as possible
- maybe other things I forgot -- feel free to ask.

jbduncan:
Hi @cpovirk, sorry for being quiet on this.

I think a few things were missed, but I'm not sure which, so I'm rather satisfied with the commit you've pushed. Thanks!

PR #2908, opened 2017-08-13
Fix javadoc link.

hawjo06:
The link was generated as plaintext. The change will result in the link
being rendered as a hyperlink.


cpovirk:
javac9 -Xdoclint rejects this version :(

It sounds like what we have is supposed to work (or at least `ImmutableTable.Builder#Builder`, perhaps?), but [javadoc8 doesn't understand it](https://bugs.openjdk.java.net/browse/JDK-8031625). Oddly, it looks like javadoc9 doesn't understand it, either.... Maybe I'm doing something wrong.

cpovirk:
Hmm, no, I was doing something wrong, though I'm not 100% sure what it was :) I'll see if I can at least get it to work with javadoc9 now.

hawjo06:
@cpovirk Thanks for looking into this, and the detailed responses.   I will close this PR.

Issue #2911, opened 2017-08-15
Add flatteningToMultimap for a stream of keys


phoenix384:
There already is the method Multimaps.flatteningToMultimap for a stream of values.
But one could also want to map an object against multiple keys coming from a stream.
So a collector method flatteningToMultimap(Function<? super T, ? extends Stream<? extends K>>, Function<? super T, ? extends V>, Supplier<M>) would be useful.

phoenix384:
Added it in PR #2912

netdpb:
Could you just invert the multimap you get from the existing method?

```java
Multimap<Key, Value> multimap = stream.collect(
    Multimaps.flatteningToMultimap(keyFunction, valueStreamFunction, multimapSupplier));

Multimap<Value, Key> inverted = Multimaps.invertFrom(multimap);
```

That doesn't seem difficult, so maybe it's not necessary to add another method. A new method would have to have a different name since the erased parameter types are the same.

PR #2912, opened 2017-08-15
Add flatteningToMultimap for a stream of keys

phoenix384:
Issue #2911


phoenix384:
@googlebot I signed it!

netdpb:
See my comment on #2911 about whether this is necessary.

But also, this method's erased parameter types are the same as the existing method, so I don't think compiles.

PR #2938, opened 2017-09-11
Update IteratorBenchmark.java

fvarg00:
Refer to objects by their interfaces(Item 52)

### DIFF
--- a/guava-tests/benchmark/com/google/common/collect/IteratorBenchmark.java
+++ b/guava-tests/benchmark/com/google/common/collect/IteratorBenchmark.java
@@ -21,6 +21,7 @@ import com.google.caliper.Benchmark;
 import com.google.caliper.Param;
 import java.util.ArrayList;
 import java.util.LinkedList;
+import java.util.List;
 
 /**
  * Tests the speed of iteration of different iteration methods for collections.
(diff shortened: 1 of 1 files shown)


fvarg00:
I signed it!

guava-tests/benchmark/com/google/common/collect/IteratorBenchmark.java
```diff
@@ -32,7 +33,7 @@
 
   // use concrete classes to remove any possible polymorphic overhead?
   Object[] array;
-  ArrayList<Object> arrayList;
+  List<Object> arrayList;
   LinkedList<Object> linkedList;
```
jbduncan: How about this `LinkedList` as well?

perceptron8: How about this "use concrete classes to remove any possible polymorphic overhead?" comment? ;)

jbduncan: Ah, I missed that. Silly me. :P

fvarg00: I shall refactor to remote all concrete classes. Thanks @perceptron8

jbduncan: @fvarg00 Uh, I _think_ what the comment is saying is that concrete classes are purposely used _instead_ of the `List` interface to "remove any possible polymorphic overhead" (not entirely sure what that means), so AFAICT this PR would probably not be accepted even if the `LinkedList` is refactored...

fvarg00: @jbduncan - thanks for your comment. It makes sense now.

PR #2945, opened 2017-09-19
Fix #2914 by switching -android version suffix to -ga; revert vanilla…

### LINKED ISSUE #2914: Guava for Android and Guava for the JRE need separate Maven namespaces
Guava for Android is lacking classes that are provided in Guava for the JRE (notably around support for Java 8 in the immutable collections). The two are not interchangeable. 

Maven's dependency resolution looks at group and project IDs, and then attempts to find the most recent version. Because the only difference between the android and JRE versions of guava is the version number, this can result in Java 8 projects getting the Android version of Guava (and, presumably, vice versa).

For reference, at the time of writing [search.maven.org](http://search.maven.org/#search%7Cga%7C1%7Cguava) believes the most recent guava release is the one for Android.

guava-oops

The correct thing to do is give the Android version a different project ID.

yborovikov:
This patch changes `-android` version suffix to `-ga` for Android / Java 7 flavored version of Guava, resolving #2914.

The resolution piggy-backs on documented Maven [`ComparableVersion`](http://maven.apache.org/ref/3.5.0/maven-artifact/apidocs/org/apache/maven/artifact/versioning/ComparableVersion.html) feature that equates `ga`-suffixed versions (e.g. `24.0-ga`) with non-suffixed ones (e.g. `24.0`) in the context of comparing / ordering / determining the latest version.

Note: This also reverts [the patch](https://github.com/google/guava/commit/61d2a6faec8b27d021b9f37f7dfb6655ae060fad) by @cpovirk, restoring the original vanilla (Java 8+) Guava versioning scheme (e.g. from patched `24.0-jre` back to `24.0`).

Note: [`README.md`](https://github.com/google/guava/blob/master/README.md#latest-release) text should be updated before releasing the next version to reflect the suffix change for the Android version.

### DIFF
--- a/android/guava-testlib/pom.xml
+++ b/android/guava-testlib/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>com.google.guava</groupId>
     <artifactId>guava-parent</artifactId>
-    <version>24.0-android-SNAPSHOT</version>
+    <version>24.0-ga-SNAPSHOT</version>
   </parent>
   <artifactId>guava-testlib</artifactId>
   <name>Guava Testing Library</name>
--- a/android/guava-tests/pom.xml
+++ b/android/guava-tests/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>com.google.guava</groupId>
     <artifactId>guava-parent</artifactId>
-    <version>24.0-android-SNAPSHOT</version>
+    <version>24.0-ga-SNAPSHOT</version>
   </parent>
   <artifactId>guava-tests</artifactId>
   <name>Guava Unit Tests</name>
--- a/android/guava/pom.xml
+++ b/android/guava/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>com.google.guava</groupId>
     <artifactId>guava-parent</artifactId>
-    <version>24.0-android-SNAPSHOT</version>
+    <version>24.0-ga-SNAPSHOT</version>
   </parent>
   <artifactId>guava</artifactId>
   <packaging>bundle</packaging>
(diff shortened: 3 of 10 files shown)


ooxi:
I have read the issue linked. Nevertheless I think using `-ga` instead of `-android` will lead to a lot of confusion. Without having seen this pull request I would not have known the difference between `24.0` and `24.0-ga` and would most likely have picked `24.0-ga` when updating (since general availability seems better than no suffix).

yborovikov:
@ooxi You'd have every right to get confused initially (even though such mistake would usually get immediately realized at your project compile time) - what Guava is trying to achieve with releasing both android/java7 and java8+ artifacts is indeed confusing. Arguably, they do have a reason for doing this, though.

As to `GA` version being considered "better" than plain (non-suffixed) release one - I'm not sure it's clear cut (and might be highly subjective, depending on one's history of exposure to these). When I encounter such versions I try to check the project documentation on their versioning scheme and semantics.

Wikipedia's Software release life cycle article has this diagram:
Software release life cycle
which probably (yet not necessarily) reflects some majority opinion / consensus on the topic.

Anyway, in case of a doubt a quick glance at [Guava project landing page](https://github.com/google/guava/#latest-release) should clear any confusion and make the choice between vanilla and android/java7-specific version reasonably obvious.

ooxi:
Thank you for such an in-depth explanation of your reasoning. While I still would have been confused and would likely have chosen the wrong version at first, you are right in your estimation that it would have been more logical to use `24.0` than `24.0-ga`. I therefore withdraw my initial rejection of this change.

cgdecker:
I'm not clear on how `-ga` makes sense. "General availability" isn't a term we use in Guava releases and really doesn't have any particular relevance to "Android flavor". The scheme we've chosen is `<version>-jre` and `<version>-android` for the two flavors of Guava, and that's what we'll be going with barring significant evidence that we should be doing something different.

Issue #2887, opened 2017-07-28
Throwables#lazyStackTrace throws UnsupportedOperationException on AIX 5.3


kyle-cackett:
The mechanism which detects whether to fallback on the non-lazy stack trace implementation seems to be broken on AIX 5.3 with IBMs JDK `jdk-6.0.0.645-aix-powerpc`. When iterating through the list returned by the method on such a system I get the following exception + trace

> Caused by: java.lang.UnsupportedOperationException: You cannot access stack trace depth this way in this implementation
        at sun.misc.SharedSecrets$1.getStackTraceDepth(SharedSecrets.java:172)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:60)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:37)
        at java.lang.reflect.Method.invoke(Method.java:611)
        at com.google.common.base.Throwables.invokeAccessibleNonThrowingMethod(Throwables.java:292)
        at com.google.common.base.Throwables.access$200(Throwables.java:48)
        at com.google.common.base.Throwables$1.size(Throwables.java:284)
        at java.util.AbstractList$SimpleListIterator.hasNext(AbstractList.java:51)
        at com.google.common.collect.Iterators$PeekingImpl.hasNext(Iterators.java:1171)
        at com.delphix.logger.Logger.getLogger(Logger.java:96)
        at com.delphix.session.impl.common.SessionManager.<clinit>(SessionManager.java:23)
        at java.lang.J9VMInternals.initializeImpl(Native Method)
        at java.lang.J9VMInternals.initialize(J9VMInternals.java:199)
        ... 21 more

Explicitly using `new Throwable().getStackTrace` resolved the problem.

cpovirk:
Thanks, and sorry for the slow response. It's unfortunate that IBM would define this method only to make it not work, but I assume they have their reasons. If you're interested in adding `catch (UnsupportedOperationException e)` and an appropriate fallback, we could take a pull request.

fdesu:
@cpovirk 
I'm interested in this one, so if no one minds, I want to make an implementation for a fallback, however, should we care only for `UnsupportedOperationException` or it is worth to catch something generic e.g. `RuntimeException`?  (hence I'm not a fan of such an idea)

cpovirk:
[...]

PR #2947, opened 2017-09-22
Throwables#lazyStackTrace unsupported on IBM JDK 6

fdesu:
Fix addresses a problem occurred on IBM JDK 6. To ensure 
that we can call underlying method, Throwables#getSizeMethod, 
to get  stack size, first a smoke test  with dummy exception. 
Once method executed with success, it is safe to return it for 
further usage.

Issue: https://github.com/google/guava/issues/2887

### DIFF
--- a/guava/src/com/google/common/base/Throwables.java
+++ b/guava/src/com/google/common/base/Throwables.java
@@ -505,12 +505,26 @@ public final class Throwables {
 
   /**
    * Returns the Method that can be used to return the size of a stack, or null if that method
-   * cannot be found (it is only to be found in fairly recent JDKs).
+   * cannot be found (it is only to be found in fairly recent JDKs). Tries to dry run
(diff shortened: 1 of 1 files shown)


fdesu:
For the sake of clarification: there were two obvious ways:

- Guard `AbstractList#size` in `Throwables#jlaStackTrace` with `try-catch`
- Make a dry call of `JavaLangAccess` and if it success, then assume, that it is safe to call it

We can go with the first option, however, I think, that second seems more elegant.

cpovirk:
The second sounds fine to me. I just submitted your change for internal review.

cpovirk:
(Another reason to prefer your second option is that it makes sure `lazyStackTraceIsLazy` returns `false` in the case in which we aren't going to provide lazy access.)

fdesu:
@cpovirk 
Exactly. I think it just ensures that API works fine, instead of make only one method out of two work as designed.
However, I don't have any idea, how to simulate/test it with a unit test (most probably there's no way even to do that, since it more likely to be a system/end-to-end one)

PR #2961, opened 2017-10-09
Add JPMS module name

### LINKED ISSUE #2920: Automatic-Module-Name missing from sub-module manifests
Version 23 was supposed to add `Automatic-Module-Name` to the manifest file but it only did so for `guava-parent`: https://github.com/google/guava/blob/v23.0/pom.xml#L119

Please add `Automatic-Module-Name` for all sub-modules otherwise JDK9 clients will end up with the wrong module name.

jodastephen:
Move to the correct pom.xml.

Fixes #2920
Commit 9feba7c949ef9fd2092c05c29dcbe0ddba7a7cf5 of #2846 placed the `Automatic-Module-Name` declaration in the wrong`pom.xml`. As such, Guava still isn't a valid module, which blocks a whole heap of projects from declaring `module-info.java`.

Feel free to make a separate change rather than merging this PR to satisfy any IP rules.

### DIFF
--- a/guava/pom.xml
+++ b/guava/pom.xml
@@ -39,6 +39,16 @@
   <build>
     <plugins>
       <plugin>
+        <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <archive>
+            <manifestEntries>
+              <Automatic-Module-Name>com.google.common</Automatic-Module-Name>
+            </manifestEntries>
+          </archive>
+        </configuration>
+      </plugin>
+      <plugin>
         <extensions>true</extensions>
         <groupId>org.apache.felix</groupId>
         <artifactId>maven-bundle-plugin</artifactId>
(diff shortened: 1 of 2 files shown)


jbduncan:
@jodastephen By my understanding, if you just sign the Google CLA, then that should be enough to allow this PR to be considered for merging.

jodastephen:
I signed it!

cpovirk:
Thanks. Any opinion on whether we should set an `Automatic-Module-Name` for our `guava-testlib` test utilities? They're much, much less heavily used, but maybe we might as well? Or maybe it's not worth worrying about modules for code that runs only in tests, since it sounds like there might not be a good story yet for running most tests with modules on?

jodastephen:
As it stands, I don't think its clear how test libraries will operate with modules - I suspect that modules don't really apply there. If runtime code might depend on testlib then of course it should have a module name.

cpovirk:
Thanks.

PR #2983, opened 2017-10-31
Remove @GwtCompatible annotations from classes in the hash package,

niloc132:
which have no corresponding .gwt.xml file.

Alternatively, if these are expected to be gwt-compatible, I can add the .gwt.xml and the missing emulation for LongAdder (which references other presently gwt-incompatible classes).


cpovirk:
Thanks.

PR #2990, opened 2017-11-11
Fix Types.toString for arrays

MaridProject:
The current behaviour of `TypeToken.of(<array type>).toString()` mismatches with the `Types.toString()` javadoc: 

> For array type {@code Foo[]}, {@code "com.mypackage.Foo[]"} are returned

For example, `TypeToken.of(String[].class)` returns `[Ljava.lang.String;`.

Fortunately, a new java8 method `getTypeName()` has appeared whose behaviour is exactly the same as `Types.toString()` except that it can deal with arrays (including primitive ones).

### DIFF
--- a/guava-tests/test/com/google/common/reflect/TypesTest.java
+++ b/guava-tests/test/com/google/common/reflect/TypesTest.java
@@ -404,9 +404,10 @@ public class TypesTest extends TestCase {
   }
 
   public void testToString() {
-    assertEquals(int[].class.getName(), Types.toString(int[].class));
-    assertEquals(int[][].class.getName(), Types.toString(int[][].class));
-    assertEquals(String[].class.getName(), Types.toString(String[].class));
+    assertEquals(String.class.getName(), Types.toString(String.class));
+    assertEquals(int.class.getName() + "[]", Types.toString(int[].class));
+    assertEquals(int.class.getName() + "[][]", Types.toString(int[][].class));
+    assertEquals(String.class.getName() + "[]", Types.toString(String[].class));
     Type elementType = List.class.getTypeParameters()[0];
     assertEquals(elementType.toString(), Types.toString(elementType));
   }
--- a/guava/src/com/google/common/reflect/TypeToken.java
+++ b/guava/src/com/google/common/reflect/TypeToken.java
@@ -821,7 +821,7 @@ public abstract class TypeToken<T> extends TypeCapture<T> implements Serializabl
 
   @Override
   public String toString() {
-    return Types.toString(runtimeType);
+    return runtimeType.getTypeName();
   }
 
   /** Implemented to support serialization of subclasses. */
--- a/guava/src/com/google/common/reflect/Types.java
+++ b/guava/src/com/google/common/reflect/Types.java
@@ -174,7 +174,7 @@ final class Types {
    * </ul>
    */
   static String toString(Type type) {
-    return (type instanceof Class) ? ((Class<?>) type).getName() : type.toString();
+    return type.getTypeName();
   }
 
   @Nullable
@@ -243,7 +243,7 @@ final class Types {
 
     @Override
     public String toString() {
-      return Types.toString(componentType) + "[]";
+      return componentType.getTypeName() + "[]";
     }
 
     @Override
@@ -646,7 +646,7 @@ final class Types {
     abstract Type usedInGenericType(Type type);
 
     String typeName(Type type) {
-      return Types.toString(type);
+      return type.getTypeName();
     }
 
     boolean jdkTypeDuplicatesOwnerName() {


MaridProject:
I signed it!

MaridProject:
I signed it!

MaridProject:
I signed it!

MaridProject:
I signed it!

MaridProject:
I signed again!

MaridProject:
Please, help me with obtaining cla:yes label. What's wrong with my account? The e-mails in the git commit and my GitHub account and the CLA account are the same.

jbduncan:
@MaridProject Oh, really? When I look at the commit list directly (https://github.com/google/guava/pull/2990/commits), it shows the following...

capture

Usually, in the place of "Dmitry Ovchinnikov", there would be your GitHub username, like in this example:

capture2

...but because there isn't, it seems to suggest to me personally that the email address you used to create the commit isn't exactly the same as your GitHub email address.

Have you tried entering `git config user.email` in Git Bash (or the terminal that you use with Git)? If it's set up correctly, it should show this:

```
$ git config user.email
d.ovchinnikow@gmail.com
```

If the email part is different to what it shows above, you can try these [GitHub instructions](https://help.github.com/articles/setting-your-commit-email-address-in-git/) to set up your email address associated with your local Git. Then if you type `git commit --amend`, then by my superficial understanding, it should update your commit locally with the up-to-date email address, which you can then push again with `git push --force-with-lease`.

Hope this helps!

MaridProject:
@jbduncan ,

thank you very much. There was a typo in my e-mail address. Here is my configuration:

```
[dm@dmpc ~]$ git config --list
user.name=Dmitry Ovchinnikov
user.email=d.ovchinikow@gmail.com
core.editor=mcedit
```
And here is a snippet from the git log

```
commit 517986723832b12c0bdd17ad707e64ac26a9ad4e (HEAD -> master, origin/master, origin/HEAD)
Author: Dmitry Ovchinnikov <d.ovchinikow@gmail.com>
Date:   Sat Nov 11 19:15:01 2017 +0300

    Fix Types.toString for arrays
```
So, I skipped one letter "n" and so the emails don't match.
Thank you again.

cpovirk:
Thanks for noticing this. I believe the bug is in the docs, as I think our goal is to match the JDK's behavior. Primary `TypeToken` author @fluentfuture would know.

MaridProject:
Curiously that the method Types.toString(Type) (with package-private modifier) has javadocs but TypeToken.toString() (with public modifier) doesn't. Nonetheless, I feel that the first sentence `Returns human readable string representation of {@code type}.` is what you wanted since the beginning (by intent). `[Ljava.lang.String;` could not be considered as a "human readable string", it's just a string representation.

cpovirk:
I do believe that `String[]` is nicer in a vacuum, but I think that we're trying to exactly match the JDK's strings. We go to some effort to do that, even across JDK versions (e.g., [here](https://github.com/google/guava/blob/4f794799aa981eba0000b54004257c0f4fa78b88/guava/src/com/google/common/reflect/Types.java#L299)).

cpovirk:
I do think I was too quick to dismiss the behavior change: I had been thinking that the problem was our implementations of **`Type`**`.toString()`, where we'd like to match the JDK, but of course you're talking about **`TypeToken`**`.toString()`, where we're freer to do what we want. (That said, it's possible that someone depends on the current behavior. We'd still be within our rights to change it, but we'd have to weigh the risks.) Sorry about that.

And now I've been starting at this for 30 minutes, and I'm pretty confused. It looks like we _mostly_ use `JavaVersion.CURRENT.typeName(Type)`, but we sometimes use `Types.toString(Type)` instead. And those two look to be intended to mostly match. They're probably intended to be different in some cases, but I suspect we're not using them all in exactly the right places -- one consequence of which is that we might be using `[Ljava.lang.String;` and `java.lang.String[]` in the wrong places. As noted above, it's one thing for `TypeToken` to have its own `toString` format, but we'd like our `Type` implementations to match the JDK's. But I'm not sure whether we're accomplishing that, either. (Certainly our tests for `toString()` are sparse.)

There's probably something better that could be done here, distinguishing between methods used in `toString()` and methods used in `getTypeName()` and then changing `TypeToken` to rely solely on `getTypeName()`. (But it's more complicated than that because we'd want identical behavior in the backport, where we can't rely on `getTypeName()`.) I assume that it's possible to do this, but it's likely to be a larger change, requiring testing on various JDK versions to account for the quirks of each.

fluentfuture:
Yeah. We went through quite some work to keep our Type#toString() consistent with JDK's native Type implementations.

In a nutshell, Guava's toString() has been always delegating to the underlying Type's toString() or getName() without doing any stunt on our part.

For non-Class Type, this has been pretty human readable because JDK's toString() has been pretty human readable (and we keep Guava's Type#toString() consistent with JDK's); for non-Array Class, this has also been pretty readable cause JDK's getName() has been pretty readable.

Unfortunately array Class before Java 8 didn't offer a method that returned human readable string. We could have tried to build a human readable string but then it'd add yet another condition where we differ from the JDK's implementation.

I share Chris's concern about backward compatibility regarding relying on Java 8's getTypeName().

If we must, perhaps adding a new TypeToken#getTypeName() method would be a more backward-compatible change although we need to make sure it carries its own weight compared to the user code simply calling `typeToken.getType().getTypeName()` in Java 8 code.

cpovirk:
If there are cases in which our `Type.toString()` implementations differ from the JDK's, we'd ideally fix that. It's also possible that we could modify `TypeToken.toString()` specifically to behave more like `Type.getTypeName()`. Additionally, could make sure our `Type.getTypeName()` implementations match the JDK implementations. (Maybe they already do.) But these are larger questions than the documentation, which is at least no longer incorrect.

PR #2993, opened 2017-11-13
Changed the ArrayIndexOutOfBoundsException to String...

muteness:
In the repeat method of Strings class, if the length of the potential output string overflows, the exception to be thrown should be StringIndexOutOfBoundsException instead of ArrayIndexOutOfBoundsException.

### DIFF
--- a/guava-tests/test/com/google/common/base/StringsTest.java
+++ b/guava-tests/test/com/google/common/base/StringsTest.java
@@ -126,7 +126,7 @@ public class StringsTest extends TestCase {
       // Massive string
       Strings.repeat("12345678", (1 << 30) + 3);
       fail();
-    } catch (ArrayIndexOutOfBoundsException expected) {
+    } catch (StringIndexOutOfBoundsException expected) {
     }
   }
 
--- a/guava/src/com/google/common/base/Strings.java
+++ b/guava/src/com/google/common/base/Strings.java
@@ -152,7 +152,7 @@ public final class Strings {
     final long longSize = (long) len * (long) count;
     final int size = (int) longSize;
     if (size != longSize) {
-      throw new ArrayIndexOutOfBoundsException("Required array size too large: " + longSize);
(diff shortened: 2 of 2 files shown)


kevinb9n:
We will think about changing this; however I don't think this is the right exception type either. It's not a matter of an index that was greater than the size of an array OR a string.

cpovirk:
We have precedent for throwing `OutOfMemoryError` in `Files`. But then we'd also commented that we _wanted_ to use it in `CharEscaper`/`UnicodeEscaper` but that we couldn't because it's not available under GWT :\

muteness:
I thought it'd be better to have a custom exception for this, but one of the constructors in the String class  (public String(char value[], int offset, int count)) throws StringIndexOutOfBoundsException when count is negative or offset > value.length - count.

cpovirk:
I guess the closest precedent is `StringBuilder`, which also throws `OutOfMemoryError` (though again, not under GWT).

muteness:
I just looked into the StringBuilder implementation. I think it throws OutOfMemoryError when it cannot allocate a capacity for the underlying array.

But for the same purpose as the repeat method, it throws StringIndexOutOfBoundException.

Take the following method as an example.

If `str.length() + count` overflows, it doesn't throw OutOfMemoryError but instead throws StringIndexOutOfBoundException from getChars method.



    public AbstractStringBuilder append(String str) {
        if (str == null)
            return appendNull();
        int len = str.length();
        ensureCapacityInternal(count + len);
        str.getChars(0, len, value, count);
        count += len;
        return this;
    }

    private void ensureCapacityInternal(int minimumCapacity) {
        // overflow-conscious code
        if (minimumCapacity - value.length > 0) {
            value = Arrays.copyOf(value,
                    newCapacity(minimumCapacity));
        }
    }

cpovirk:
`str.length() + count` shouldn't overflow in `getChars`, though, should it? I think that's what the `ensureCapacityInternal` call is there to prevent.

muteness:
I think you're right. Thanks!

Issue #2870, opened 2017-07-11
Add Uninterruptibles support for Condition


ChristianSchwarz:
It would be nice to have Uninterruptibles support for Condition.await() and overlaods.

So we could write

```
lock.lock();
try{
  awaitUninterrutibly(condition,1,MINUTE);
}finally{
 lock.unlock();
}
```

cpovirk:
This is a case where this probably won't be a high priority for us but we'd be happy to take a patch. You might also be interested in [`Monitor`](http://google.github.io/guava/releases/snapshot/api/docs/com/google/common/util/concurrent/Monitor.html), which includes `waitForUninterruptibly`.

cpiotr:
Hey, I've created https://github.com/google/guava/pull/3005 and it would be great, if you could review it.

cpiotr:
Hey, I've created #3010 where I squashed commits into one and it would be great, if you could review it.

cpovirk:
Do you care much about the `Date` (`awaitUntilUninterruptibly`) version included in your CL? Even though we're just mirroring an existing API, we're always a little hesitant to even suggest that `Date` is a good idea, especially for elapsed time.

cpiotr:
I don't care much about this one, to be honest. 
Shall I remove it and repeat the process (squash and send PR)?

ronshapiro:
No need to squash and create a new PR, just add new commits to the PR and we'll handle the squashing

cpovirk:
We make the changes internally, anyway, so I can just remove it from my local copy.

cpiotr:
Sounds good to me.

PR #3005, opened 2017-12-06
Extend Uninterruptibles with methods allowing awaiting on Condition

cpiotr:
Issue: #2870 

Condition class exposes a number of methods to await on it.
Most of them throw exception when executing thread gets interrupted.
In scope of this change utility methods are introduced to serve as uninterruptible proxies.

### DIFF
--- a/guava/src/com/google/common/util/concurrent/Uninterruptibles.java
+++ b/guava/src/com/google/common/util/concurrent/Uninterruptibles.java
@@ -21,6 +21,7 @@ import com.google.common.annotations.GwtCompatible;
 import com.google.common.annotations.GwtIncompatible;
 import com.google.common.base.Preconditions;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
+import java.util.Date;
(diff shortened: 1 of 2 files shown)


cpiotr:
I signed it!

guava/src/com/google/common/util/concurrent/Uninterruptibles.java
```diff
@@ -93,6 +95,62 @@ public static boolean awaitUninterruptibly(CountDownLatch latch, long timeout, T
     }
   }
 
+  /** Invokes {@code condition.}{@link Condition#await() await()} uninterruptibly. */
+  @GwtIncompatible // concurrency
+  public static void awaitUninterruptibly(Condition condition) {
```
ChristianSchwarz: @cpiotr I think this method can be removed, cause it adds no value.

cpiotr: @ChristianSchwarz Good point, removed.

PR #3018, opened 2017-12-24
POC: replace explicit null checks by automatic

denis-zhdanov:
This *PR* illustrates an approach where explicit *null*-checks are replaced by automatic checks.  

Consider a sample class - [Iterators](https://github.com/denis-zhdanov/guava/commit/71eee2dea89f6827f340e231bc715311120df6ca#diff-d2ea06a03a4988cf54d6d3d923deae88):  

*Original source code*  

```java
public static <T> Iterator<T> concat(Iterator<? extends T> a, Iterator<? extends T> b) {
  checkNotNull(a);
  checkNotNull(b);
  return concat(consumingForArray(a, b));
}
```  

*Suggested source code*  

```java
public static <T> Iterator<T> concat(Iterator<? extends T> a, Iterator<? extends T> b) {
  return concat(consumingForArray(a, b));
}
```  

Resulting bytecode looks like if it's compiled from the source below:  

```java
public static <T> Iterator<T> concat(Iterator<? extends T> a, Iterator<? extends T> b) {
  if (a == null) {
    throw new NullPointerException("a was null");
  }
  if (b == null) {
    throw new NullPointerException("b was null");
  }
  return concat(consumingForArray(new Iterator[] { a, b }));
}
```  

*Raw bytecode*  

```
javap -c ./guava/target/classes/com/google/common/collect/Iterators.class
...
  public static <T> java.util.Iterator<T> concat(java.util.Iterator<? extends T>, java.util.Iterator<? extends T>);
    Code:
       0: aload_0
       1: ifnonnull     14
       4: new           #5                  // class java/lang/NullPointerException
       7: dup
       8: ldc           #51                 // String a was null
      10: invokespecial #7                  // Method java/lang/NullPointerException."<init>":(Ljava/lang/String;)V
      13: athrow
      14: aload_1
      15: ifnonnull     28
      18: new           #5                  // class java/lang/NullPointerException
      21: dup
      22: ldc           #52                 // String b was null
      24: invokespecial #7                  // Method java/lang/NullPointerException."<init>":(Ljava/lang/String;)V
      27: athrow
      28: iconst_2
      29: anewarray     #53                 // class java/util/Iterator
      32: dup
      33: iconst_0
      34: aload_0
      35: aastore
      36: dup
      37: iconst_1
      38: aload_1
      39: aastore
      40: invokestatic  #54                 // Method consumingForArray:([Ljava/lang/Object;)Ljava/util/Iterator;
      43: invokestatic  #55                 // Method concat:(Ljava/util/Iterator;)Ljava/util/Iterator;
      46: areturn
```  

**Implementation**  

The checks are generated by the [Traute](http://traute.oss.harmonysoft.tech/) *javac* plugin - it finds [existing package-level *ParametersAreNonnullByDefault*](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/package-info.java#L215) annotation and adds checks for all method parameters not marked by *Nullable* annotation (*Checker*'s *NullableDecl* is [supported by default](http://traute.oss.harmonysoft.tech/core/javac/#73-nullable-annotations)).  

**Restrictions**  

*Traute* is a *javac* plugin and corresponding *API* is available since *java8*. Do you guys use something like `javac8 -source 1.7 -target 1.7` for generating *javac7* bytecode?  

**Proposal**  

I'd be glad to create a *PR*/modify current *PR* which replaces all explicit *null*-checks if the team likes the approach.

### DIFF
--- a/guava/pom.xml
+++ b/guava/pom.xml
@@ -37,10 +37,6 @@
       <artifactId>animal-sniffer-annotations</artifactId>
       <version>${animal.sniffer.version}</version>
     </dependency>
-    <dependency>
-      <groupId>tech.harmonysoft</groupId>
-      <artifactId>traute-javac</artifactId>
-    </dependency>
     <!-- TODO(cpovirk): does this comment belong on the <dependency> in <profiles>? -->
     <!-- TODO(cpovirk): want this only for dependency plugin but seems not to work there? Maven runs without failure, but the resulting Javadoc is missing the hoped-for inherited text -->
   </dependencies>
--- a/guava/pom.xml
+++ b/guava/pom.xml
@@ -37,6 +37,10 @@
       <artifactId>animal-sniffer-annotations</artifactId>
       <version>${animal.sniffer.version}</version>
     </dependency>
+    <dependency>
+      <groupId>tech.harmonysoft</groupId>
+      <artifactId>traute-javac</artifactId>
+    </dependency>
     <!-- TODO(cpovirk): does this comment belong on the <dependency> in <profiles>? -->
     <!-- TODO(cpovirk): want this only for dependency plugin but seems not to work there? Maven runs without failure, but the resulting Javadoc is missing the hoped-for inherited text -->
   </dependencies>
--- a/guava/src/com/google/common/collect/ImmutableSortedMap.java
+++ b/guava/src/com/google/common/collect/ImmutableSortedMap.java
@@ -569,7 +569,7 @@ public final class ImmutableSortedMap<K, V> extends ImmutableSortedMapFauxveride
   ImmutableSortedMap(
       RegularImmutableSortedSet<K> keySet,
       ImmutableList<V> valueList,
-      ImmutableSortedMap<K, V> descendingMap) {
+      @NullableDecl ImmutableSortedMap<K, V> descendingMap) {
     this.keySet = keySet;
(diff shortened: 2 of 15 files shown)


guava/pom.xml
```diff
@@ -37,6 +37,10 @@
       <artifactId>animal-sniffer-annotations</artifactId>
       <version>${animal.sniffer.version}</version>
     </dependency>
+    <dependency>
+      <groupId>tech.harmonysoft</groupId>
+      <artifactId>traute-javac</artifactId>
```
liach: What's this for?

denis-zhdanov: @liach this adds *Traute* *javac* plugin to compiler's classpath

lowasser:
I'm not sure we would want this?  Historically we have resisted "silently generated" code -- e.g. this is why we favor AutoValue over Lombok -- and strongly prefer code to be explicit and traceable.

denis-zhdanov:
@lowasser, thanks for the feedback!

My assumption was that preserving existing contracts (implied by *ParametersAreNonnullByDefault*) and making the code more clear by removing explicit checks might be good for the project.  

Regarding the traceability - exceptions thrown from the failed generated checks reference corresponding parameter line. Also exception message makes it clear what wrong has happened

Issue #3011, opened 2017-12-11
Add containsIgnoreCase to StringUtils


martinfrancois:
I regularly use StringUtils.isEmptyOrNull, inspired by it came the idea for the function "containsIgnoreCase" of StringUtils. It is a function I usually find myself writing over and over again because of its usefulness.

It compares two strings with the "contains" operation, but turns both strings lowercase before doing so (hence the "ignoreCase" aspect of it).

Use cases would be especially search filters. When a search string (in a search field) should be compared to items, this function is particularly useful. Case doesn't matter in most cases and partial results are often desired as well, which is why contains is useful.

Upon acceptance I would happily implement this and submit a pull request.

cpovirk:
For string comparisons, we've taken the position that there are 2 reasonable ways to do them:
- handle ASCII only
- use ICU4J

Guava doesn't depend on ICU4J, so we'd be looking at the ASCII approach here. That's fine; I just want to define the scope.

We do already have the [`Ascii`](http://google.github.io/guava/releases/snapshot-jre/api/docs/com/google/common/base/Ascii.html) class. It contains `equalsIgnoreCase` but not `containsIgnoreCase`.

With `containsIgnoreCase`, unlike `equalsIgnoreCase`, parameter order matters. I think that people would naturally expect `(haystack, needle)`, but it's a small potential problem.

How often are these needed?

- `toLowerCase(...).contains(....toLowerCase())` appears almost twice as often as `toLowerCase(...).equals(....toLowerCase())`. But `Ascii.equalsIgnoreCase` is called more than enough to make up the difference -- and in fact we have static analysis internally that suggests `Ascii.equalsIgnoreCase` in exactly this case :) Still, that's an argument that `contains`, while probably not *more* common than `equals`, is comparably common.

- `toLowerCase(...).contains("...")` is called more than `toLowerCase(...).equals("...")`. (And the literal (`"..."`) case is similarly common to the non-literal case, for whatever that's worth. Such users don't benefit quite as much from a library, since we're substituting `containsIgnoreCase(...)` for only `toLowerCase().contains(...)`, not  `toLowerCase().contains(....toLowerCase())`.)

[...]

PR #3023, opened 2018-01-03
base.Ascii: Add methods indexOfIgnoreCase, containsIgnoreCase, startsWithIgnoreCase and endsWithIgnoreCase

martinfrancois:
Implements the changes proposed in issue #3011
I was not sure in which version this is going to be pulled in - so I left `// TODO` comments at the `@since` tags. Feel free to tell me which number to change it to 😃


martinfrancois:
I signed it!

guava/src/com/google/common/base/Ascii.java
```diff
@@ -629,4 +630,78 @@ private static int getAlphaIndex(char c) {
...
+  /**
+   * Returns the index within {@code sequence} of the first occurrence of the
+   * specified {@code subSequence}, ignoring the case of any ASCII alphabetic characters
+   * between {@code 'a'} and {@code 'z'} or {@code 'A'} and {@code 'Z'} inclusive, or
+   * {@code -1}, if no occurence is found.
+   *
+   * @since 16.0 // TODO: which number?
+   */
+  public static int indexOfIgnoreCase(CharSequence sequence, CharSequence subSequence) {
+    return toLowerCase(sequence).indexOf(toLowerCase(subSequence));
```
liach: I fear this implementation may be too slow: If the length of `sequence` is `n`, the length of `subSequence` is `m`, then the time complexity is always O(n+m) or higher, because all the chars need to be converted into a string first; if we can change a letter to lower case one by one, then the execution time may be faster.

guava/src/com/google/common/base/Ascii.java
```diff
@@ -629,4 +630,78 @@ private static int getAlphaIndex(char c) {
...
+   * @since 16.0 // TODO: which number?
+   */
+  public static boolean startsWithIgnoreCase(CharSequence seq, CharSequence prefix, int offset) {
+    // an empty prefix starts everywhere in the sequence
+    if (prefix.length() == 0) {
+      return true;
+    }
+    // if the prefix is longer than the sequence itself or from the offset, it is impossible
+    // for the sequence to contain the prefix
+    if (prefix.length() > seq.length() || prefix.length() > seq.length() - offset) {
```
liach: You can just put `if (prefix.length() > seq.length() - offset)` unless someone maliciously passed a negative number for `offset` parameter.

guava/src/com/google/common/base/Ascii.java
```diff
@@ -629,4 +630,78 @@ private static int getAlphaIndex(char c) {
...
     return (char) ((c | 0x20) - 'a');
   }
+
+  /**
+   * Returns the index within {@code sequence} of the first occurrence of the
+   * specified {@code subSequence}, ignoring the case of any ASCII alphabetic characters
+   * between {@code 'a'} and {@code 'z'} or {@code 'A'} and {@code 'Z'} inclusive, or
+   * {@code -1}, if no occurence is found.
+   *
+   * @since 16.0 // TODO: which number?
```
liach: You can change these to `@since NEXT`

martinfrancois: Thanks for letting me know! I will update them.

guava/src/com/google/common/base/Ascii.java
martinfrancois: Good point. I didn't initially think of this, but I'll definitely have a look into it.

guava/src/com/google/common/base/Ascii.java
martinfrancois: I had it like that initially but in the case of seq as empty string and prefix as `"x"` it will result in a negative prefix, which lead me to this implementation. Should I check this in another way?

liach: then this is totally fine.

guava/src/com/google/common/base/Ascii.java
```diff
@@ -629,4 +630,78 @@ private static int getAlphaIndex(char c) {
...
+    // an empty prefix starts everywhere in the sequence
+    if (prefix.length() == 0) {
+      return true;
+    }
+    // if the prefix is longer than the sequence itself or from the offset, it is impossible
+    // for the sequence to contain the prefix
+    if (prefix.length() > seq.length() || prefix.length() > seq.length() - offset) {
+      return false;
+    }
+    return indexOfIgnoreCase(seq, prefix) == offset;
```
Maaartinus: What do you expect `startsWithIgnoreCase("aaa", "a", 1)` to return? I'd vote for true, but this is impossible to implement without a `indexOfIgnoreCase(seq, prefix, offset)`.

guava/src/com/google/common/base/Ascii.java
```diff
@@ -629,4 +630,78 @@ private static int getAlphaIndex(char c) {
...
+  }
+
+  /**
+   * Returns if the character sequence {@code seq} starts with the character sequence {@code prefix}
+   * at the specified {@code offset}, ignoring the case of any ASCII alphabetic characters
+   * between {@code 'a'} and {@code 'z'} or {@code 'A'} and {@code 'Z'} inclusive.
+   *
+   * @since 16.0 // TODO: which number?
+   */
+  public static boolean startsWithIgnoreCase(CharSequence seq, CharSequence prefix, int offset) {
```
Maaartinus: A negative `offset` should probably be forbidden.

guava/src/com/google/common/base/Ascii.java
```diff
@@ -629,4 +630,78 @@ private static int getAlphaIndex(char c) {
...
+
+  /**
+   * Returns if the character sequence {@code seq} starts with the character sequence {@code prefix}
+   * ignoring the case of any ASCII alphabetic characters
+   * between {@code 'a'} and {@code 'z'} or {@code 'A'} and {@code 'Z'} inclusive.
+   *
+   * @since 16.0 // TODO: which number?
+   */
+  public static boolean startsWithIgnoreCase(CharSequence seq, CharSequence prefix) {
+    return startsWithIgnoreCase(seq, prefix, 0);
```
Maaartinus: This is much less efficient than a direct check, e.g., for `startsWithIgnoreCase("aaaaaaaaaaaaaaaaa", "b")`.

guava/src/com/google/common/base/Ascii.java
liach: This one is 
1. logically flawed
2. Too slow (You should just use a for loop to check. If the length of prefix is m, then the time complexity is O(m) maximum)

guava/src/com/google/common/base/Ascii.java
liach: If the last one is fixed, then it is fine here.

guava/src/com/google/common/base/Ascii.java
```diff
@@ -629,4 +630,155 @@ private static int getAlphaIndex(char c) {
...
+   * @param string    the string to be searched in.
+   * @param subString the substring to search for.
+   * @param fromIndex the index from which to start the search.
+   * @return the index of the first occurrence of the {@code subString}, or {@code -1} if there is
+   *     no such occurrence.
+   *
+   * @since NEXT
+   */
+  public static int indexOfIgnoreCase(String string, String subString, int fromIndex) {
+    return indexOfIgnoreCase(string.toCharArray(), 0, string.length(),
```
liach: `toCharArray()` involves a linear time copy of array. It is slow, too.

guava/src/com/google/common/base/Ascii.java
```diff
@@ -629,4 +630,155 @@ private static int getAlphaIndex(char c) {
...
+    // Calling length() is the null pointer check (so do it before we can exit early).
+    int length = sequence.length();
+    if (sequence == subSequence) {
+      return true;
+    }
+    // if subSequence is longer than sequence, it is impossible for sequence to contain subSequence
+    if (subSequence.length() > length) {
+      return false;
+    }
+    return indexOfIgnoreCase(sequence.toString(), subSequence.toString()) > -1;
```
liach: Still should not `toString`

guava/src/com/google/common/base/Ascii.java
```diff
@@ -629,4 +630,155 @@ private static int getAlphaIndex(char c) {
...
+    // an empty prefix starts everywhere in the sequence
+    if (prefix.length() == 0) {
+      return true;
+    }
+    // if the prefix is longer than the sequence itself or from the offset, it is impossible
+    // for the sequence to contain the prefix
+    if (prefix.length() > seq.length() || prefix.length() > seq.length() - offset) {
+      return false;
+    }
+    return indexOfIgnoreCase(seq.toString(), prefix.toString()) == offset;
```
liach: This one is
1. Logically flawed
2. Too slow (A single for loop can fix it)

guava/src/com/google/common/base/Ascii.java
```diff
@@ -629,4 +630,155 @@ private static int getAlphaIndex(char c) {
...
+      return fromIndex;
+    }
+
+    char first = toLowerCase(target[targetOffset]);
+    int max = sourceOffset + (sourceCount - targetCount);
+
+    for (int i = sourceOffset + fromIndex; i <= max; i++) {
+      /* Look for first character. */
+      if (toLowerCase(source[i]) != first) {
+        while (++i <= max && toLowerCase(source[i]) != first) {
```
liach: Can you change it to a do-while with increment of i in the loop? that way the code look cleaner.

guava/src/com/google/common/base/Ascii.java
```diff
@@ -629,4 +630,155 @@ private static int getAlphaIndex(char c) {
...
+      return (targetCount == 0 ? sourceCount : -1);
+    }
+    if (fromIndex < 0) {
+      fromIndex = 0;
+    }
+    if (targetCount == 0) {
+      return fromIndex;
+    }
+
+    char first = toLowerCase(target[targetOffset]);
```
liach: You should just use getAlphaIndex to make things efficient

guava/src/com/google/common/base/Ascii.java
```diff
@@ -629,4 +630,155 @@ private static int getAlphaIndex(char c) {
...
+   *
+   * @param source       the characters being searched.
+   * @param sourceOffset offset of the source string.
+   * @param sourceCount  count of the source string.
+   * @param target       the characters being searched for.
+   * @param targetOffset offset of the target string.
+   * @param targetCount  count of the target string.
+   * @param fromIndex    the index to begin searching from.
+   */
+  private static int indexOfIgnoreCase(char[] source, int sourceOffset, int sourceCount,
```
liach: You can copy the whole method, changing two char array arguments to char sequence arguments so that you can deal with strings and string builders directly without `toCharArray` calls that duplicate all the char contents.

guava/src/com/google/common/base/Ascii.java
martinfrancois: Thanks, I totally forgot about that. I fixed it now.

guava/src/com/google/common/base/Ascii.java
martinfrancois: Just fixed that one by implementing the toLowerCase() call in the for-loop itself

guava/src/com/google/common/base/Ascii.java
martinfrancois: Sounded like a good idea at first but now I agree - you guys are 100% correct. Will fix that.

guava/src/com/google/common/base/Ascii.java
```diff
@@ -644,8 +644,8 @@ private static int getAlphaIndex(char c) {
    * @param targetCount  count of the target string.
    * @param fromIndex    the index to begin searching from.
    */
-  private static int indexOfIgnoreCase(char[] source, int sourceOffset, int sourceCount,
-                                       char[] target, int targetOffset, int targetCount,
+  private static int indexOfIgnoreCase(String source, int sourceOffset, int sourceCount,
```
liach: Charsequense instead

guava/src/com/google/common/base/Ascii.java
```diff
@@ -700,8 +700,8 @@ private static int indexOfIgnoreCase(char[] source, int sourceOffset, int source
    * @since NEXT
    */
   public static int indexOfIgnoreCase(String string, String subString, int fromIndex) {
```
liach: String ->CharSequence

guava/src/com/google/common/base/Ascii.java
martinfrancois: Thanks for noticing! I fixed it in the latest commit.

guava/src/com/google/common/base/Ascii.java
```diff
@@ -629,4 +630,160 @@ private static int getAlphaIndex(char c) {
...
+
+    char first = toLowerCase(target.charAt(targetOffset));
+    int max = sourceOffset + (sourceCount - targetCount);
+
+    for (int i = sourceOffset + fromIndex; i <= max; i++) {
+      /* Look for first character. */
+      if (toLowerCase(source.charAt(i)) != first) {
+        do {
+          ++i;
+        } while(i <= max && toLowerCase(source.charAt(i)) != first);
```
liach: Can you replace all `toLowerCase` calls to `getAlphaIndex` calls for efficiency?

martinfrancois: will do!

martinfrancois: It was quite some work but now I changed the implementation to use `getAlphaIndex` instead of `toLowerCase`. Please have a look again. If everything is fine, I'll squash the commits.

liach:
Thanks for this great pull request!

martinfrancois:
You're welcome 😃  thanks a lot, too, for the very thorough and good code reviews of you all, they were very constructive and helpful.
I now squashed the commits. Maybe one of you Googlers can run the benchmark (I hope I did it right!)? I didn't see any possibility to run them myself. Would be interesting to see the performance differences. Besides that, from my side it's good to merge.

martinfrancois:
Hey there, just wanted to ask if there's any news concerning this pull request?

jbduncan:
@martinfrancois I'm not on the Guava team myself, so I can't truly speak on their behalf, but as someone who's contributed to the project for a couple of years now, I think it's worth letting you know that they can be very slow addressing new issues and PRs, which I think happens because they're busy as a whole dealing with Google-internal priorities.

So if they take forever to respond (which seems likely at this stage, sadly), it absolutely won't be because you did something wrong; they're just apparently very, _very_ busy with their day-to-day jobs, and someone from the team will eventually respond to this PR. And if you don't get a response as quickly as you like, writing a reminder message like the one you wrote [here](https://github.com/google/guava/pull/3023#issuecomment-357791229) every now and then should encourage a more timely response, even if they're a bit too busy to review the PR properly.

I hope this helps! :)

cpovirk:
There are various reasons that we haven't done a good job about replying to issues. Busyness is part of it, but we should do better. It looks like our current triage process misses pull requests. I'll try to get that fixed.

martinfrancois:
@jbduncan thanks for the very kind words and the tips 😃 I appreciate that a lot, it's good to know that it's not an ununsual case.

@cpovirk also thanks for taking it serious and trying to get pull requests integrated into your triage process 👍

martinfrancois:
Thanks for accepting my pull request! I wanted to ask when the changes of it are going to get synced out? I haven't seen it on the master yet.

martinfrancois:
Any news on this PR?

kevinb9n:
I'm sorry that we are so slow on this. Fundamentally we appreciate the request, but we have a bit of work to do on our end that we haven't been able to prioritize yet.

FYI, I do see a pretty significant amount of Google code writing stuff like `if (s.toLowerCase().startsWith(...))` so I do think the case for adding these methods to Guava may be a pretty good one.

martinfrancois:
@kevinb9n thanks for the update on this! It has now been almost half a year, are there any internal processes it still needs to go through or did the team just not get around to merging it into the codebase yet?
Thanks!

cgdecker:
I want to quickly note here that when an issue or PR is marked "triaged" and assigned to someone, that doesn't necessarily mean it's accepted. Triaged just means that someone has looked at it and thinks it's well-formed and worth thinking about more, and assigned it to someone who they think should best be able to consider it. That said, Kevin has noted that this seems like something that there's a good chance we'll want to do.

Another note: API additions do generally need to go through an API review process we have internally, which tends to involve writing up a doc exploring pros and cons of an API, basic statistics for how often existing code in our codebase would want to use the API(s), options for how to address the problem, etc., followed by scheduling it to be discussed with a group of reviewers. That's obviously a bit of work, so even when you've written the code, it can be hard to fit it in with everything else we're doing. That's not to say we can't do better--we can and should do better, and are working on improving our processes around bugs/PRs--but just to give some idea of why things might take so long sometimes.

martinfrancois:
Thanks a lot @cgdecker! I tried to find a description of your process but was unable to. Your description made it very clear to me. Now I understand what goes into it and it explains the long process, but I do appreciate your attention to detail and sense of quality, it gives me an even greater confidence than I already had about the guava code base.

martinfrancois:
Hi everyone,
I don't want to sound impatient, but it's been almost a year now. Is there any news on this one?
Thanks,
François

maehly:
François Martin has personally and elaborately explained to me this PR. Actually he was so kind to copy over the implementation into a project we are working on together. I vote for merging this PR.

martinfrancois:
@kevinb9n or @cgdecker could I please get an update on the current status of your internal review process concerning this PR?

PR #3027, opened 2018-01-08
Restrict length of varargs array in immutable collection factory methods

### LINKED ISSUE #3026: NegativeArraySizeException in collection factory methods
The immutable collections `ImmutableDoubleArray`, `ImmutableLongArray`, `ImmutableIntArray`, `ImmutableList`, and `ImmutatbleSet` all provide methods called `of`.

If an array of length near `Integer.MAX_VALUE` is passed as the last argument, an unexpected `NegativeArraySizeException` is thrown.

To reproduce, run this code on a computer with enough RAM:

```java
import com.google.common.collect.ImmutableList;

public class NegArrSize {
     Object[] array = new Object[Integer.MAX_VALUE - 11];
     ImmutableList.of(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, array);
   }
}
```

The program throws `NegativeArraySizeException`.

The problem is that these methods do an unchecked addition to compute the size of an array that will be allocated. The methods work for most argument array sizes, but it would be helpful to document and check the restriction on the argument.

panacekcz:
This PR documents the restriction that immutable collection factory methods with varargs pose on the length of the last argument, and adds a check for this precondition to prevent `NegativeArraySizeException` from being thrown.

Fixes #3026.

### DIFF
--- a/guava/src/com/google/common/collect/ImmutableList.java
+++ b/guava/src/com/google/common/collect/ImmutableList.java
@@ -16,6 +16,7 @@
 
 package com.google.common.collect;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkElementIndex;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Preconditions.checkPositionIndexes;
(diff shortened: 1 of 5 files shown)


guava/src/com/google/common/collect/ImmutableList.java
```diff
@@ -190,12 +191,18 @@
...
+   *
    * @throws NullPointerException if any element is null
    * @since 3.0 (source-compatible since 2.0)
    */
   @SafeVarargs // For Eclipse. For internal javac we have disabled this pointless type of warning.
   public static <E> ImmutableList<E> of(
       E e1, E e2, E e3, E e4, E e5, E e6, E e7, E e8, E e9, E e10, E e11, E e12, E... others) {
+    checkArgument(
+        others.length <= Integer.MAX_VALUE - 12,
+        "others.length (%s) must be <= Integer.MAX_VALUE - 12",
```
lowasser: I'd prefer just saying here that the total number of elements must fit in an int, here and everywhere else.

panacekcz:
@lowasser, I have changed the messages.

panacekcz:
Can I ask what is the status of this? Are you willing to merge this PR or do you want me to do some changes?

lowasser:
Working on merging it, there's some shenanigans and I haven't gotten to it yet.

PR #3045, opened 2018-02-15
Change .orderedBy() builders generic type to <? super>

tgyurci:
Using `.orderedBy(...)` over `new Builder<>(...)` is much more compact.

It compiles on both Java 7, 8 and 9.  It is source-compatible change, but I am not sure, whether it is fully binary-compatible or not.

### DIFF
--- a/android/guava-tests/test/com/google/common/collect/ImmutableSortedMapTest.java
+++ b/android/guava-tests/test/com/google/common/collect/ImmutableSortedMapTest.java
@@ -861,4 +861,20 @@ public class ImmutableSortedMapTest extends TestCase {
     ImmutableSortedMap.Builder<SuperComparableExample, Object> reverse =
         ImmutableSortedMap.reverseOrder();
   }
+
+  private static final Comparator<Object> TO_STRING =
+      new Comparator<Object>() {
+        @Override
+        public int compare(Object o1, Object o2) {
+          return o1.toString().compareTo(o2.toString());
+        }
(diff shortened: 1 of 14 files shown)


tgyurci:
I signed it!

cpovirk:
`MinMaxPriorityQueue.Builder` doesn't need this because its `build()` method lets you create a queue for subtype of the comparator's type: `<T extends B> MinMaxPriorityQueue<T> create()`.

As for the immutable types: They're this way because, in 2008:

> Add the "? super" make you need a ".<String>" in this line:
> 
>     SortedSet<String> set = ImmutableSortedSet.<String>orderedBy(STRING_LENGTH).of(
>         "in", "the", "quick", "jumped", "over", "a");
> 
> If the types are the same, that generic parameter isn't needed.

Java's type inference has improved in the last 10 years, so perhaps we no longer need this. Can you add a test like the above (without `<String>`) to confirm that it works now?

tgyurci:
I reverted the changes in MinMaxPriorityQueue but added some test for it's `.orderedBy()` too.

cpovirk:
Sorry, the test I was asking about is a little different than the ones you added (but those ones are good to have, too, so thanks). I'll add the test locally.

cpovirk:
The new tests break under GWT :( That might mean they break under various versions of Eclipse, too.

We should try this again someday, but for now, we'll leave it the way it is.

PR #3061, opened 2018-03-08
Add lcm method for big integers

srinivascodes:
`/*Returns least common multiple of two big integers using gcd method*/`


kevinb9n:
Hi - we will discuss adding lcm methods and get back to you. (Personally I want to add them everywhere there is gcd, but we don't have overwhelming evidence that it's actually needed much, except for people all implementing their own `Rational` class.)

srinivascodes:
I have signed the e-cla.

srinivascodes:
I signed

Issue #3060, opened 2018-03-08
public static BigInteger lcm(BigInteger a, BigInteger b)


srinivascodes:
/*Returns the least common multiple of two BigIntegers using gcd method.*/

cgdecker:
(This proposal seems to be for `BigIntegerMath`, see PR #3063.)

Louis, any thoughts? It seems likely to me that there would be very few usages for this method... the existing methods on `BigIntegerMath` don't have many usages to begin with.

kevinb9n:
I'm a bit motivated to write up a quick proposal.

nikhilbarar:
Is this open? Can i pick it?

kevinb9n:
Sorry, there's nothing to do on this right now.

PR #3063, opened 2018-03-08
Add LCM method for BigIntegers

srinivascodes:
`/*Returns least common multiple of two BigIntegers using gcd method*/`

### DIFF
--- a/guava/src/com/google/common/math/BigIntegerMath.java
+++ b/guava/src/com/google/common/math/BigIntegerMath.java
@@ -22,6 +22,7 @@ import static com.google.common.math.MathPreconditions.checkRoundingUnnecessary;
 import static java.math.RoundingMode.CEILING;
 import static java.math.RoundingMode.FLOOR;
 import static java.math.RoundingMode.HALF_EVEN;
+import static java.math.RoundingMode.UNNECESSARY;
 
 import com.google.common.annotations.Beta;
 import com.google.common.annotations.GwtCompatible;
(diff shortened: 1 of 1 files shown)


perceptron8:
@srinivascodes I suspect this PR won't be merged anyway, so here's my two cents:

- Do not check for equality using `==`. Use `equals` instead. It's possible to construct `BigInteger` equal to `BigInteger.ZERO` but not being the same instance as `BigInteger.ZERO`. One of many possibilities is to call `new BigInteger(0, new byte[0])`.
- First divide, then multiply. It's more efficient. As rounding is indeed `UNNECESSARY`, don't use it at all (it is sufficient to use `BigInteger.divide`).
- There's no need to call `abs()` more than once.

```
public static BigInteger lcm(BigInteger a, BigInteger b) {
	if (a.equals(BigInteger.ZERO) || b.equals(BigInteger.ZERO)) {
		return BigInteger.ZERO;
	}
	BigInteger gcd = a.gcd(b);
	return a.divide(gcd).multiply(b).abs();
}
```

srinivascodes:
@perceptron8 Thanks for your valuable advice. Changed my code's implementation. 
By the way, why do you suspect that this PR won't be merged? Is it because, there is no requirement of lcm method or is it like the implementation is not good enough?

jbduncan:
@srinivascodes I can't speak for @perceptron8, but I suspect that from reading [this comment](https://github.com/google/guava/pull/3061#issuecomment-371537843) from one of the Guava team members, and also from what I personally know of the Guava team, that they will want to go through their rigorous internal API review process first, which will probably mean that you'd need to reimplement your efforts from scratch by the time they're ready to give feedback, if they haven't gone and implemented it themselves inside Google.

It's for that reason that they suggest in their contribution guidelines that [people who want new APIs should raise an issue first and that code should come afterwards](https://github.com/google/guava/blob/master/CONTRIBUTING.md#api-changes). :)

jbduncan:
I hope that this sheds some light on things for you. ;)

perceptron8:
@srinivascodes My intuition was based mostly on contribution guidelines that @jbduncan mentioned. BTW you may still like to include some tests.
> 1. Any significant changes should be accompanied by tests. The project already has good test coverage, so look at some of the existing tests if you're unsure how to go about it.

cgdecker:
I'm going to close this for now; there's already an issue (#3060) open requesting that this be added and until we decide that it's something we want to do, having a PR with code for it doesn't help much. Writing the code is generally the easy part.

PR #3071, opened 2018-03-14
static method of print/println with formatting

liweys:


cgdecker:
Sorry, this isn't something I see us wanting to add to Guava.

liweys:
This is really one of the most frequently used functions of Java. I think it will be good if you guys do something with it.

jbduncan:
@liweys I can't speak for the Guava team, but I can think of two reasons why short-hand methods for `System.out::println` do not add enough value to meet Guava's ["utility times ubiquity"](https://github.com/google/guava/wiki/PhilosophyExplained#when-in-doubt) bar:

1. IDEs (or IntelliJ IDEA, at least) have code completion function(s) which allow `System.out.println(...)` to be typed easily. For example, if you type `sout` in IntelliJ IDEA and hit Enter, it will automatically expand to `System.out.println()`.

   Thus, providing short-hand functions would probably not save any time in writing and especially reading of code (the most important of the two), as most if not every Java programmer understands `System.out::println`.

2. We are in a time where a lot of Java applications are server-side and web applications, where loggers are a lot more useful that "println" statements (as messages written to loggers can be redirected to logs and then interrogated later for debugging purposes).

I hope this sheds some light on things.

PR #3076, opened 2018-03-20
HashBiMap: unnecessary call to rehashIfNecessary

fxleyu:
```java
if (oldEntryForKey != null) {
      delete(oldEntryForKey);
      insert(newEntry, oldEntryForKey);
      oldEntryForKey.prevInKeyInsertionOrder = null;
      oldEntryForKey.nextInKeyInsertionOrder = null;
     rehashIfNecessary();
     return oldEntryForKey.value;
}
```
After to insert the new key-value, the map's size will not increase. So, the `rehashIfNecessary();` is not necessary .


fxleyu:
I signed it!

TeemoJUN:
` oldEntryForKey.prevInKeyInsertionOrder = null;oldEntryForKey.nextInKeyInsertionOrder = null;` why `prevInKeyInsertionOrder ,nextInKeyInsertionOrder` set to null.but `nextInKToVBucket,nextInVToKBucket` not set to null

Issue #2999, opened 2017-11-28
Add overloaded expireAfterWrite(), refreshAfterWrite() and expireAfterAccess() methods to CacheBuilder class


Shenker93:
According to p.1 "API changes require discussion, use cases, etc. Code comes later.", I'm going to start discussion about benefits of proposed changes. 

Now com.google.common.cache.CacheBuilder contains such methods as
`refreshAfterWrite(long duration, TimeUnit unit)`
`expireAfterWrite(long duration, TimeUnit unit)`
`expireAfterAccess(long duration, TimeUnit unit)`
Parameters of methods listed below are used to define durations in time.
However, in Java 8 new class,  [java.time.Duration](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html), was presented. It can be used to save info about time durations in exactly one object. That is easier and more transparent in some real-world scenarios. 

The idea is to add overloaded methods:
`refreshAfterWrite(Duration duration)
expireAfterWrite(Duration duration)
expireAfterAccess(Duration duration)`

Proposed changes provide possibilities to simplify code:
`expireAfterWrite(Duration.ofHours(3))
// instead of
expireAfterWrite(3, TimeUnit.HOURS)
`
It makes easier to declare constant durations too:

`public static final Duration EXPIRES_IN = Duration.ofHours(1);
// some code after statement
expireAfterWrite(EXPIRES_IN)
`
instead of
`public static final long DURATION_LENGTH = 1;
 public static final TimeUnit DURATION_TIME_UNIT = TimeUnit.HOURS;
 // some code after statement
 expireAfterWrite(DURATION_LENGTH, DURATION_TIME_UNIT)
`

Also, it allows to execute some arifmetic operations like addition or subtraction, that can simplify program logic if multiple (and dependant) durations in different methods and/or in different caches are used. Just something like this (simplified synthetic example):
`
Duration expiresAfterAccessIn = Duration.ofMinutes(30);
CacheBuilder.newBuilder()
                    .expireAfterAccess(expiresAfterAccessIn)
                    .expireAfterWrite(expiresAfterAccessIn.plusHours(1))
                    .build();
`

This improvements do not lead to changing min java sdk version and do not break existing API.

fdesu:
If we would agree on e.g. `expireAfterWrite(long)` time unit, it would also be a one-arg method with calls like `expireAfterWrite(TimeUnit.SECONDS.toMillis(60))`. Maybe that's a better way to go?

[...]

PR #3083, opened 2018-03-24
Adds Overloaded Methods to CacheBuilder

slyr91:
Address #2999
I have overloaded the methods listed in the issue so that they accept Duration as a parameter. One function I couldn't quite implement without a bunch of additional code was if the value was negative the error message would return the value + the unit. I cut the unit variable out and replaced it with nanoseconds. The value is also converted to nano.

### DIFF
--- a/guava/src/com/google/common/cache/CacheBuilder.java
+++ b/guava/src/com/google/common/cache/CacheBuilder.java
@@ -32,6 +32,7 @@ import com.google.common.cache.LocalCache.Strength;
 import com.google.errorprone.annotations.CheckReturnValue;
 import java.lang.ref.SoftReference;
 import java.lang.ref.WeakReference;
+import java.time.Duration;
 import java.util.ConcurrentModificationException;
 import java.util.IdentityHashMap;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


slyr91:
I signed it!

slyr91:
Also, this is my very first contribution to real world code. Constructive feedback is always welcome and if you hate it there will be no hard feelings. 
As I was signing that CLA it looks like the Travis app failed but I don't understand why. I only modified the one file and it was all contained to those 3 methods and the imports. Nothing complicated.

slyr91:
This Travis failure appears to be related to travis-ci/travis-ci#9379
A possible fix is to ignore the java8 install temporarily until they resolve the issue.

guava/src/com/google/common/cache/CacheBuilder.java
```diff
@@ -650,6 +651,35 @@ Strength getValueStrength() {
...
+   * in the class javadoc.
+   *
+   * @param duration the length of time after an entry is created that it should be automatically
+   *                 removed
+   * @return this {@code CacheBuilder} instance (for chaining)
+   * @throws IllegalArgumentException if {@code duration} is negative
+   * @throws IllegalStateException is the time to live or time to idle was already set
+   */
+  public CacheBuilder<K, V> expireAfterWrite(Duration duration) {
+    checkState(
```
kluever: Unless we're going to plumb the Duration instance all the way through the cache internals (which would require a serious re-write), then I think we should just call `expireAfterWrite(duration.toNanos(), TimeUnit.NANOSECONDS)` here.

Ditto for the other overloads.

slyr91: Your right, I didn't see that at first. That makes way more sense. I will make the modification soon.

kluever: I'm about to submit a change to add these overloads, as well as updates for all of the tests (so this pull request can probably be closed).
Thanks for the push in the right direction!

ben-manes: I noticed that we don't check for overflow after ns conversion. I'm planning on fixing that sometime soon in Caffeine, just in case. You might want to bundle that into your CL.

kluever: I'm not quite sure I follow.

I was planning on just letting `duration.toNanos()` blow up if passed > ~292 years (and document as such)

ben-manes: sorry, I meant for the `TimeUnit` one being delegated to. I thought one could write `expireAfterWrite(Long.MAX_VALUE, TimeUnit.DAYS)` without it blowing up due to the negative check preceding the conversion. But it looks like the conversion doesn't overflow and saturates instead.
```
java> TimeUnit.DAYS.toNanos(Long.MAX_VALUE)
java.lang.Long res1 = 9223372036854775807
```

kluever: Yep, the saturation behavior of `TimeUnit.toFoo()` is extremely surprising...I'm shocked it hasn't caused more bugs actually :-)

PR #3090, opened 2018-03-28
Fix broken badge in Readme

hrldcpr:
maven-badges.herokuapp.com is frequently (and currently) broken.


cpovirk:
Thanks.

The _new_ badge just showed up as broken for me, but it worked after a refresh. Hopefully it will prove more reliable over time.

hrldcpr:
🤦‍♂️  haha yeah I noticed that too. Seems fine now…

I think it should be at least as reliable as before, since `maven-badges.herokuapp.com` also depends on `shields.io`

PR #3107, opened 2018-04-09
Fix toString of NonSerializableMemoizingSupplier

alicederyn:
Delegate is set to null once the value has been computed, but is still referenced by the toString method. When it's null, use the value instead.


alicederyn:
First solution I thought of

PR #3111, opened 2018-04-12
Add Queues.reversedDeque(Deque)

### LINKED ISSUE #3110: Queues.descending(Deque)
Currently, to use `Deque.descendingIterator()`, we have to write code like this:

    for(Iterator<T> it = myDeque.descendingIterator(); it.hasNext(); ) {
      T element = it.next();
      // Do something
    }

Would it be possible to have something akin to the following:

    for(T element: Queues.descending(myDeque)) {
      // Do something
    }

Ideally the method content would be something like:

    static <T> Iterable<T> descending(final Deque deque) {
      checkNotNull(deque);
      return ()-> deque.descendingIterator();
    }

ogregoire:
Fixes #3110

### DIFF
--- a/guava/src/com/google/common/collect/Queues.java
+++ b/guava/src/com/google/common/collect/Queues.java
@@ -428,7 +428,7 @@ public final class Queues {
    * @param deque the deque to be wrapped in a reversed view
    * @return a reversed view of the specified deque
    */
-  public static <E> Iterable<E> reversedDeque(Deque<E> deque) {
+  public static <E> Deque<E> reversedDeque(Deque<E> deque) {
     if (deque instanceof ReversedDeque) {
       return ((ReversedDeque<E>)deque).deque;
     }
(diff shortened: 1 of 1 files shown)


guava/src/com/google/common/collect/Queues.java
```diff
@@ -419,4 +421,181 @@ private Queues() {}
...
+  
+  /**
+   * Creates a reversed view of {@code deque}.
+   * 
+   * @param deque the deque to be wrapped in a reversed view
+   * @return a reversed view of the specified deque
+   */
+  public static <E> Deque<E> reversedDeque(Deque<E> deque) {
+    if (deque instanceof ReversedDeque) {
+      return ((ReversedDeque<E>)deque).deque;
```
liach: If you access this field, you should declare the deque field package private to prevent generation of a synthetic accessor.

Maaartinus: Or use `delegate()`, which is already accessible as it's `protected`.

jbduncan: Or, even better I imagine, call `delegate()` instead? :)

jbduncan: @Maaartinus got there before me. 😜

cpovirk:
Closing per https://github.com/google/guava/issues/3110#issuecomment-2197409380

PR #3117, opened 2018-04-18
Moe Sync

ronshapiro:

Commits:
=====
<p> Fix toString of NonSerializableMemoizingSupplier

Fixes #3107

9185822ee16cdd641657c0f8e7ead755c9e9a564

-------

<p> Use Splitter instead of String.split, in InternetDomainName.

e56f1aa3d21b04eed8f5481d4103c05af5c51ec7

-------

<p> Use Splitter for IPv6, too. This eliminates usage of java.util.regex from this file (which is used via String.split, otherwise).

65a2a7de4ef4eea913f769a5cbbce6e35e1fccce

-------

<p> Use Diamond.

27eb336c2a2d9068878edbf49edcca493cd69cd8

-------

<p> Automated rollback of a12ef6b46b26fe178021c099e3db6a8a99f5174e.

It looks like the original change can cause an infinite loop in LocalCache.LoadingValueReference.get().

c2266a6ff8c6f298673688d6779b8b7368fe9147

-------

<p> Don't enable use of Unsafe.getLong() on aarch64 devices.

Android devices may report to run aarch64 while running in 32-bit mode, and then
crash when loading 64-bit values at unaligned addresses.

263bbcf6cf0cfae6768ddf9e9f3d749da555e422

### DIFF
--- a/android/guava-testlib/src/com/google/common/testing/GcFinalization.java
+++ b/android/guava-testlib/src/com/google/common/testing/GcFinalization.java
@@ -66,7 +66,7 @@ import java.util.concurrent.TimeoutException;
  * <p>Here's an example that uses a user-defined finalization predicate:
  *
  * <pre>{@code
- * final WeakHashMap<Object, Object> map = new WeakHashMap<Object, Object>();
+ * final WeakHashMap<Object, Object> map = new WeakHashMap<>();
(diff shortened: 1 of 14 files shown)


ronshapiro:
CLA verified in #3107

PR #3123, opened 2018-04-23
Remove references to @DoNotMock

cushon:
which is not longer part of open source Error Prone.


ooxi:
Shouldn't at least the comment be kept?

PR #3127, opened 2018-04-24
Use @SafeVarargs annotation for Sets#cartesianProduct

### LINKED ISSUE #2724: Sets.cartesianProduct(Set<? extends B>... sets) is missing @SafeVarargs annotation
`Lists.cartesianProduct(List<? extends B>... lists)` has it, so I assume Sets version should also have it.

sagar-shah94:
We already had `@SafeVarargs` annotation for `Lists.cartesianProduct(List<? extends B>... lists`. but `Sets.cartesianProduct(Set<? extends B>... sets) ` was missing this annotation so i have added that annotation in order to fix issue.
Fixes #2724

### DIFF
--- a/guava/src/com/google/common/collect/Sets.java
+++ b/guava/src/com/google/common/collect/Sets.java
@@ -1380,6 +1380,7 @@ public final class Sets {
    *     provided set is null
    * @since 2.0
    */
+  @SafeVarargs
   public static <B> Set<List<B>> cartesianProduct(Set<? extends B>... sets) {
     return cartesianProduct(Arrays.asList(sets));
   }


sagar-shah94:
I signed it!

cgdecker:
Thanks, your change has been submitted internally and will be mirrored out soon.

sagar-shah94:
@cgdecker Can you please suggest me an issue that i should work on next. since i am new to contributing to google guava, i don't know which issue would be the best to work on for beginners like me?

cgdecker:
@sagar-shah94 Your changes in this PR are being synced out in #3130. Please reply there to confirm you're ok with that to make our CLA bot happy.

PR #3129, opened 2018-04-26
collect: add Sets#newSingleElementSet

kofemann:
In many situations APIs force to use Set, where only a single element is
required. Such use cases can be covered by Sets#newHashSet(element).
However, if we know in advance that this is an immutable Set with a
single element a more efficient implementation can be used.

Add a new Sets#newSingleElementSet method to create a lightweight Set
to store a single element.


lowasser:
I'm not sure I follow.  What is wrong with `ImmutableSet.of(element)` or, if you need nullability, the JDK `java.util.Collections.singleton(element)`?

jbduncan:
What does this bring to the table that `ImmutableSet.of(element)` and `Collections.singleton(element)` doesn't?

jbduncan:
@lowasser got there before me. :)

kofemann:
Crap! Some how I have missed it in JDK and was surprized that something like this does not exists in guava. Now I know why :). Sorry for the noise!

kofemann:
@jbduncan every thing is ok.

PR #3134, opened 2018-05-04
Moe Sync

ronshapiro:

Commits:
=====
<p> Bump jsr305 version.

31047249de250f2835640e02485f47a0b8bb4e84

### DIFF
--- a/android/pom.xml
+++ b/android/pom.xml
@@ -221,7 +221,7 @@
       <dependency>
         <groupId>com.google.code.findbugs</groupId>
         <artifactId>jsr305</artifactId>
-        <version>1.3.9</version>
+        <version>3.0.2</version>
       </dependency>
       <dependency>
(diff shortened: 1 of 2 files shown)


jbduncan:
Oh! I admit that I'm surprised that we're still depending on JSR-305 annotations, given that the Checker Framework's annotations were introduced as a dependency at some point.

Whereabouts in Guava are the JSR-305 annotations still being used?

ronshapiro:
`ParamatersAreNonNullByDefault`. It doesn't seem like https://youtrack.jetbrains.com/issue/KT-21408 has gone anywhere, but that's the remaining holdout.

ronshapiro:
`ParamatersAreNonNullByDefault`. It doesn't seem like https://youtrack.jetbrains.com/issue/KT-21408 has gone anywhere, but that's the remaining holdout.

Issue #2824, opened 2017-05-23
Unnecessary-looking dependencies


OrangeDog:
New in Release 22 are a bunch of Maven dependencies that look like they should be compile-only, but are now required at runtime.
```
[INFO] +- com.google.guava:guava:jar:22.0:compile
[INFO] |  +- com.google.code.findbugs:jsr305:jar:1.3.9:compile
[INFO] |  +- com.google.errorprone:error_prone_annotations:jar:2.0.18:compile
[INFO] |  +- com.google.j2objc:j2objc-annotations:jar:1.1:compile
[INFO] |  \- org.codehaus.mojo:animal-sniffer-annotations:jar:1.14:compile
```

cpovirk:
The goal here was to fix https://github.com/google/guava/issues/2721. Evidently "compile-only" doesn't actually make the annotations available at compile time to projects that compile against Guava, and that causes problems for people who compile with all warnings as errors. But if making the dependencies required at runtime causes problems, too, we'll have to figure out what to do. Are they causing problems for you, or is it just surprising that we made this change?

OrangeDog:
In case you didn't know "compile-only" is the `provided` scope. The `compile` scope is compile-time and later.

Sounds like they should be either `<optional>true</optional>`, `<scope>provided</scope>`, or moved into a new "build tools" pom that people can depend on separately.

cpovirk:
They used to be `<optional>true</optional>`, and that's what caused #2721. It's possible that `<scope>provided</scope>` would fare better; I don't know.

OrangeDog:
I believe `<scope>provided</scope>` matches your intent, so is probably the correct solution.

efenderbosch:
Any chance of a quick 22.1 release to resolve this? Otherwise we'll be adding a ton of exclusions for dependency convergence.

t-tsutsumi:
I believe #2721 (0e29934933f33379de953727171f7ca1ec616a58) is a reasonable change, but since it affects backward compatibility, I think it should be clarified at least in the release notes. Please be aware that changes to pom.xml do not appear in the Full JDiff Report.

cpovirk:
I just tested with `<scope>provided</scope>`, and it reintroduces #2721 :(

I'll put something in the release notes. We should think more about whether dependency version bumps should show up in the release notes in general.

cpovirk:
https://github.com/google/guava/wiki/Release22#additional-changes

[...]

Issue #2721, opened 2017-01-20
Many warning messages related to errorprone "CompatibleWith"


enwired:
When compiling code that depends on Guava 21.0, [Edit: I mistakenly said 20.0 originally] using Maven, I get many warning messages like this:

> Cannot find annotation method 'value()' in type 'com.google.errorprone.annotations.CompatibleWith': class file for com.google.errorprone.annotations.CompatibleWith not found

I do not use the errorprone library in my code, so this error message is due to its use by Guava.

Obviously, these are only warning messages and are not critical.

cpovirk:
The reason is that we don't make the Error Prone annotations available to our users at compile time. We could maybe change that, but our understanding has been that it shouldn't be necessary. Do you know if you're using any Maven plugins that might be plugging into the compiler (annotation processors, mainly)?

enwired:
These warnings are happening even in some pretty simple packages where there are no exotic maven plugins.

It does not happen on all of my modules that use Guava.  But I haven't narrowed-down which Guava class is involved.

The warnings happen during the "default-compile" step from maven-compiler-plugin.  I've tried version 3.3 and 3.6 of that plugin.

The warnings appear even if I set `<showWarnings>false</showWarnings>`

enwired:
Ok.  So this is related to annotation processors of some basic sort.  My standard build involves this:

```xml
<build>
  <plugins>
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-compiler-plugin</artifactId>
        <version>3.3</version>
        <configuration>
          <source>1.8</source>
          <target>1.8</target>
          <compilerArgs>
            <arg>-Xlint:all</arg>
            <arg>-Xlint:-serial</arg>
          </compilerArgs>
          <showWarnings>true</showWarnings>
          <showDeprecation>true</showDeprecation>
        </configuration>
      </plugin>
  </plugins>
</build>
```

The warnings are shown only when I include `Xlint` along with either of `showWarnings` or `showDeprecation`.

Then 9 identical warnings appear if my java code contains one single line like this:
`ImmutableTable.Builder<?,?,?> builder = ImmutableTable.builder();`

So, I don't know if there is a bug in maven compiler plugin, or in Guava's `ImmutableTable` or what.  But I really shouldn't be seeing these warnings.

Again, not a bug in the program operation, just an annoying set of warning messages.

stevenschlansker:
[...]

PR #3154, opened 2018-05-23
Use api instead of compile for Android

jereksel:


JakeWharton:
The direct analogous is `api`. `implementation` is merely an optimization that should be opted into.

jereksel:
I signed it!

jbduncan:
@JakeWharton I thought it was the other way round? Isn't `implementation` the same as `compile`, with `api` being the optimisation? :/

jbduncan:
@JakeWharton Wait, I'm challenging myself now. I realise that I'm not 100% clear on the difference between `api` and `implementation`, so never mind. :)

JakeWharton:
No. Compile transitively exposes the dependency on the classpath of
downstream modules which is the same as api. Implementation hides the
dependency. Of course Gradle doesn't enforce or validate any of this so
it's a giant footgun unless you know what you're doing.

jereksel:
`api` it is then.

cpovirk:
(Found https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation to back up Jake.)

cpovirk:
Hmm, so should we change the JRE one to `api`, as well? It sounds like `compile` may be deprecated.

JakeWharton:
Android projects use neither. API and implementation are available for all
Android plugins for about a year now. The reason these PRs are sweeping the
web is that they now wan when you use compile.

cpovirk:
Ah! Thanks.

jbduncan:
Hmm. It sounds to me that `implementation` is the most sensible default, and `api` should be used only if the `java-library` plugin is imported _and_ the given project wants to expose Guava types to its users. Is my understanding correct @JakeWharton?

JakeWharton:
Your understanding is correct because you understand the implications.
Anyone who needs to look at the README to figure out how to add Guava
beyond maybe just copy/pasting the coordinates will not have that
understanding.

jbduncan:
@JakeWharton Cool! So you're suggesting `api` here because it's the most liberal option and most equivalent to the older `compile`, am I right?

cpovirk:
(Even for people who don't expose Guava in their APIs, I'd be a little nervous about excluding transitive dependencies from the classpath, given problems with static analyzers and perhaps annotation processors discussed in #2824 and #2721. (There could also be problems if people use Guava annotations but don't consider that "exposing an API," but that seems less likely.) I suppose that those problems are relatively unlikely to affect people, though.)

cpovirk:
Anyway, I do like the idea of suggesting the thing that is more likely to work (even if it's slower in some cases -- and I suspect only marginally so, since Guava isn't going to be changing between one build and the next). It is a little sad that this exports Guava to downstream targets, but that seems like the lesser evil (and no worse than what people see today).

One final thought: We _could_ decide to give instructions for each of...

- Java 8+
- Java 7
- Android

...where the Java 7 instructions look like the _current_ Android instructions and the Android instructions look like your new ones. But I'm not sure that's worth the potential confusion.

jbduncan:
@cpovirk I think it's worth at least changing the remaining `compile` instruction for "guava-jre" to `api`, so that the Gradle installation instructions are consistent. :)

cpovirk:
@jbduncan, @JakeWharton said above:

> I believe 'api' only exists for JRE projects if you're using the
> 'java-library' plugin. If you're using the regular 'java' plugin you only
> get 'compile' and 'implementation' (the idea being you're a leaf node and
> don't have anywhere to expose dependencies to).

So I worry that `api` won't exist for some people, and they'll be confused.

Does Gradle warn about `compile` in JRE projects like it does in Android projects?

Issue #3086, opened 2018-03-26
FinalizableReferenceQueue might keep ClassLoaders around


netdpb:
Based on #92, in 2012 we made sure that using `FinalizableReferenceQueue` didn't stop `ClassLoader`s from being destroyed once they were no longer used.

Testing this behavior in JDK 9 revealed some strange failures. `FinalizableReferenceQueueClassLoaderUnloadingTest` has three tests:

1. `testUnloadableWithoutSecurityManager`
2. `testUnloadableWithSecurityManager`
3. `testUnloadableInStaticFieldIfClosed`

(3) fails in JDK 9, but the others pass. During testing, we tried ignoring (3) in JDK 9. Then, (2) fails in JDK 9 but (1) passes. If we ignore (2) and (3) in JDK 9, then (1) fails.

Tracking down stray references to `ClassLoader`s for this behavior is difficult, and this heisenbug-like behavior will make it even more difficult.

Since Guava v10, `MapMaker` no longer uses `FinalizableReferenceQueue`, and we believe it to be infrequently used in general. Because of that, we're going to ignore all of these tests in JDK 9 and treat this as a known, but low-priority, bug.

Please comment here if you are impacted by this bug.

Please also feel free to work on tracking down the underlying issue. If you fix it, we would welcome a PR.

YuryYaroshevich:
After running `mvn clean package` inside of guava-tests dir I didn't see failed tests from `FinalizableReferenceQueueClassLoaderUnloadingTest ` but instead I saw a bunch of failing tests from `MoreFilesFileTraverserTest`.  
My machine setup:
```
administrators-MacBook-Pro-3:guava yyar$ java -version
java version "9.0.4"
Java(TM) SE Runtime Environment (build 9.0.4+11)
Java HotSpot(TM) 64-Bit Server VM (build 9.0.4+11, mixed mode)
```

jrtom:
@YuryYaroshevich please clarify (a) what version of Guava you're using and (b) what test failures you're seeing; this might be a completely separate issue.

YuryYaroshevich:
@jrtom ,
I forked guava repo and then I did:
1) `cd guava-tests`
2) `mvn clean test`
and I got these results:
```
Tests in error: 
[...]

PR #3186, opened 2018-06-24
add 'oraclejdk10' to Travis build matrix

sullis:


cpovirk:
Well, we should get the tests working with Java 10 so that the continuous build at least verifies that nothing new breaks under Java 8 :( Looks like we should update Mockito, for starters. We'll see how things look and how soon we can get it all working. Additional help is welcome. Thanks for raising this.

cpovirk:
Looks like only 2 failures after that:

```
Failed tests: 
  testLazyStackTraceWorksInProd(com.google.common.base.ThrowablesTest)

Tests in error: 
  testUnloadableInStaticFieldIfClosed(com.google.common.base.FinalizableReferenceQueueClassLoaderUnloadingTest): Predicate did not become true within 10 second timeout
```

The former is just checking that `Throwables.lazyStackTrace` is able to use its fast path, which may just become impossible with new versions of Java. It will still return the stack trace, albeit by using the slow path. Users should consider Java 9's `StackWalker`. 

The latter is https://github.com/google/guava/issues/3086, which we hope affects few users.

In both cases, I'll update the test to not run on Java 10, just as we did for Java 9.

Issue #3210, opened 2018-07-19
Quantiles.ScaleAndIndexes#computeInPlace throws an unexpected ArrayIndexOutOfBoundsException if the array indexes is empty


NhatDinh:
In `Quantiles.ScaleAndIndexes#computeInPlace`, if the array `indexes` of `Quantiles.ScaleAndIndexes` is empty, an unexpected `ArrayIndexOutOfBoundsException` is thrown.
The problem is if the last argument(`int[] indexes`) of `Quantiles.ScaleAndIndexes` constructor is length of 0, `computeInPlace` passes an invalid negative value as the third argument to the private method `selectAllInPlace` and causes the error. 

To reproduce, run this code:
```java
import com.google.common.math.Quantiles;
import java.util.Map;

public class EmptyIndexes {
   public static void main(String args[]) {
       int[] empty = {};
       int scale = 10;
       int[] dataset= {0, 1, 2, 3, 4, 5};

       Quantiles.ScaleAndIndexes sai = Quantiles.scale(scale).indexes(empty);
       Map<Integer, Double> map = sai.computeInPlace(datatset_double);
   }
}
```

The program throws `ArrayIndexOutOfBoundsException`.
This is an unfriendly error that can be easily fixed by using `checkArgument()` in constructor of `Quantiles.ScaleAndIndexes`.

NhatDinh:
Link to proposed fix: [PR #3211](https://github.com/google/guava/pull/3211)

PeteGillin:
Thanks for this, @NhatDinh , I guess we missed this error case.

PR #3211, opened 2018-07-19
Check the array indexes in Quantiles.ScaleAndIndexes is non empty

NhatDinh:
Fix [Issue #3210 ](https://github.com/google/guava/issues/3210)


NhatDinh:
I signed it!

Issue #3209, opened 2018-07-19
 Missing Performance Optimization for toArray of size zero


thespags:
After reading [“Arrays of Ancient Wisdom”](https://shipilev.net/blog/2016/arrays-wisdom-ancients/#_experimental_setup) I was curious as to `ImmutableList`'s and `ImmutableSet`'s performance. It turns out the zero case, `listOfStrings.toArray(new String[0])`, is not optimized for array backed collections as `ImmutableCollection`'s will create a new array of appropriate size so we don't get the use of `Arrays.copyOf(src, size, Foo[].class)` mentioned in the article. I ran the same benchmark tests as the post but with `ImmutableSet` and `ImmutableList`. Interestingly `ImmutableSet` is way more performant than `HashSet`'s implementation. Locally, I've added overriding `RegularImmutableList` to use `ArrayList`'s implementation of `toArray` to see the performance gain.

```
ToArrayBenchmark.simple 0 arraylist avgt 15 31.465 ± 4.786 ns/op
ToArrayBenchmark.simple 0 hashset avgt 15 32.851 ± 1.709 ns/op
ToArrayBenchmark.simple 0 immutableset avgt 15 5.016 ± 0.686 ns/op
ToArrayBenchmark.simple 0 immutablelist avgt 15 4.508 ± 1.252 ns/op
ToArrayBenchmark.simple 1 arraylist avgt 15 11.687 ± 0.747 ns/op
ToArrayBenchmark.simple 1 hashset avgt 15 33.714 ± 4.696 ns/op
ToArrayBenchmark.simple 1 immutableset avgt 15 6.244 ± 0.711 ns/op
ToArrayBenchmark.simple 1 immutablelist avgt 15 6.495 ± 0.665 ns/op
ToArrayBenchmark.simple 10 arraylist avgt 15 12.733 ± 0.575 ns/op
ToArrayBenchmark.simple 10 hashset avgt 15 54.044 ± 0.771 ns/op
ToArrayBenchmark.simple 10 immutableset avgt 15 26.258 ± 0.349 ns/op
ToArrayBenchmark.simple 10 immutablelist avgt 15 26.280 ± 0.365 ns/op
ToArrayBenchmark.simple 100 arraylist avgt 15 56.674 ± 0.894 ns/op
ToArrayBenchmark.simple 100 hashset avgt 15 410.268 ± 8.487 ns/op
ToArrayBenchmark.simple 100 immutableset avgt 15 66.360 ± 3.243 ns/op
[...]

PR #3213, opened 2018-07-20
Missing Performance Optimization for toArray of size zero #3209

thespags:
Proof of concept for improving RegularImmutableList's toArray performance.

### DIFF
--- a/guava/src/com/google/common/collect/ImmutableCollection.java
+++ b/guava/src/com/google/common/collect/ImmutableCollection.java
@@ -194,7 +194,7 @@ public abstract class ImmutableCollection<E> extends AbstractCollection<E> imple
 
   @CanIgnoreReturnValue
   @Override
-  public final <T> T[] toArray(T[] other) {
+  public <T> T[] toArray(T[] other) {
     checkNotNull(other);
     int size = size();
     if (other.length < size) {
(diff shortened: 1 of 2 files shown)


thespags:
I signed it!

nglorioso:
https://github.com/google/guava/commit/35ca2ccdb761de615df2f32b30ef497595f93a68#diff-68927d2f7d9093f2ba963bdf10537151

created the existing implementation. I'm not sure that this implementation would be much faster, considering that RegularImmutableList uses the `System.arraycopy` in its implementation of `copyIntoArray` (which both parts of the new implementation you propose delegate down to).

Sorry for the delay in getting to reviewing this PR!

thespags:
@nglorioso my idea was a proof of concept as requested by @lowasser. This implementation **is** faster. I've shown it on the original issue. Check out #3209 with my results. @lowasser expanded the the optimization with https://github.com/google/guava/commit/f5a3541fd118d370488b9f9c15facd408e180b17

thespags:
@nglorioso Further, if you read #3209, you'll see the issue was ImmutableCollection always creates an array of correct size before calling `System.arraycopy` ignoring the optimization of passing a 0 length to `Array.copyOf`. A more in depth analysis is in the orginal blog post I reference, “Arrays of Ancient Wisdom” .

PR #3217, opened 2018-07-30
Change toString() redundant use of StringBuilder since Java 1.6 >

trentkjackson:
From Java 1.6 > the plus operator is changed into StringBuilder by compiler.

### DIFF
--- a/guava/src/com/google/common/net/HostAndPort.java
+++ b/guava/src/com/google/common/net/HostAndPort.java
@@ -297,16 +297,13 @@ public final class HostAndPort implements Serializable {
   @Override
   public String toString() {
     // "[]:12345" requires 8 extra bytes.
-    StringBuilder builder = new StringBuilder(host.length() + 8);
-    if (host.indexOf(':') >= 0) {
-      builder.append('[').append(host).append(']');
-    } else {
-      builder.append(host);
-    }
+    String str = "";
+    boolean hostHasSemicolon = host.indexOf(":") >= 0;
+    str += hostHasSemicolon ? String.format("[%s]", host) : host;
     if (hasPort()) {
-      builder.append(':').append(port);
+      str += String.format(":%s", port);
     }
-    return builder.toString();
+    return str;
   }
 
   /** Return true for valid port numbers. */


guava/src/com/google/common/net/HostAndPort.java
```diff
@@ -297,16 +297,13 @@ public int hashCode() {
...
-      builder.append('[').append(host).append(']');
-    } else {
-      builder.append(host);
-    }
+    String str = "";
+    boolean hostHasSemicolon = host.indexOf(":") >= 0;
+    str += hostHasSemicolon ? String.format("[%s]", host) : host;
     if (hasPort()) {
-      builder.append(':').append(port);
+      str += String.format(":%s", port);
```
MarkMielke: Have you timed the code to ensure it is actually faster? The code looks more complex than before, and slower due to the excess complexity of using `String.format()` and ternary operator. It's not clear to me what is wrong with the previous code, which is very straight forward.

MarkMielke: You piqued my curiosity. According to javap:

1. The original version disassembles toString() to 86 byte codes.
2. The proposed change disassembles toString() to 108 byte codes.

In terms of actual runtime using Oracle Java 8u162: (A = original, B = proposed)

- Test A completed in 418 ms, with length 180000000
- Test B completed in 86 ms, with length 180000000
- Test A + port completed in 574 ms, with length 230000000
- Test B + port completed in 4891 ms, with length 230000000
- Test A + ipv6 + port completed in 579 ms, with length 120000000
- Test B + ipv6 + port completed in 8890 ms, with length 120000000

This shows that your code is much faster in the "hostname only" scenario, but much slower once you add the ipv6 processing, or the port.

Taking a look at your code, it seems clear to me why this would be. In the "fast path", you are returning the hostname without any processing. You are only invoking `String.format()` or `StringBuilder` in the other cases.

This code has the same benefit as yours, but has almost none of the overhead in the slower cases:

```
    public String toString() {
        boolean hasIpv6 = host.indexOf(":") >= 0;
        boolean hasPort = hasPort();

        if (!hasIpv6 && !hasPort) {
            /* This is the fast path. No string building is required. */
            return host;
        }

        StringBuilder builder = new StringBuilder(host.length() + 8);
        if (hasIpv6) {
            builder.append('[').append(host).append(']');
        } else {
            builder.append(host);
        }
        if (hasPort) {
            builder.append(':').append(port);
        }
        return builder.toString();
    }
```

Benchmark results: (A = original, B = your proposed, C = above proposed)

- Test A completed in 418 ms, with length 180000000
- Test B completed in 86 ms, with length 180000000
- Test C completed in 105 ms, with length 180000000
- Test A + port completed in 574 ms, with length 230000000
- Test B + port completed in 4891 ms, with length 230000000
- Test C + port completed in 588 ms, with length 230000000
- Test A + ipv6 + port completed in 579 ms, with length 120000000
- Test B + ipv6 + port completed in 8890 ms, with length 120000000
- Test C + ipv6 + port completed in 572 ms, with length 120000000

+/- measurement error, the above code is the best choice of the three options.

But, for the benchmarks, I am executing the loop 10 million times. Of the cases above, only the first "hostname only" case sees any speed up, and for that one, it takes 10 million iterations to see 300 ms of speed up.

I don't think 300 ms per 10 million iterations is worth making the code more complex, and introducing a new code path which requires additional test cases to confirm that every code path works correctly every time.

I think the current code as-is, works just fine and there is no need for a change here.

ogregoire:
You take the assumption that

> From Java 1.6 > the plus operator is changed into StringBuilder by compiler.

However, [the compiler uses invokedynamic instructions since Java 9](https://www.guardsquare.com/en/blog/string-concatenation-java-9-untangling-invokedynamic) (OK, Guava is not yet compiled against Java 9+, but what about the future? What about future Java versions using something different than invokedynamic?)

Maaartinus:
@ogregoire IMHO this all doesn't matter here. The plus operator gets changed to something, but this only for a single expression. Maybe even for a sequence like `String s = ...; s += "a"; s += 5; ...`, but not for an arbitrarily complex code containing conditionals like here. With Java 8, there are multiple `StringBuilder`s there...

ogregoire:
@Maaartinus What I meant is that something that is true today might become false tomorrow. His basic assumption that `+` is translated to `StringBuilder`'s `append` from Java 6 onwards is proven as false with Java 9's indy concat.

lowasser:
I'm pretty sure that's the intent of http://openjdk.java.net/jeps/280...

lowasser:
I'm with @MarkMielke: the current code as-is works just fine and there is no need for a change here.

PR #3222, opened 2018-08-02
Joiner returned from skipNulls() now returns this from an override of skipNulls()

rgoldberg:
`Joiner` returned from `skipNulls()` now returns `this` from an override of `skipNulls()`.

This prevents creating additional anonymous inner class `Joiner`s on subsequent calls to `skipNulls()`.

Such calls are probably rare, but, given that each new anonymous inner class `Joiner` instance will keep a reference to its parent, it's best to minimize the number of these instantiations in any ways possible.

### DIFF
--- a/android/guava/src/com/google/common/base/Joiner.java
+++ b/android/guava/src/com/google/common/base/Joiner.java
@@ -272,6 +272,11 @@ public class Joiner {
       }
 
       @Override
+      public Joiner skipNulls() {
+        return this;
+      }
+
+      @Override
       public MapJoiner withKeyValueSeparator(String kvs) {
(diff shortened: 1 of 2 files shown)


cpovirk:
I suspect that few users call `skipNulls()` twice. Certainly this would help them somewhat, which is a small help, but there's also the small downside of extra code, in particular on Android.

cpovirk:
> I suspect that few users call skipNulls() twice. Certainly this would help them somewhat, which is a small help, but there's also the small downside of extra code, in particular on Android.

I think that consideration probably wins out. Sorry for taking 5 years to make that call.

PR #3225, opened 2018-08-03
Change ImmutableSet.Builder to keep overflow hashtables and only deduplicate within table to improve performance

### LINKED ISSUE #3223: Large Regression in Performance of ImmutableSet Building.
https://github.com/google/guava/commit/a98632ee83a6c4c93f4b31820437f68ddde041dd#diff-439cf531381acbc7a3a95b99000d22ca
Introduced a significant performance regression.

The regression caused certain commands of our project to regression 50%.

I performed JMH benchmarks and got the following results (Note: small set is 3 distinct elements, large set is 1000000 insertions, 500000 distinct elements, elements are Integers, building is `add()` and `build()` operations)

testname | Without Mentioned Changed: | With Mentioned Changes 
----------|-------------------------------|--------------------------
addAndBuildSmallSet | 11841379.432 ± 605281.339 ops/s | 11402608.445 ± 145690.757 ops/s 
addAndBuildLargeSet | 16.375 ± 4.594  ops/s |  11.539  ± 1.504  ops/s

This is almost a 100% regression in small set building and 50% regression in large set building.

bobyangyf:
This fixes #3223 

Building large sets in ImmutableSet.RegularSetBuilderImpl regressed with change mentioned in the issue, with a slightly gain over the original performance.

Instead of deduplicating all elements and rehashing when table needs to expand, we keep a list of overflow tables, and only deduplicate within each hashtable. This saves work from consistently rehashing elements while still allowing deduplication of elements as they come.

JMH benchmarks results (Note: small set is 3 distinct elements, large set is 1000000 insertions, 500000 distinct elements, elements are Integers, building is `add()` and `build()` operations)

testname | Without Regression: | With Regression | With this change
----------|---------------------|------------------|------------------
addAndBuildSmallSet | 12315677.054 ±  690237.444  ops/s | 14177773.234 ± 1175789.157  ops/s  ops/s | 12664284.533 ± 1089245.357  ops/s
addAndBuildLargeSet | 17.900 ± 1.522  ops/s |  11.184 ± 1.794  ops/s | 19.312 ± 2.329  ops/s

https://gist.github.com/bobyangyf/05189f515d37b6c3a471ff1659024e12

### DIFF
--- a/guava/src/com/google/common/collect/ImmutableSet.java
+++ b/guava/src/com/google/common/collect/ImmutableSet.java
@@ -31,6 +31,7 @@ import com.google.j2objc.annotations.RetainedWith;
 import java.io.Serializable;
 import java.math.RoundingMode;
 import java.util.Arrays;
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
@@ -571,7 +572,7 @@ public abstract class ImmutableSet<E> extends ImmutableCollection<E> implements 
     abstract SetBuilderImpl<E> add(E e);
 
     /** Adds all the elements from the specified SetBuilderImpl to this SetBuilderImpl. */
-    final SetBuilderImpl<E> combine(SetBuilderImpl<E> other) {
+    SetBuilderImpl<E> combine(SetBuilderImpl<E> other) {
       SetBuilderImpl<E> result = this;
       for (int i = 0; i < other.distinct; i++) {
         result = result.add(other.dedupedElements[i]);
@@ -698,9 +699,19 @@ public abstract class ImmutableSet<E> extends ImmutableCollection<E> implements 
   }
 
   /**
-   * Default implementation of the guts of ImmutableSet.Builder, creating an open-addressed hash
-   * table and deduplicating elements as they come, so it only allocates O(max(distinct,
-   * expectedCapacity)) rather than O(calls to add).
+   * Default implementation of the guts of ImmutableSet.Builder, creating many open-addressed hash
+   * tables and deduplicating elements within each table as they come.
+   *
+   * <p>We do not deduplicate all elements as they come as doing so requires a lot of hash table
+   * resizing if many elements are being added. Instead, we will deduplicate and hash all elements
+   * until the table needs resizing. Then we will create a new table without copying the previous
+   * table, and hash and deduplicate only newly added elements until the new table is full, and then
+   * merging the deduplicated tables together at the end. This way, adding n elements, m of which
+   * are distinct, saves about O(m/2) rehashes of elements compared to deduplicating all elements
+   * and resizing a single hashtable.
+   *
+   * <p>while worst case, we allocate O(n) during building, given uniformly distributed duplicates
+   * during the build stage, the average memory allocated is still O(m)
    *
    * <p>This implementation attempts to detect hash flooding, and if it's identified, falls back to
    * JdkBackedSetBuilderImpl.
@@ -710,6 +721,12 @@ public abstract class ImmutableSet<E> extends ImmutableCollection<E> implements 
     private int maxRunBeforeFallback;
     private int expandTableThreshold;
     private int hashCode;
+
+    // this is number of elements in `hashTable`, which is the currently active hashTable that has
+    // yet to overflow.
+    // We use this to track when we should resize.
+    private int numElementsInHashTable = 0;
+    private ArrayList<Object[]> overflowTables;
 
     RegularSetBuilderImpl(int expectedCapacity) {
       super(expectedCapacity);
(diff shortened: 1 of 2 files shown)


bobyangyf:
I signed it!

guava/src/com/google/common/collect/ImmutableSet.java
```diff
@@ -722,20 +739,92 @@ static int maxRunBeforeFallback(int tableSize) {
...
 
+    @SuppressWarnings("unchecked")
+    void flattenOverflowTables() {
+      if (overflowTables == null) {
+        return;
+      }
+
+      // ensure capacity first to avoid resizing and inserting into a high load table as much as
+      // possible
+      ensureTableCapacity(hashTable.length * 2);
```
bobyangyf: I believe this can actually be calculated slightly differently for better performance.
Investigating...

guava/src/com/google/common/collect/ImmutableSet.java
```diff
@@ -698,9 +699,19 @@ static int maxRunBeforeFallback(int tableSize) {
...
+   *
+   * <p>We do not deduplicate all elements as they come as doing so requires a lot of hash table
+   * resizing if many elements are being added. Instead, we will deduplicate and hash all elements
+   * until the table needs resizing. Then we will create a new table without copying the previous
+   * table, and hash and deduplicate only newly added elements until the new table is full, and then
+   * merging the deduplicated tables together at the end. This way, adding n elements, m of which
+   * are distinct, saves about O(m/2) rehashes of elements compared to deduplicating all elements
+   * and resizing a single hashtable.
+   *
+   * <p>while worst case, we allocate O(n) during building, given uniformly distributed duplicates
```
lowasser: I believe you're underselling this, actually: I think the allocation is still O(n) worst case, not O(m).

AFAICT, the worst-case input for this implementation is e.g.

1, 1, 2, 1, 2, 3, 4, 1, 2, 3, 4, 5, 6, 7, 8, ....

in which case overflowTables contains

{1}, {1, 2}, {1, 2, 3, 4}, {1, 2, 3, 4, 5, 6, 7, 8}

which has "maximum duplication": every element that can be duplicated is duplicated as many times as possible.  But we've still only allocated O(m): m = n/2.

guava/src/com/google/common/collect/ImmutableSet.java
```diff
@@ -711,6 +722,12 @@ static int maxRunBeforeFallback(int tableSize) {
     private int expandTableThreshold;
     private int hashCode;
 
+    // this is number of elements in `hashTable`, which is the currently active hashTable that has
+    // yet to overflow.
+    // We use this to track when we should resize.
+    private int numElementsInHashTable = 0;
+    private ArrayList<Object[]> overflowTables;
```
lowasser: Just use `List`, here and below

guava/src/com/google/common/collect/ImmutableSet.java
bobyangyf: m is the number of distinct elements, n is the number of puts.

The above cause you mentioned would still end up with O(n) total memory in the worst case, since we keep every one of those {1}, {1,2} .... tables, making total of O(n) space.
It's true that the largest/latest overflow table, and the final allocation for the set would only be O(m) = O(n/2) though.

guava/src/com/google/common/collect/ImmutableSet.java
```diff
@@ -722,20 +739,92 @@ static int maxRunBeforeFallback(int tableSize) {
...
+      if (other instanceof RegularSetBuilderImpl) {
+        RegularSetBuilderImpl<E> otherR = (RegularSetBuilderImpl<E>) other;
+        SetBuilderImpl<E> result = this;
+        for (int i = 0; i < other.distinct; i++) {
+          result = result.add(other.dedupedElements[i]);
+        }
+        if (otherR.overflowTables != null) {
+          for (int i = 0; i < otherR.overflowTables.size(); i++) {
+            Object[] tables = otherR.overflowTables.get(i);
+            for (int j = 0; j < tables.length && tables[j] != null; j++) {
```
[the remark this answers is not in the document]
bobyangyf: The overflow tables are arrays of deduped elements (per each of the hash sets), which are not necessarily filled full (and in fact most often not filled completely full).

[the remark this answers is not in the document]
lowasser: I think they should be, I guess?  Here's what I thought your implementation was going for, every time a reset occurs:

- dedupedElements is set to a new, empty array of size k (which is twice the old length of dedupedElements)
- hashTable is set to a new array of size chooseTableSize(k)
- each time an element is added, we add it to the *current* hash table, and if it's not a duplicate of something already in there, it gets added to dedupedElements
- we reset when dedupedElements is full, meaning the hash table is saturated to its intended capacity

If that's not what you were going for, can you explain what you meant instead?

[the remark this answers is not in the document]
bobyangyf: In your outlined case, because `chooseTableSize(k)` returns powers of 2s, we end up wasting some capacity of the hashTable.
Since rehashes are the most expensive, I chose to maximize the use of the hashTable instead.

That is on the reset:
* hashTable is set to a new array of size chooseTableSize(currentHashTableCap * 2)
* dedupedElements is set to new array of size `expandTableThreshold`.
* we reset when `expandTableThreshold` is met, that is table is saturated.

corollary is that `dedupedElements` is full, expect that the initial construction of the hashTable and `dedupedElements` don't follow these conditions. I could modify the constructor to follow these conditions as well, guaranteeing max fill.

However, the flattened result of `dedupedElements` is not max filled. It is sized according to the max possible size resulting from the flattening of all overflows, but there could be some duplicates between the overflows.
While not the case in this implementation, my next patch with some fixes makes it clear that when we copy builders, we can not guaranty that `dedupedElements` are filled.
Furthermore, the constructor of `RegularSetBuilderImpl` doesn't guarantee the sizes that result in always filled conditions either.

bobyangyf:
This contains some bugs not caught in tests. Will submit some patches later with more tests.

guava/src/com/google/common/collect/ImmutableSet.java
```diff
@@ -722,20 +739,92 @@ static int maxRunBeforeFallback(int tableSize) {
...
+      // ensure capacity first to avoid resizing and inserting into a high load table as much as
+      // possible
+      ensureTableCapacity(hashTable.length * 2);
+
+      ArrayList<Object[]> overflows = overflowTables;
+      overflowTables = null;
+      // now we rehash.
+      for (int i = 0; i < overflows.size(); i++) {
+        Object[] table = overflows.get(i);
+        for (int j = 0; j < table.length && table[j] != null; j++) {
```
lowasser: I'm concerned about the semantics of this, unfortunately: it seems like it'd destroy insertion order, because elements from overflowTables are getting added *after* elements from the current chunk, when they were added to the builder before.  I'm not sure I see a good way around that. =(

bobyangyf: Yeah this was a bug I found. I'm closing this pull request due to some new findings that I'm going to post on the issue.

bobyangyf:
See #3223 for details.

PR #3236, opened 2018-08-14
Use newGuard() to make example in Monitor class javadoc more concise.

jsampson-gw:
This commit also cleans up the method javadoc for newGuard() itself and replaces a lingering `throw Throwables.propagate(throwable)` with simply `throw throwable` (not related to newGuard(), but good to clean up).

_Note:_ I have not modified the copy of Monitor.java under the android directory because the newGuard() method was never added there.

### DIFF
--- a/guava/src/com/google/common/util/concurrent/Monitor.java
+++ b/guava/src/com/google/common/util/concurrent/Monitor.java
@@ -18,7 +18,6 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.annotations.Beta;
 import com.google.common.annotations.GwtIncompatible;
-import com.google.common.base.Throwables;
 import com.google.errorprone.annotations.concurrent.GuardedBy;
 import com.google.j2objc.annotations.Weak;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


jsampson-gw:
Regarding the CLA: I've signed an individual CLA and Guidewire has also signed a corporate CLA.

cpovirk:
> Guidewire has also signed a corporate CLA.

Do you know how long ago? They're not showing up in our system.

jsampson-gw:
Huh. I'll dig it up and follow up in email.

cpovirk:
Justin's company signed the CLA a while back, and it's just been migrated to our new system. We'll see if CLABot updates as a result of this post. But regardless, this PR is now compliant.

guava/src/com/google/common/util/concurrent/Monitor.java
```diff
@@ -1031,7 +1024,7 @@ private boolean isSatisfied(Guard guard) {
       return guard.isSatisfied();
     } catch (Throwable throwable) {
       signalAllWaiters();
-      throw Throwables.propagate(throwable);
+      throw throwable;
```
cpovirk: I had apparently missed/forgotten that this was possible [since Java 7](https://docs.oracle.com/javase/7/docs/technotes/guides/language/catch-multiple.html) -- nice.

jsampson-gw: Yeah. I'm sensitive to Throwables.propagate() because we actually had an exception-handling bug in Monitor's early days due to overuse of it, which was fixed with a more involved set of changes in 287bc67cac97052b13cbbc0358aed8054b14bd4a (long before Java 8 came out).

PR #3237, opened 2018-08-14
Optimize ImmutableSet.copyOf to use expectedSize and avoid redundant copying

### LINKED ISSUE #3223: Large Regression in Performance of ImmutableSet Building.

bobyangyf:
ImmutableSet.copyOf() was not taking advantage of expectedSize when copying other sets.
It was also creating redundant copies of arrays during copying that is later discarded.

This partially fixes #3223 in terms of this method.
Benchmarks below: These are adding 2000 distinct Objects with unique hashcodes that simply return the value of a field.

test name | with regression | without regression | with fix
-----------|------------|---------|--------
copySet |   12295.914 ±  264.252  ops/s | 32206.487 ±  882.676  ops/s | 35881.393 ± 2244.604  ops/s
copyList | 14.723 ± 6.335  ops/s | 4.547 ± 5.599  ops/s | 14.331 ± 4.146  ops/s
copyImmutableList | 13571.475 ± 1155.980  ops/s | 43166.369 ± 2471.499  ops/s | 43718.413 ± 1952.942  ops/s

### DIFF
--- a/guava/src/com/google/common/collect/ImmutableSet.java
+++ b/guava/src/com/google/common/collect/ImmutableSet.java
@@ -91,7 +91,7 @@ public abstract class ImmutableSet<E> extends ImmutableCollection<E> implements 
    * the first are ignored.
    */
   public static <E> ImmutableSet<E> of(E e1, E e2) {
-    return construct(2, e1, e2);
+    return construct(2, 2, Iterators.forArray(e1, e2));
   }
 
   /**
@@ -100,7 +100,7 @@ public abstract class ImmutableSet<E> extends ImmutableCollection<E> implements 
    * the first are ignored.
    */
   public static <E> ImmutableSet<E> of(E e1, E e2, E e3) {
-    return construct(3, e1, e2, e3);
+    return construct(3, 3, Iterators.forArray(e1, e2, e3));
   }
 
   /**
@@ -109,7 +109,7 @@ public abstract class ImmutableSet<E> extends ImmutableCollection<E> implements 
    * the first are ignored.
    */
   public static <E> ImmutableSet<E> of(E e1, E e2, E e3, E e4) {
-    return construct(4, e1, e2, e3, e4);
+    return construct(4, 4, Iterators.forArray(e1, e2, e3, e4));
   }
 
   /**
@@ -118,7 +118,7 @@ public abstract class ImmutableSet<E> extends ImmutableCollection<E> implements 
    * the first are ignored.
    */
   public static <E> ImmutableSet<E> of(E e1, E e2, E e3, E e4, E e5) {
-    return construct(5, e1, e2, e3, e4, e5);
+    return construct(5, 5, Iterators.forArray(e1, e2, e3, e4, e5));
   }
 
   /**
@@ -130,6 +130,7 @@ public abstract class ImmutableSet<E> extends ImmutableCollection<E> implements 
    *
    * @since 3.0 (source-compatible since 2.0)
    */
+  @SuppressWarnings("unchecked")
   @SafeVarargs // For Eclipse. For internal javac we have disabled this pointless type of warning.
   public static <E> ImmutableSet<E> of(E e1, E e2, E e3, E e4, E e5, E e6, E... others) {
     checkArgument(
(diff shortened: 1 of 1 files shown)


lowasser:
I'm not sure what you're going for here?  You mention "inserting into the set would make a copy," but you can't insert into the set; it's an immutable set.  Additionally, I'm not sure why you're switching to iterators?  Allocating an iterator seems unnecessary.

It totally makes sense that if we're copying a Set, we can use its size as a stronger-than-usual sizing hint, but I'm not sure about the other parts of this change.

guava/src/com/google/common/collect/ImmutableSet.java
```diff
@@ -193,23 +193,41 @@
...
      * TODO(lowasser): consider checking for Multiset here
      */
     // Don't refer to ImmutableSortedSet by name so it won't pull in all that code
     if (elements instanceof ImmutableSet && !(elements instanceof SortedSet)) {
-      @SuppressWarnings("unchecked") // all supported methods are covariant
       ImmutableSet<E> set = (ImmutableSet<E>) elements;
       if (!set.isPartialView()) {
         return set;
+      } else {
+        // We use the iterator to avoid making a copy of the set, since inserting into the set would
```
bobyangyf: I meant "builder" not "set". Oops. I'll update that.

guava/src/com/google/common/collect/ImmutableSet.java
```diff
@@ -193,23 +193,41 @@
...
+    } else if (elements instanceof Set) {
+      // we know that sets are all unique, so we can supply the set size to avoid resizing.
+
+      // In case of mutation, make a copy first for thread safety.
+      Object[] array = elements.toArray();
+      return construct(array.length, array.length, Iterators.forArray(array));
     }
+
+    // In case of mutation, make a copy first for thread safety.
     Object[] array = elements.toArray();
```
bobyangyf: Previously, even when we have an ImmutableCollection, we would create a copy of the collection as an array, and then insert them into the Builder, which would create an another array, which unnecessarily creates an array.
This made for a regression in the copyOf(ImmutableSet) benchmark.
Creating an iterator for the ImmutableCollection is much more efficient than copying the array.

I changed everything else to use iterators so that we can have one consistent api for `construct()`. And profiling didn't show that big a difference.
Otherwise, we need two `construct()` implementations where one takes in an `Iterable` and another takes in an `Array`.

bobyangyf:
ping?

guava/src/com/google/common/collect/ImmutableSet.java
```diff
@@ -144,37 +145,36 @@
...
    * <p>{@code elements} may contain only values of type {@code E}.
    *
    * @throws NullPointerException if any of the first {@code n} elements of {@code elements} is null
    */
-  private static <E> ImmutableSet<E> construct(int n, Object... elements) {
+  private static <E> ImmutableSet<E> construct(int n, Iterator<Object> elements) {
+    return construct(n, Builder.DEFAULT_INITIAL_CAPACITY, elements);
+  }
+
+  private static <E> ImmutableSet<E> construct(int n, int expectedSize, Iterator<Object> elements) {
```
lowasser: I'd prefer for this to be `Iterator<E>` and you have to cast the result of `toArray()` instead of the other way around.

guava/src/com/google/common/collect/ImmutableSet.java
```diff
@@ -91,7 +91,7 @@
    * the first are ignored.
    */
   public static <E> ImmutableSet<E> of(E e1, E e2) {
-    return construct(2, e1, e2);
+    return construct(2, Iterators.forArray(e1, e2));
```
lowasser: I think it makes sense to use the known number of elements as a sizing hint for the of(...) methods.

bobyangyf:
https://github.com/google/guava/commit/57e4b6d560874946d7142103386f7f17384e1708 seems to do this

PR #3261, opened 2018-09-12
Moe Sync

cpovirk:


cpovirk:
We have the author's permission in https://github.com/google/guava/pull/3236

Issue #2041, opened 2015-04-27
Add a random weighted selector


ogregoire:
I've often had to implement a random weighted selector in my various assignments (mostly for routing). As most programmers, I've turned to StackOverflow to find a solution to this issue. [Most of the solutions suggested by Google on SO](https://www.google.be/search?q=random+weighted+selection+java+site:stackoverflow.com) use algorithms with an init in O(n) and a get in O(n) or O(log n). The thing is that some nice solutions exist with O(n) for the init and O(1) for the get method, for instance. And those are mostly unknown, I think mostly about the "alias method" which return a value in O(1).

I've implemented [a generalized RandomSelector](https://github.com/ogregoire/fror-common/blob/master/src/main/java/be/fror/common/collection/RandomSelector.java) to solve the random selection issue with both a uniform and a weighted implementations.

If you feel that this is indeed something that the Guava users might need, feel free to copy it / inspire from it (CLA is signed).

aliakhtar:
I'm using this library in a project and have had pretty good results. Makes getting random weighted values really simple. Recommended.

PR #3266, opened 2018-09-16
Implement getRandom() for Iterables

dgray16:
Hi, I want you to present my minor improvement for **Iterables** class.

In commercial application I am working with I faced a problem that I need to take random element from **ArrayList**. Thing that confused me the most that such functionality has been rejected for _Apache Commons (Collections)_ long time ago and has not been reviewed.

I hope that _Guava_ community does not think the same way.
Thanks.

### DIFF
--- a/guava-tests/test/com/google/common/collect/IterablesTest.java
+++ b/guava-tests/test/com/google/common/collect/IterablesTest.java
@@ -1440,6 +1440,11 @@ public class IterablesTest extends TestCase {
         .testNulls();
   }
 
+  public void testGetRandom() {
+      ArrayList<String> strings = Lists.newArrayList("d", "a", "o");
+      strings.forEach(string -> assertTrue(strings.contains(Iterables.getRandom(strings))));
+  }
+
   private static void verifyMergeSorted(
       Iterable<Iterable<Integer>> iterables, Iterable<Integer> unsortedExpected) {
     Iterable<Integer> expected = Ordering.natural().sortedCopy(unsortedExpected);
--- a/guava/src/com/google/common/collect/Iterables.java
+++ b/guava/src/com/google/common/collect/Iterables.java
@@ -25,6 +25,7 @@ import com.google.common.annotations.GwtCompatible;
 import com.google.common.annotations.GwtIncompatible;
 import com.google.common.base.Function;
 import com.google.common.base.Optional;
+import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
 import com.google.common.base.Predicates;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
@@ -34,6 +35,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Queue;
+import java.util.Random;
 import java.util.RandomAccess;
 import java.util.Set;
 import java.util.Spliterator;
@@ -1015,6 +1017,18 @@ public final class Iterables {
           }
         };
     return new UnmodifiableIterable<>(iterable);
+  }
+
+  /**
+   * Picks up random value from {@link Iterable}.
+   * <p>
+   *     Consider providing {@link Iterable} with at least 1 element,
+   *     in other way you will get {@link IllegalArgumentException}.
+   * </p>
+   */
+  public static <T> T getRandom(final Iterable<T> iterable) {
+    Preconditions.checkNotNull(iterable);
+    return Iterables.get(iterable, new Random().nextInt(Iterables.size(iterable)));
   }
 
   // TODO(user): Is this the best place for this? Move to fluent functions?


dgray16:
I signed it!

jbduncan:
Hi @dgray16, welcome to Guava! :)

> Thing that confused me the most that such functionality has been rejected for Apache Commons (Collections) long time ago and has not been reviewed.

Huh, that's interesting. Could you point us towards a web page or something that confirms this, e.g. an entry in the Apache Commons mailing list?

> I hope that Guava community does not think the same way.

In a way, I hope so too! But I'm not a Googler, and even though a community of sorts has developed around Guava over the years, it's still the Guava team over at Google who have the ultimate say over what goes in and what doesn't. And the Guava team (1) have very high standards and (2) have historically not been the best at responding to issues or PRs quickly.

So to give a realistic idea of what to expect:
1. I'd say that you should expect to wait a few months or so before someone on the team does respond.
2. When they do respond, expect that they will want to follow their thorough internal review process first and [design the API themselves](https://github.com/google/guava/blob/master/CONTRIBUTING.md#api-changes) to meet their own standards, at which point they'll probably do the implementation as well since that's the easy part. :)

dgray16:
@jbduncan 
Hi there.
Appriciate your advices : )

Apache:
https://issues.apache.org/jira/browse/COLLECTIONS-541
https://issues.apache.org/jira/browse/COLLECTIONS-156

jbduncan:
You're welcome @dgray16! And thank you very much for providing the links. :+1:

jbduncan:
Ah okay, I see that https://issues.apache.org/jira/browse/COLLECTIONS-541 is still open, actually, but that [it's just not been at the top of anyone's priority list](https://issues.apache.org/jira/browse/COLLECTIONS-541?focusedCommentId=14606343&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14606343) at Apache Commons.

I see that PR https://github.com/apache/commons-collections/pull/5 was opened to implement that issue, but it was closed and I do not (yet?) understand why.

I see that [this Apache Commons JIRA comment](https://issues.apache.org/jira/browse/COLLECTIONS-541?focusedCommentId=14606343&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14606343) raises a concern over the method using a `new Random()` internally (which the code in this PR also suffers from). This is a problem because it means that one can't use an external `Random` with a deterministic "seed". Being able to set the "seed" for a `Random` is rather important in tests, since otherwise not the same sequence of pseudo-random numbers would get generated each time the `Random`'s methods are called in the tests, meaning that test results would be hard to reproduce.

There's also the problem that users may want to use pseudo-random number generators other than `Random`, e.g. `SecureRandom` or [Apache Commons RNG](https://commons.apache.org/proper/commons-rng/)'s implementations. Thus to accomodate these, `Iterables.getRandom(Iterable)` would really need to be something like `Iterable.getRandom(Iterable, IntSupplier)`.

But I'm sure that when the Guava team gets around to this, they'll cover these details if they think the idea behind this PR is a good idea. So no need to worry about them yet. :)

dgray16:
@jbduncan 
I think that approach with _IntSupplier_ is good one, but do not you think that in this way we are trying to solve the problem that does not exist?
Like, I am typical user of this API, and I want to get the result in the easiest way.

I will repeat myself once more that I am not against additional overloaded methods, I want to define motivation of potential API user.

jbduncan:
@dgray16 Good question! I've only just started working as a grad software engineer, so I don't have loads of experience yet, and for all I know, maybe going for a solution like using `IntSupplier` is indeed trying to solve a problem that doesn't exist!

All I can say is it's my gut feeling that to not include such an overload will be questioned by people eventually, especially if [this JEP](http://openjdk.java.net/jeps/8193209) is ever accepted, which aims to introduce more pseudo-random number generators in the Java standard library and provide common interfaces for them.

Guava team, WDYT?

JakeWharton:
You can do this with streams by skipping a random amount and invoking findFirst().

jrtom:
Here are some issues that I'd expect to come up in the API review.  (@jbduncan has already pointed out some of the key ones; thanks for that!)

First of all, this is what you're proposing as the implementation, which seems reasonable for what it's doing:

```
Iterables.get(iterable, new Random().nextInt(Iterables.size(iterable)))
```

Let's consider the [primary metric](https://github.com/google/guava/wiki/PhilosophyExplained#utility-times-ubiquity) by which Guava evaluates possible API additions.

Utility: your proposal would replace the above one-liner by a shorter one-liner that is (in my opinion) not difficult to do correctly, or to read.  So the utility--the amount of added value--is fairly low.

Ubiquity: generally we try to approximate that by looking for the number of times that the replaced code is used inside Google (since that's where we can easily search, and it's a pretty big code base).  

Doing a search on `Iterables.get\(.*new\sRandom` reveals that this pattern is currently used fewer than 10 times inside Google, so the 'ubiquity' aspect isn't looking great.

One thing that I think is a minus to this proposal: it hides the call to `Iterables.size()`.  This is fine if you're supplying a `Collection`, but if you're supplying an `Iterable` that isn't one, every time you call this method you re-incur the cost of calculating the size.

When combined with @jbduncan's good points--that you would want to be able to provide both the RNG and the seed, thus increasing the surface area of this API--this suggests to me that this proposal is unlikely to be accepted.

dgray16:
@JakeWharton 
Thanks for suggestion.

Do not you think that it would be more efficient to get element from array by index instead of iterating through list in case of _List_?
With _iterator_ we forced to iterate anyway.

dgray16:
@jrtom 
Thanks for reply.

1. Poor implemetation was made because of my confidence of a fact that such a "toy"method should not be too complex in its implementation.
2. Performance issue: this is kind of eternal debate : ) I think that hardware in 2018 is powerful enough to handle task to find size even for 10k elements in iterable.

Anyway it is just a discussion, my point is to present idea rather then argue about it.

ogregoire:
This looks like a duplicate of #2041 .

jrtom:
@ogregoire You could think of this as being a special case of #2041, but since selection using a uniform random distribution is such a common thing to want to do (that, importantly, requires considerably less specification than any non-uniform distribution) I think it's reasonable to treat them as mostly-separate problems, while recognizing that they're related and that a solution for one should be reasonably congruent to the other.

kashike:
A version of this that accepts a `Random` would be nice too.

kluever:
@jrtom I found 26 hits inside google using "Iterables\.get\(.*random", which is more than you, but still fairly low on the ubiquity scale.

I'm also generally -1 on this proposal (all of the above comments are good), but I'm happy to leave this open in the "research" category in case other folks want to make a stronger case for it.

In the case that you have a `List` (which I suspect is the majority case), we're not actually saving much:
`list.get(random.nextInt(list.size()); // before`
`Iterables.getRandom(list, random); // after`

If you have an `Iterable`/`Collection`, sure it's a little longer, but still not that much of a savings:
`Iterables.get(iterable, random.nextInt(Iterables.size(iterable))); // before`
`Iterables.getRandom(iterable, random); // after`

PR #3268, opened 2018-09-18
Update truth, animal sniffer, error prone annotations, and JUnit

swankjesse:
All of these dependencies are out of date from those used in gRPC, and preventing
me from getting a current build of Guava into gRPC.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,8 +18,8 @@
   <properties>
     <!-- Override this with -Dtest.include="**/SomeTest.java" on the CLI -->
     <test.include>%regex[.*.class]</test.include>
-    <truth.version>0.41</truth.version>
-    <animal.sniffer.version>1.14</animal.sniffer.version>
+    <truth.version>0.42</truth.version>
+    <animal.sniffer.version>1.17</animal.sniffer.version>
     <maven-javadoc-plugin.version>3.0.0</maven-javadoc-plugin.version>
   </properties>
   <issueManagement>
@@ -227,7 +227,7 @@
       <dependency>
         <groupId>com.google.errorprone</groupId>
         <artifactId>error_prone_annotations</artifactId>
-        <version>2.1.3</version>
+        <version>2.2.0</version>
       </dependency>
       <dependency>
         <groupId>com.google.j2objc</groupId>
@@ -237,7 +237,7 @@
       <dependency>
         <groupId>junit</groupId>
         <artifactId>junit</artifactId>
-        <version>4.11</version>
+        <version>4.12</version>
         <scope>test</scope>
       </dependency>
       <dependency>


cpovirk:
Thanks. I'm submitting this internally.

Do you know what problem it causes for gRPC? I would expect them to override dependency versions when necessary, since their deps' deps are likely to get out of date (like here...). We should still update, but I have been getting more involved in Maven/Gradle dependency-version issues lately, so I'd be interested for any info you have.

swankjesse:
When I tried to upgrade to Guava only in gRPC it failed because [they’re using](https://github.com/grpc/grpc-java/blob/master/build.gradle#L260) strict version matching:
```
resolutionStrategy.failOnVersionConflict()
```

So when I try to use Guava `26.0-jre` their build fails where gRPC and Guava dependencies disagree.
```
~/Projects/grpc-java master * ./gradlew build
*** Skipping the build of codegen and compilation of proto files because skipCodegen=true
> Task :grpc-protobuf-lite:extractIncludeProto FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Could not resolve all dependencies for configuration ':grpc-protobuf-lite:compile'.
> A conflict was found between the following modules:
    - com.google.errorprone:error_prone_annotations:2.2.0
    - com.google.errorprone:error_prone_annotations:2.1.3
  A conflict was found between the following modules:
    - org.codehaus.mojo:animal-sniffer-annotations:1.17
    - org.codehaus.mojo:animal-sniffer-annotations:1.14
  Run with:
      --scan or
      :grpc-protobuf-lite:dependencyInsight --configuration compile --dependency com.google.errorprone:error_prone_annotations
  to get more insight on how to solve the conflict.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 13s
16 actionable tasks: 2 executed, 14 up-to-date
```

I don’t think the strict versioning is worth the friction it causes, but gRPC isn’t my project. As long as Guava stays up-to-date on its own dependencies and releases frequently, it shouldn’t be a problem to get a gRPC release that can support it.

cpovirk:
Thanks. I think [gRPC agrees](https://github.com/grpc/grpc-java/commit/a3ff9cd784fce8a3a05054f0cd695954283e51f5) and is in turn doing it only for Spring. We've since [cleared up some misunderstands](https://github.com/spring-projects/spring-boot/issues/14275) we had about Spring, so gRPC might be willing to back that out. Anyway, we should still do these updates. I've submitted the change internally, and it should be mirrored out today.

cpovirk:
(@ejona86 so that I'm not talking about him behind his back :))

ejona86:
> So when I try to use Guava 26.0-jre their build fails where gRPC and Guava dependencies disagree.

We exclude mis-matched dependencies in our build. Yes, it is a pain, but it can be solved fully from the gRPC-side. But bumping the versions here isn't a bad thing.

I'm going to be _very_ happy to remove `failOnVersionConflict()`, but we need to replace it with another check. I've not found a way to check requireUpperBoundDeps from Gradle (for obvious reasons). I think we may be forced to have a "test Maven project" just to run requireUpperBoundDeps.

cpovirk:
> I've not found a way to check requireUpperBoundDeps from Gradle (for obvious reasons).

Oh, ha, of course. Thanks for the update.

PR #3269, opened 2018-09-19
A utility for single execution of  Runnable

tareqhs:
A funtionality similar to ``Suppliers.memoize()`` for Runnable

### DIFF
--- a/android/guava/src/com/google/common/util/concurrent/Runnables.java
+++ b/android/guava/src/com/google/common/util/concurrent/Runnables.java
@@ -14,6 +14,8 @@
 
 package com.google.common.util.concurrent;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import com.google.common.annotations.Beta;
 import com.google.common.annotations.GwtCompatible;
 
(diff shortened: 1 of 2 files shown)


tareqhs:
I signed it!

jolkdarr:
Hi @tareqhs 

Thanks for the reply.
For this kind of synchronization, I prefer to use an AtomicBoolean or a single short integer (0/1) avoiding the use of an object as lock.

tareqhs:
Hi @jolkdarr

This is a slightly different behavior. My intended behavior is that everyone depends on the outcome of the Runnable and waits for it to finish if it has not finished yet. The behavior you are describing is _triggering_ the task only once.  Maybe we should have a separate ``Runnables.triggerOnce(Runnable)`` for that?

ben-manes:
This seems similar in spirit to `RunnableFuture` in j.u.c.

netdpb:
What does this do that `java.util.concurrent.FutureTask` doesn't?

```
Runnable once = new FutureTask<>(runnable);
once.run(); // runs!
once.run(); // already ran; doesn't run!
```

PR #3270, opened 2018-09-19
Moe Sync

cpovirk:


cpovirk:
One change is by a Googler; the other has a verified CLA in https://github.com/google/guava/pull/3268.

PR #3273, opened 2018-09-23
Update the “No removed APIs since” counter to 26.0

swankjesse:
Removal of one of the catchingAsync overloads caused me some grief
because one of my projects’ transitive dependencies was using it.

https://github.com/google/guava/commit/87d87f5cac5a540d46a6382683722ead7b72d1b3
https://github.com/vitessio/vitess/pull/4214


cgdecker:
`Futures` itself is actually `@Beta`, so no non-`@Beta` API was removed. (Given how long it's been around, though, I do feel like it's something we should really prioritize getting as much as we can of out of `@Beta`.)

PR #3275, opened 2018-09-24
Check input collection size in PowerSet before actually copying it's content

tgyurci:
It helpes avoiding OOMs for big Sets.


kluever:
Thanks - I've submitted this internally and it should be mirrored out soon!

PR #3276, opened 2018-09-25
Add ofNull() method to Suppliers.

merlinosayimwen:
Despite the fact that I avoid the use of nullable references, there are often few alternatives, especially if third-party libraries are used.

If, in such cases, a supplier is needed that returns `null` continuously, the current solution with the Guava library is to call the `ofInstance()` method with the argument `null` . This is not really clean and can be easily improved.

My added method makes the user's code cleaner and more readable.

_Without my method:_
```java
Supplier<?> nullSupplier = Suppliers.ofInstance(null);
```

_With my method:_
```java
Supplier<?> nullSupplier = Suppliers.ofNull();
```

### DIFF
--- a/guava-tests/test/com/google/common/base/SuppliersTest.java
+++ b/guava-tests/test/com/google/common/base/SuppliersTest.java
@@ -278,8 +278,12 @@ public class SuppliersTest extends TestCase {
     assertNull(nullSupplier.get());
   }
 
-  @GwtIncompatible // Thread
+  public void testOfNullSuppliesNull() {
+    Supplier<?> tested = Suppliers.ofNull();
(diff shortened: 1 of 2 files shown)


netdpb:
I don't think `ofInstance(null)` is very awkward. And as you suggest, `null`s should be used sparingly, so I'm not sure we want to make that case any easier. :)

Also, I see only 141 cases of people using `Suppliers.ofInstance(null)` across all of Google's codebase, so I don't think it's worth adding to the API surface.

Does that make sense?

ronshapiro:
Also, with a lambda it's even simpler: () -> null

PR #3287, opened 2018-10-06
Narrow @Beta annotations to the methods of Futures

swankjesse:
Right now the @Beta annotation on the class is easy to miss, especially
since certain methods are also directly-annotated @Beta.

Remove the annotation from the class and apply it to each public member.

In a follow-up PR I'd like to remove the @Beta annotation from a few
methods (transformAsync, catchingAsync, immediateFuture).

### DIFF
--- a/guava/src/com/google/common/util/concurrent/Futures.java
+++ b/guava/src/com/google/common/util/concurrent/Futures.java
@@ -70,7 +70,6 @@ import org.checkerframework.checker.nullness.qual.Nullable;
  * @author Sven Mawson
  * @since 1.0
  */
-@Beta
 @GwtCompatible(emulated = true)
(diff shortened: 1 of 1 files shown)


ronshapiro:
The change LGTM. As for the removals, I know @cpovirk had plans for addressing some of these, but I can't find the doc (if there was one) and he's currently out on leave.

ronshapiro:
Hrm, those JDK 11 failures are bizarre. Maybe they're flaky?

cgdecker:
Ugh, actually I think the build is just failing on JDK 11 in general. I noticed that last week when I pushed the change adding JDK 11 to the travis build, but failed to do anything about it so far.

cpovirk:
Thanks. I've imported this into an internal CL I have going -- which will also remove a small number of `@Beta` tags as a starting point.

PR #3296, opened 2018-10-16
Added split to Guava immutable list metod

Dunemaster:
For those, who want to use ImmutableList and dont want to pay extra overhead of creating a unmodifiableWrapper or copying to ImmutableList


Dunemaster:
@lowasser , what would you suggest?

lowasser:
I would suggest doing nothing.  There's really no better thing to do than what's already there.

Issue #3285, opened 2018-10-05
Remove @Beta from Files


robeden:
The `Files` class has been around since 1.0, so it would be nice to remove it from beta. 
(FYI, our main need at the moment is the `asByteSource` method [related to #3284]).

LianyanGu:
Can I work on this issue?

glassm:
The @Beta annotation says to me, "I can't be trusted". It would be better to use alternative libraries for serious (production) applications. Please explain your policy for software stability and your use of @Beta annotation

netdpb:
We should absolutely review the use of `@Beta` on this class. However, Guava's `@Beta` definitely doesn't mean an API isn't trustworthy!

As documented [here](https://github.com/google/guava/blob/master/README.md#important-warnings) and [here](https://guava.dev/releases/snapshot/api/docs/com/google/common/annotations/Beta.html), APIs annotated with `@Beta` may change incompatibly in a future release (that is, there's no forwards-compatibility guarantee if you use them), but they're fine to use _except in libraries meant for downstream use_. If a library uses a `@Beta` API, then that library's _users_ may run into diamond dependency problems.

robeden:
@netdpb - Cool. Thanks for the reply. So... uh... `Files`? 🥺

kluever:
`@Beta` was removed from the class, but added to individual methods.

That being said, the method the OP asked about (`asByteSource`) was de-betaed, so I'm closing this. If folks have other specific APIs they want de-betaed, please open a new issue.

PR #3297, opened 2018-10-17
Removed @Beta from Files class

LianyanGu:
Issue number #3285


LianyanGu:
I signed it!

netdpb:
We are working on removing `@Beta` from APIs that are stable. Thanks for the PR, but we have to go through our internal decision process.

Issue #3303, opened 2018-10-20
{Boolean,Byte,Char,Double,Float,Int,Long,Short}s.concat() may throw {NegativeArraySize,ArrayIndexOutOfBounds}Exception


ineuwirth:
The `concat()` method in the per-type primitive utility classes throws unexpected `NegativeArraySizeException` or `ArrayIndexOutOfBoundsException` if the input arrays contain too many elements.

`NegativeArraySizeException` is being thrown if there is an overflow ending up with a negative `int`. In this case, the result array cannot be constructed with this number.

`ArrayIndexOutOfBoundsException` is being thrown if there is an overflow ending up with a non-negative `int`. In this case, the result array will be shorter than expected and `System.arraycopy()` will throw `AIOOBE`.

ineuwirth:
Test cases to demonstrate the issue can be found in #3304.

I am happy to cover other classes and modules (android, gwt) as well if you think the test cases/provided fix are good.

kluever:
This isn't actually a bug, right? It's just improving the API docs?

ineuwirth:
Well, a similar issue has been fixed in code instead of documenting the behavior e.g. in #3026.
Also, in case of an overflow ending up with a positive number, we could prevent a potentially big array from being allocated for nothing.

cpovirk:
I _think_ the PR from last year and the followup today have covered all this, but let me know if I'm missing anything.

PR #3304, opened 2018-10-20
Unhandled overflow in the `concat()` method of per-type primitive utility classes

ineuwirth:
The `concat()` method in the per-type primitive utility classes throws unexpected `NegativeArraySizeException` or `ArrayIndexOutOfBoundsException` if the input arrays contain too many elements.

Added test cases to demonstrate both overflow cases. A possible fix is provided.

In this commit, only the module `guava` and only the utility class `Longs` is covered along with its unit tests.

See https://github.com/google/guava/issues/3303

### DIFF
--- a/guava/src/com/google/common/primitives/Longs.java
+++ b/guava/src/com/google/common/primitives/Longs.java
@@ -247,19 +247,28 @@ public final class Longs {
    *
    * @param arrays zero or more {@code long} arrays
    * @return a single array containing all the values from the source arrays, in order
+   * @throws IllegalArgumentException if the total number of elements in {@code arrays} does not
+   *     fit in an {@code int}
    */
   public static long[] concat(long[]... arrays) {
-    int length = 0;
+    long length = 0;
     for (long[] array : arrays) {
       length += array.length;
     }
-    long[] result = new long[length];
+    checkNoOverflow(length);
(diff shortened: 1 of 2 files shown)


guava/src/com/google/common/primitives/Longs.java
```diff
@@ -247,13 +247,16 @@ public static long constrainToRange(long value, long min, long max) {
...
+   *     fit in an {@code int}
    */
   public static long[] concat(long[]... arrays) {
-    int length = 0;
+    long length = 0;
     for (long[] array : arrays) {
       length += array.length;
     }
-    long[] result = new long[length];
+    checkNoOverflow(length);
```
JakeWharton: `Math.toIntExact`?

ineuwirth: I preferred writing a custom checker to provide
* a better description ("the total number of elements in the arrays must fit in an int" over "integer overflow")
* a better exception type (`IllegalArgumentException` over `ArithmeticException` - I would be really confused seeing `ArithmeticException` after concatenating `long` arrays)

ineuwirth: (Of course I can replace to the JDK utility if you think it is not worth adding a custom one.)

Issue #3299, opened 2018-10-17
GraphBuilder.directed() and GraphBuilder.undirected() should not be parameterized on Object


ParceledTongue:
I feel like I must be missing something, but I can't figure out at all why these methods don't look like:

```
public static <N> GraphBuilder<N> directed() { ... }
public static <N> GraphBuilder<N> undirected() { ... }
```

The current design means that it's not possible to get a `GraphBuilder` parameterized on anything besides `Object` using these patterns until you call `nodeOrder`, which will properly infer the generic type and perform a cast. (Incidentally, it feels strange that `nodeOrder` has this generic downcasting behavior but, e.g., `allowsSelfLoops` does not.)

jrtom:
It's a bit nonintuitive, but it's the same thing that `MultimapBuilder` and `CacheBuilder` does, and for the same reasons.

Note that `build()` also performs the cast, of course.

The methods that don't take typed parameters (e.g. `allowsSelfLoops()`) don't do any casting because there's no type information to use.

ParceledTongue:
If you could elaborate or provide a link to what you're referring to when you say "the same reasons," I'd appreciate it. That may clear this whole thing up for me.

In the cases I'm thinking of, there is type information; it's just provided by the context, as it is when calling `build()`.

I understand why the generic type binding technically can't be put off any longer once you call `nodeOrder`, but that just seems to indicate that the binding should occur as early as possible . Otherwise `nodeOrder` is special compared to other operations, which feels pretty incidental.

```
GraphBuilder<String> builder1 = GraphBuilder.directed() // illegal
GraphBuilder<String> builder2 = GraphBuilder.directed().allowSelfLoops(true) // illegal
GraphBuilder<String> builder3 = GraphBuilder.directed().allowSelfLoops(true).nodeOrder(ElementOrder.unordered()); // legal
```

kevinb9n:
`nodeOrder` *is* special, because it naturally restricts what type of graphs the builder will be capable of producing.  You're proposing a way to "artificially" restrict the type of graphs ahead of time, but calling `nodeOrder` would *still* be "special" because unlike other methods you may break your compile if you hand it a conflicting type.

The equivalent CacheBuilder issue (resolved WAI) is issue 738.

kevinb9n:
Oh, I will say that CacheBuilder has the following doc, and GraphBuilder should probably have the equivalent:

```
@param <K> the most general key type this builder will be able to create caches for. This is
[...]

PR #3305, opened 2018-10-22
Moe Sync

ronshapiro:

Commits:
=====
<p> Enable goodtime-api for Guava (with some exceptions).

603ecca60c11d9a343423ae908bbc325521a7f50

-------

<p> *Builder.java: add @param documentation for the node/value/edge types that clarifies their behavior.

This is in response to (and should resolve) https://github.com/google/guava/issues/3299.

9db70c6519ef588697a9c522ee36aea439f36241

-------

<p> Suppress warnings in classes that implement equals() without also implementing
hashCode().

The contract for Object.hashCode states that if two objects are equal, then
calling the hashCode() method on each of the two objects must produce the same
result. Implementing equals() but not hashCode() causes broken behaviour when
trying to store the object in a collection.

73f7e84a345805dc4fcc9927201e03fdbf408c5f

### DIFF
--- a/android/guava-testlib/test/com/google/common/testing/EqualsTesterTest.java
+++ b/android/guava-testlib/test/com/google/common/testing/EqualsTesterTest.java
@@ -326,6 +326,7 @@ public class EqualsTesterTest extends TestCase {
       this.aspect2 = aspect2;
     }
 
+    @SuppressWarnings("EqualsHashCode")
     @Override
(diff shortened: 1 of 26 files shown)


ronshapiro:
Travis seemed to have missed this PR during Github's partial outage yesterday. Closing and reopening with the hopes that it will run the tests.

PR #3310, opened 2018-11-02
release of references in ExpiringMemoizingSupplier

tareqhs:
nullify the value reference before calling the delegate to avoid
unexpected leaks if the delegate fails

### DIFF
--- a/android/guava/src/com/google/common/base/Suppliers.java
+++ b/android/guava/src/com/google/common/base/Suppliers.java
@@ -240,8 +240,13 @@ public final class Suppliers {
       if (nanos == 0 || now - nanos >= 0) {
         synchronized (this) {
           if (nanos == expirationNanos) { // recheck for lost race
-            T t = delegate.get();
-            value = t;
+            // assign null if the delegate fails to avoid leaks in garbage collection
+            T t = null;
+            try {
+              t = delegate.get();
+            } finally {
+              value = t;
+            }
             nanos = now + durationNanos;
             // In the very unlikely event that nanos is 0, set it to 1;
             // no one will notice 1 ns of tardiness.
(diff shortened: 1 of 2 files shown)


netdpb:
Can you describe the bug you're trying to fix? Exactly what is leaked?

netdpb:
I see.  Thanks for the clarification.

android/guava/src/com/google/common/base/Suppliers.java
```diff
@@ -240,8 +240,13 @@ public T get() {
       if (nanos == 0 || now - nanos >= 0) {
         synchronized (this) {
           if (nanos == expirationNanos) { // recheck for lost race
-            T t = delegate.get();
-            value = t;
+            // assign null if the delegate fails to avoid leaks in garbage collection
```
netdpb: Could this be even simpler? Just set `value` to null before calling `delegate.get()`. Yes, there will be two volatile writes in the no-exception case, but that's negligible because we rarely get here.

```
value = null;
T t = delegate.get();
value = t;
```

tareqhs: This was my initial thought but I preferred avoiding a redundant write because mostly we do not fail and the redundancy is near constant here.. do you think it is worth it?

netdpb: From a colleague, and the original author of the lines in question:

> Concurrency is hard!
> 
> The current code relies on the value field never being invalidated.
Thread A may see that the time has not yet expired, then be indefinitely delayed, and then return value, so Thread B that is updating the value must not set it to null.
OTOH we would like to be able to null out expired values - we would even like to go further and null out expired values while we are computing a new value, so the GC can reclaim one humongous object graph while we are working on the next one.  But that will require more surgery.  It might help to be able to use VarHandles and/or StampedLock from recent jdk releases.  Or maybe reuse the strategy from StampedLock by hand.
> 
> Whenever we null out a field for the benefit of the GC, we we would like to use a plain (non-volatile) write, but there's no public API to do that on older JDKs.

That makes sense to me, and suggests that neither your approach nor mine is correct.

netdpb:
I'm going to close this PR for now. Feel free to reopen if you think we should proceed.

PR #3314, opened 2018-11-10
Field, parameter, and variable.

reudismam:
The final modifier can be used in fields, parameters, and local variables to indicate they cannot be re-assigned. This edit improves clarity and it helps with debugging.


kluever:
See https://github.com/google/truth/pull/501#issuecomment-437729683

PR #3315, opened 2018-11-10
Use the class Path to represent a path instead of a String.

reudismam:
Strings can be used to represent a file system path even though some classes are specifically designed for this task. For instance, java.nio.Path. It is useful to change the type of the variable to Path, since strings can be combined in an undisciplined way, which can lead to invalid paths. Second, different operating systems use different file separators, which can cause bugs.

### DIFF
--- a/guava/src/com/google/common/reflect/ClassPath.java
+++ b/guava/src/com/google/common/reflect/ClassPath.java
@@ -43,6 +43,8 @@ import java.net.URISyntaxException;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.nio.charset.Charset;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
@@ -497,7 +499,8 @@ public final class ClassPath {
      */
     @VisibleForTesting
     static URL getClassPathEntry(File jarFile, String path) throws MalformedURLException {
-      return new URL(jarFile.toURI().toURL(), path);
+      Path fpath = Paths.get(path);
+      return new URL(jarFile.toURI().toURL(), fpath.toString());
     }
   }
 


jbduncan:
This seems like a good change to me! It wasn't immediately obvious to me, but I see now that packing a string into a `Path` (even though we unpack it immediately afterwards) ensures that an exception is thrown if the passed string doesn't represent a valid path.

So as a non-Googler this change receives my :+1:.

jbduncan:
Although, I suppose I wonder if this change will work on Android? Does Android have `java.nio.file.Path` now? :thinking:

jbduncan:
And if so, should the `File` parameter for the method `getClassPathEntry()` be changed to a `Path` as well?

jbduncan:
Or even the `String` parameter for that matter?

liach:
is there more instances of such unsafe usage?

MarkMielke:
I think it is important to note that most of the reasons for and against using Path instead of String, also apply to File, and it would be useful to understand why File wasn't already used in these situations.

I don't think it being a Path, or a File, guarantees that it is "valid" as a path. It may improve the odds, by doing additional checks - but there are likely several paths that are in fact valid, that are not possible to represent with a Path. For example... when I call "mv some_file some_directory", I normally spell this "mv some_file some_directory/", as the trailing "/" is valid in a Unix path, and in the case of "mv" it will assert that the target is a directory before potentially overwriting a file of the same name:

```
$ touch x
$ touch y
$ mv x y/
mv: failed to access 'y/': Not a directory
$ mv x y
$ ls -l x y
ls: cannot access 'x': No such file or directory
-rw-rw-r-- 1 mark mark 0 Dec 24 19:07 y
```

A `Path` or `File` will have additional overhead in processing to ensure it is valid, and to track it. At a minimum, it is a `String` wrapped in another class object. However, every alteration or representation may have additional processing. What is the performance overhead of switching everything from `String` to `Path` or `File` from a runtime perspective? How much more memory does it require? How much more CPU does it require? How is `.toString()` implemented, as might be commonly called?

Generally, I like the idea of `Path` and `File`, which is why I have long used `File` in appropriate places, and I started to switch these to `Path` in Java 7 and later. However, I don't think it belongs everywhere. As an example - sometimes Integer are stored in String. While it makes sense to sometimes represent it as an Integer... sometimes it is simpler and faster, to leave it as a String.

ChristianCiach:
As far as I understand the method in question, it is used to construct a URL pointing to something inside a jar file. In particular, the String parameter is used to denote the path inside the given (jar-) File.

A `Path` is heavily dependent on the local operating- and file system. For example, under Windows, `Paths.get(...)` will return a different implementation of Path than under Linux. The contents of a jar file are surely not dependent of the file system of the local host, so using Path in this context is surely the wrong thing to do.

cpovirk:
I'll close this, since it's not clear what the problem being solved is. Feel free to reopen with a test that demonstrates an issue, and we can have a look.

PR #3324, opened 2018-11-25
Didn't update readme with latest version for Android.

abuicke:
Didn't update readme with latest version for Android.


ronshapiro:
This will be in our next public sync - we're just behind because of the holiday.

Issue #3360, opened 2019-01-07
Remove @Beta from EventBus


AlekssGu:
EventBus has been marked as Beta for 8 years already (since Jul 25, 2011).

nglorioso:
Thanks for the issue opening! We're looking through our `@Beta` API's and will keep this in mind.

philipwhiuk:
Did the review go anywhere @nglorioso?

SwingGuy1024:
Why is EventBus still @Beta? I keep getting inspection warnings from my IDE whenever I use it. It's a pretty old and stable class. Isn't it time to remove @Beta?

gm100861:
Please remove it, please, please , pleassssssssssssssssssssssssssssssse

AmerAlsabbagh:
Yes please remove it!

jerryoung:
EventBus is unstable? It make me puzzled. Anyone can talk me why?

nanachimi:
Yes please!

zhuhaoman:
Now is 2022, @Beta is still there...

PR #3362, opened 2019-01-07
Remove @Beta from EventBus #3360

rickybiswas25:


rickybiswas25:
I signed it

kluever:
We're likely going to start discouraging use of `com.google.common.eventbus` in favor of alternatives, so it won't be coming out of `@Beta`.

cgdecker:
Update: [it came out of `@Beta`](https://github.com/google/guava/commit/781ac8a447256fe0185ffdecd0fc993bbf84889f). But note that it mostly came out of `@Beta` because we're discouraging its use now, and won't be making further changes to it.

PR #3372, opened 2019-01-31
Improve efficiency of bulk removals in transformed lists.

hwaite:
Override removeRange(int,int) for Lists.TransformingSequentialList and
Lists.TransformingRandomAccessList so that source list has opportunity
to do more efficient deletions. Call fromList.subList(from, to).clear()
instead of removing with iterator.


hwaite:
Anything further needed from me in order to get this merged?

PR #3382, opened 2019-02-10
Document the JDK 9+ alternative to `Iterators#forEnumeration`

Stephan202:
- Copied the phrasing from the various `Streams.stream(Optional*)` methods.
- Didn't update the documentation in the Android code base. (Which also mirrors the approach taken for the `Streams.stream(Optional*)` methods.)
- Not completely sure whether the `UnmodifiableIterator`/`Iterator` distinction is worth pointing out.


Stephan202:
The JDK 11 build failures are unrelated to this change.

PR #3383, opened 2019-02-10
Document the JDK 9+ alternative to `MoreObjects#firstNonNull`

Stephan202:
Analogous to #3382.

PR #3384, opened 2019-02-11
Document the JDK 9+ alternatives to `Preconditions#{checkElementIndex,checkPositionIndexes}`

Stephan202:
Analogous to #3382 and #3383. I can imagine these two suggestions will cause some debate, since they have overloads that accept a `@Nullable String desc` for which there is no counterpart in the standard library.

One way to settle the debate could be to benchmark whether the JDK builtins are more efficient. (`Objects#checkIndex` is annotated `@ForceInline` and delegates to a `@HotSpotIntrinsicCandidate`, so it may be.)

### DIFF
--- a/guava/src/com/google/common/base/Preconditions.java
+++ b/guava/src/com/google/common/base/Preconditions.java
@@ -1300,6 +1300,8 @@ public final class Preconditions {
    * Ensures that {@code index} specifies a valid <i>element</i> in an array, list or string of size
    * {@code size}. An element index may range from zero, inclusive, to {@code size}, exclusive.
    *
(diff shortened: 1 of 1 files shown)


kevinb9n:
My bigger concern is that it's confusing enough as it is for users to understand and pay attention to the key difference between an "element index" and a "position index", and one of my goals here was to educate users on that difference and prompt them to choose wisely between the two. It's quite sad that the JDK chose to implement only one of them and not even explain the situation. In all, I don't think I really want to point people that way.

Stephan202:
@kevinb9n fully understand. I wonder whether the JDK implementers considered the Guava naming during their implementation process.

In any case: feel free to close this PR if/when the verdict is final :)

PR #3387, opened 2019-02-14
Deduplicate common logic in Uninterruptibles class, avoiding allocations

njhill:
I recently came across [this discussion](https://github.com/google/guava/issues/2384#issuecomment-181137856) around the reason for having duplicated logic in `Uninterruptibles`, and am offering this as an approach which avoids per-method call allocations and reflection. There is some overhead in terms of class count though, apologies if this was already discounted for that reason.

Also related: #1409

### DIFF
--- a/guava/src/com/google/common/util/concurrent/Uninterruptibles.java
+++ b/guava/src/com/google/common/util/concurrent/Uninterruptibles.java
@@ -43,27 +43,75 @@ import java.util.concurrent.locks.Condition;
 @GwtCompatible(emulated = true)
 public final class Uninterruptibles {
 
-  // Implementation Note: As of 3-7-11, the logic for each blocking/timeout
-  // methods is identical, save for method being invoked.
+  /**
+   * Shared logic
+   * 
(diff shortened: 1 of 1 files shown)


netdpb:
I'll let @cpovirk comment in more detail, but the number of classes is also a sensitive issue on Android, as well as classloading on static initializers (as you do with static final fields).

cpovirk:
There are also some performance concerns, even with more efficient approaches. It's not clear that they would actually matter in practice (or even exist, since I haven't benchmarked -- and probably wouldn't trust myself if I had :)), but, in particular, a large fraction of `Future.get` calls at Google are actually `Futures.getUninterruptibly` calls (by way of `Futures.getDone`), so some of these methods do get called a lot.

It's still possible that this is worthwhile, but in practice, we never have to edit the existing code, so I'm not too concerned about duplication from a maintenance perspective.

nglorioso:
Thanks for contributing, but I think we'll close this PR for now citing concerns about the class count.

PR #3392, opened 2019-02-24
CharMatcher: Improving the way it is deprecated

mirraj2:
I've updated the documentation to try to provide clearer instructions on ways that people can migrate away from some of the deprecated methods.

I also added a method called asciiDigit() because I think it is important to give developers an easy way to handle one of the most common use cases.

### DIFF
--- a/guava/src/com/google/common/base/CharMatcher.java
+++ b/guava/src/com/google/common/base/CharMatcher.java
@@ -192,6 +192,14 @@ public abstract class CharMatcher implements Predicate<Character> {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


mirraj2:
I signed it!

netdpb:
I don't think moving the information about alternatives to the deprecation tag is that helpful. `isDigit()` was never equivalent to 0123456789.

And I'm not convinced that `asciiDigit()` is useful enough to add weight to this method.

What might help with the latter is if you had data suggesting how many users would actually use `asciiDigit()`.

I'm closing this PR for now, but feel free to reopen if you have stronger data.

Thanks!

PR #3356, opened 2018-12-25
Remove unused imports and warnings

OleksiiBondar:
* Remove unused imports.
* Remove suppress warning annotations where not applicable and/or update suppression types

Issue #3340, opened 2018-12-17
Remove Beta from ListeningScheduledExecutorService?


pkwarren:
The `MoreExecutors#listeningDecorator(ScheduledExecutorService delegate)` method is not marked Beta but returns a ListeningScheduledExecutorService which is marked Beta. Is it planned to remove the Beta designation from the ListeningScheduledExecutorService interface in the future?

Today (in order to avoid errors about using beta APIs), consumers must downcast to a ListeningExecutorService or ScheduledExecutorService in order to use the returned result from this method.

Issue #3251, opened 2018-08-26
Remove @Beta from the immutable collector methods


zshamrock:
Like `ImmutableList.toImmutableList()`, `ImmutableMap.toImmutableMap()`, etc.

As per https://github.com/google/guava/wiki/PhilosophyExplained, specifically
> On the other hand, if you want something taken out of `@Beta`, file an issue. We generally promote features out of `@Beta` only when it's specifically requested, so if you don't ask, it won't happen.

According to the source these were introduced on version 21.0 and still marked as `@Beta`. As the latest IntelliJ IDEA now displays warnings on any beta/unstable API usage, wondering whether you have plans to remove or reason to keep `@Beta` flag on these API.

yuyumao:
I've opened a PR to remove them. https://github.com/google/guava/pull/3280

alimbada:
Hi, any chance `@Beta` can be removed from `MoreCollectors` methods as part of this issue too? If not, I'll create a separate issue.

Issue #3239, opened 2018-08-15
What's the status of ByteStreams API?


pluttrell:
The `com.google.common.io.ByteStreams` class is marked with `@Beta`, which I understand to mean that it's public API might change and thus we're warned about using it with libraries ([reference](https://google.github.io/guava/releases/26.0-jre/api/docs/com/google/common/annotations/Beta.html)).

Is there an ETA when the API might be considered stable enough to remove `@Beta`?

cgdecker:
`ByteStreams`, and much of `common.io`, is definitely on the list of things we should be trying to get out of `@Beta`, we just haven't been good about prioritizing reviewing things for graduation from beta. Thanks for the reminder.

kluever:
@elharo just requested (internally) a de-@Beta'ing of `ByteStreams.exhaust()`

PR #3393, opened 2019-02-26
Moe Sync

ronshapiro:

Commits:
=====
<p> Add Report-To header.

https://www.w3.org/TR/reporting/

504978f1032ab9b6284ffadd1f94bdb7c870557a

-------

<p> Remove unused imports and warnings.

Fixes #3356

f0945a2ca741ead2a435cf52ddae2217763cc092

-------

<p> Reformat with google-java-format

RELNOTES=N/A

7809097021927058b21523e7b4d24d1f7fdb91b8

-------

<p> Add missing @Override annotations

8843b6ca39c8baddcf01991ea34ad91f87b30d82

-------

<p> Document the retention of listeners added to Guava implementations of ListenableFuture.

RELNOTES=Document the retention of listeners added to Guava implementations of `ListenableFuture`.

f6287e1b85eb4daf5678d252aaaf215a82f4fd2d

-------

<p> Reformat with google-java-format

RELNOTES=N/A

28cff42bc92b7685937c5771868167e0c03155d3

-------

<p> Name the method parameter 'unused' for ImmutableSortedSet.readObject(ObjectInputStream) (so the static analysis doesn't complain)

0967c796d6e6be72b0d9e8d68af66d6a7b524b54

-------

<p> Remove @Beta from APIs:

collect
- most Collectors (aside from BloomFilter and ImmutableRange*, as those whole classes are @Beta (though I did remove the redundant @Beta annotation from the Collector-returning methods for clarity))
- Maps: asConverter, immutableEnumMap
- MultimapBuilder
- Streams: stream(Iterable), concat

io
- ByteStreams: copy, toByteArray
- CharStreams: copy, toString
- Files: as(Char|Byte)(Source|Sink)

util.concurrent
- Futures: addCallback, getDone, getUnchecked, immediateCancelledFuture, immediateFailedFuture, immediateFuture
- ListeningScheduledExecutorService
- Uninterruptibles

other
- HtmlEscapers
- Splitter.splitToList
- Ticker

Fixes https://github.com/google/guava/pull/3287
Fixes https://github.com/google/guava/issues/3251 (aside from ImmutableRange*, but those whole classes are @Beta, anyway)
Addresses the main concerns of https://github.com/google/guava/issues/3285 but doesn't cover the broader request
Fixes https://github.com/google/guava/issues/3340
Partially addresses https://github.com/google/guava/issues/3239

RELNOTES=Removed `@Beta` from a number of frequently used APIs.

068cb9005cc701ed98e31df9b56117b86df21d0f

-------

<p> Add some known UTF-8 hash values to Murmur3_32 tests. UTF-8 is more comparable between platforms than relying on Java's internal encoding.

e7ec53773bcf89533496de1611ac4dddd876fb54

-------

<p> Remove @CanIgnoreReturnValue from Ordering.sortedCopy() and immutableSortedCopy().

RELNOTES=Remove `@CanIgnoreReturnValue` from `Ordering.sortedCopy()` and `immutableSortedCopy()`.

da2895f5c3a55b84771488e0f8a38c67e0f63543

-------

<p> Make a constructor private.

(I'm looking into EnumMap stuff for other reasons, and I got to wondering if this was used outside the file. Nope.)

b1933ce7d8c8ff0fac90b494b33ab8af159e1906

-------

<p> Delete WellBehavedMap.

The bug it worked around was fixed in Java 7:
https://bugs.java.com/bugdatabase/view_bug.do?bug_id=6312706

And I don't believe it was ever present under Android:
https://android.googlesource.com/platform/libcore/+/fdb2704414a9ed92394ada0d1395e4db86889465/luni/src/main/java/java/util/EnumMap.java#146

I haven't looked for the absolute first version ever of GWT, but the still very old GWT file doesn't have the bug, either.

Deleting this class will simplify some work I'm doing around our EnumMap usage for j2cl.

2115c74c21ab47c3a41c2fb73a16aa6d051bf784

-------

<p> Internal GWT change

1d49ade94c3e7d2c056893a630739c8a6b897f66

-------

<p> Automated rollback of commit 1d49ade94c3e7d2c056893a630739c8a6b897f66

Internal GWT change

***

46a2123f4b077840b4b466fe6344579ac4b52bad

-------

<p> Use J2ObjC's @RetainedWith annotation instead of @Weak on ImmutableMap.keySet
to prevent crashes on iOS.

TESTED=Created a map inside a autorelease pool and saved a reference to the
keyset. Checked that the keyset functions correctly outside the autorelease
pool. Checked that both the keyset and the map are deallocated by the
autorelease pool when not holding a reference to the keyset. Did this test with
both ImmutableMap.of(<2 entries>) and ImmutableBiMap.of(<2 entries>).

RELNOTES=J2ObjC: Fixes crashes from use of `ImmutableMap.keySet()`.

76c954518752acb36256009840ed004eef412291

-------

<p> Annotate checkNotNull's input and output as @NonNull.

We don't want to go down the road of annotating most things @NonNull, as we eventually want to use tools that assume that that's the default. However, it's pretty weird for checkNotNull() to return a @Nullable value in the minds of some tools, so let's annotate just that method and see how it goes.

RELNOTES=Annotated `checkNotNull`'s input and output as `@NonNull`.

63d0b31e62b4d4da33ec61a1d84d026391ffd126

-------

<p> Document the JDK 9+ alternative to `Maps#immutableEntry`

Fixes #3385

998858f9abd06b3774bf72a3b4f2fdaa18a88ceb

-------

<p> Modernize book reference.

7eafede7552490351956e8038bb07596b227e3ef

-------

<p> Create constants for the SourceMap header

None of the tests like this header name, so each required some workarounds.

RELNOTES=Add the SourceMap header to HttpHeaders

81f191e80265001b3dc8b4544cd0bb277a05ba7b

-------

<p> Internal GWT change

db6ada7b4f9db07b8de6ec549ca7e2ab6079b354

-------

<p> Use J2ObjC's @RetainedWith annotation instead of @Weak on ImmutableMap.values
to prevent crashes on iOS.

TESTED=Created a map inside an autorelease pool and saved a reference to the
values collection. Checked that the values collection functions correctly
outside the autorelease pool. Checked that both the values collection and the
map are deallocated by the autorelease pool when not holding a strong reference
to the values collection. Did this test with ImmutableMap.of(<2 entries>) and
ImmutableMap.copyOf(<EnumMap with 2 entries>).

RELNOTES=J2ObjC: Fixes crashes from use of `ImmutableMap.values()`.

05ac56322f6c3a4b0109aab2b46671716f109cb5

-------

<p> Fix Javadoc breakage under JDK11 canary.

https://bugs.openjdk.java.net/browse/JDK-8212233
https://bugs.openjdk.java.net/browse/JDK-8193030

d8bdef58d54712286d86a9146778d25614fd3c9a

-------

<p> Remove GWT emulations that are no longer needed as of GWT 2.8.2.

RELNOTES=Removed GWT emulations that are no longer needed as of GWT 2.8.2. This means that GWT projects that use this version of Guava must use GWT 2.8.2 or higher.

be02df5ad4a530d5cb206a7042de8c2f5dbbff32


kevinmost:
> Annotate checkNotNull's input and output as @NonNull

Hi. Just curious about the motivation behind this decision (well, behind part of this decision). I totally agree that "it's pretty weird for checkNotNull() to return a @Nullable value in the minds of some tools", and am in favor of `checkNotNull()` returning a non-null value.

I'm just unclear on why the input is considered `@NonNull` now instead of `@Nullable`. Maybe I'm misusing it, but I use `checkNotNull()` as a fluent assertion on expressions where the compiler can't infer the nullability, but I know it's `@NonNull`. For my uses, the ideal signature would be:

```java
@CanIgnoreReturnValue
@NonNull
public static <T extends Object> T checkNotNull(@Nullable T ref) { ... }
```

PR #3408, opened 2019-03-19
Initialize uninterruptibles with current interrupt state of thread

belugabehr:
If a thread comes into Uninterruptibles already in an interrupted state, the first call to the target action will immediately fail with InterruptedException.  Initialize the 'interrupted' flag with the current state of the thread so that the first attempt can complete (assuming the thread is not interrupted again immediately after the initialization).


BELUGABEHR:
I signed it!

belugabehr:
@cpovirk Any thoughts on this PR? Thanks!!

cpovirk:
I believe we had felt that interruption was the less common case, so it wasn't worth sacrificing performance in the more common case by reading the interrupt bit.

That said, I don't think we have real-world numbers about _how_ much more common the common case is (nor about the relative cost of reading the interrupt bit vs. throwing an exception, though that would at least be easier to measure). Without such numbers, it's hard to justify making changes.

belugabehr:
@cpovirk Thanks for the review and consideration.  We can close this PR then.

Issue #3403, opened 2019-03-12
Remove @Beta from Futures.transform, catching, and *Async variants


cpovirk:
These methods are very heavily used internally. They're probably less so externally, but that may change as AndroidX exposes `ListenableFuture` in more APIs. Plus, Dagger Producers uses these methods.

Blockers:
- likely change to make `catching` not catch the `CancellationException` from cancellation (internal bug 27587642)
- possible changes to interruption behavior (https://github.com/google/guava/issues/1989)

cpovirk:
We could even try to change the parameter type to the JDK `Function` someday (using a trick to temporarily maintain binary compatibility), but I'm not sure we'd do it.

cpovirk:
https://github.com/google/guava/issues/3620 mentions the possibility of changing the return type of these methods (and others). But we could still do that after removing `@Beta` as long as we don't "just change" it: We'd have to inject bridge methods with the old signatures, as we already do for some other cases of signature changes.

cpovirk:
Anyway, my thinking lately has been that the possible interruption change doesn't necessarily block this, either, but that the possible `CancellationException` change for `catching` probably does.

I'm not entirely sure why I feel that way: Both are changes to behavior but not to mention signatures, and both are likely to have negative results for a very small percentage of callers. Maybe it's the fact that the `CancellationException` behavior is actually [documented](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/util/concurrent/Futures.html#catching(com.google.common.util.concurrent.ListenableFuture,java.lang.Class,com.google.common.base.Function,java.util.concurrent.Executor))?

I'm going to avoid making a snap judgment after not having thought about this in a long time. But maybe I can talk myself into this. Worst case, there are a number of other methods in `Futures` that we probably _can_ easily remove `@Beta` from.

kluever:
This is now fixed.

Issue #3253, opened 2018-08-28
Remove @Beta from the Striped


jsunsoftware:
Similar to [issues/3251](https://github.com/google/guava/issues/3251)

EarthCitizen:
This is incredibly frustrating. `Striped` is EXACTLY what I need, but I am making a library, and `Striped` is still `@Beta`, so I cannot use it. And, sadly, I am not aware of any alternative to Guava's implementation of this.

According to the JavaDoc of `Striped`, it has been around since Guava 13.0, which appears to have been released in 2012. As of this comment, that is approximately 10 years. How can the design not be finalized after 10 years? I don't get this.

Guava's infinitely long betas really deter me from using the library. Anything marked as `@Beta` is basically useless to me, and I imagine to many others. I appreciate the quality work that has been done, but 10 years just seems, from the outside at least, insane.

kluever:
Apologies for the inexcusably long delay here. This is now fixed.

PR #3410, opened 2019-03-21
Guava - debeta and spring clean

ThomasKoreman:
First of all this is my first project on Guava so I just did the simple part, removing @beta's and some deprecated methods that should've been removed earlier.
If there is anything incorrect to this PR please reply and feel free to give some feedback.
The de-``@beta`` stuff will close several issues already: #3403, #3360, #3253, (#3362 is also part of ``3360``).

### DIFF
--- a/guava/src/com/google/common/eventbus/EventBus.java
+++ b/guava/src/com/google/common/eventbus/EventBus.java
@@ -16,15 +16,15 @@ package com.google.common.eventbus;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
-import com.google.common.annotations.Beta;
-import com.google.common.base.MoreObjects;
-import com.google.common.util.concurrent.MoreExecutors;
 import java.lang.reflect.Method;
 import java.util.Iterator;
 import java.util.Locale;
 import java.util.concurrent.Executor;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+
+import com.google.common.base.MoreObjects;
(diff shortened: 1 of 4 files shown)


ThomasKoreman:
I signed it

ChristianCiach:
I am not a Google, but I can't see this PR being merged. You've reformatted the entire source code, so there is no way to tell what was actually changed. In general, pull requests should touch as little surrounding code as possible. 

Because of all this reformatting, the diff is useless. I cannot check if the removed deprecated methods are actually allowed to be removed. According to the guava philosophy, deprecated api will not be removed anymore unless it is marked as beta:

> APIs without @Beta will remain binary-compatible for the indefinite future. (Previously, we sometimes removed such APIs after a deprecation period. The last release to remove non-@Beta APIs was Guava 21.0.) Even @Deprecated APIs will remain (again, unless they are @Beta). We have no plans to start removing things again, but officially, we're leaving our options open in case of surprises (like, say, a serious security problem).

ronshapiro:
Until we've reviewed each of these tickets, I think it's worth holding off on this PR. We'll also possibly make those changes one-by-one.

cgdecker:
We will not be merging this. Aside from the unnecessary reformatting, removing `@Beta` from any API is something we need to make a decision on on a case-by-case basis (we should be prioritizing this higher than we have, but actually removing the annotation is the easy part). And `@Deprecated @Beta` APIs that we haven't removed yet mostly haven't been removed for internal reasons; we'll get to them.

PR #3412, opened 2019-03-22
Update checker-qual to 2.8.0, because it is now OSGi compatible

travishaagen:
Update checker-qual to 2.8.0, because it is now OSGi compatible. Prior to this fix, an OSGi application that uses Guava would encounter OSGi compatibility failures.

See also https://github.com/typetools/checker-framework/pull/2281

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -224,7 +224,7 @@
       <dependency>
         <groupId>org.checkerframework</groupId>
         <artifactId>checker-qual</artifactId>
-        <version>2.5.2</version>
+        <version>2.8.0</version>
       </dependency>
       <dependency>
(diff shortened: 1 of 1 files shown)


travishaagen:
I signed it!

travishaagen:
2.8.0 was released, so updated the PR to use that version instead of 2.7.0

travishaagen:
@cgdecker Can this be merged? This is a major problem for codebases that use OSGi.

WilliamCollishaw:
I believe this can be closed due to updating to 2.8.1 in #3463

cgdecker:
Yep, this has been addressed.

PR #3431, opened 2019-04-01
Moe Sync

ronshapiro:

Commits:
=====
<p> Remove unused configuration of load factor

This allows removal of loadFactor/threshold which saves 8 bytes per object.

While here:
- Reduce diff between Map/Set classes
- Make base forEach() work for Linked classes

Shallow object size (bytes):
- CompactHashMap: 64 -> 56
- CompactLinkedHashMap: 80 -> 72
- CompactHashSet: 40 -> 32
- CompactLinkedHashSet: 56 -> 48

062afc318cc7f60f6d681d6a3fa6cdc4f437a2d7

### DIFF
--- a/android/guava/src/com/google/common/collect/CompactHashMap.java
+++ b/android/guava/src/com/google/common/collect/CompactHashMap.java
@@ -96,10 +96,7 @@ class CompactHashMap<K, V> extends AbstractMap<K, V> implements Serializable {
     return new CompactHashMap<>(expectedSize);
   }
 
-  private static final int MAXIMUM_CAPACITY = 1 << 30;
-
-  // TODO(user): decide, and inline, load factor. 0.75?
-  static final float DEFAULT_LOAD_FACTOR = 1.0f;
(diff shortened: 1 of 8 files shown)


android/guava/src/com/google/common/collect/CompactHashMap.java
```diff
@@ -143,53 +140,34 @@
...
   /** Constructs a new empty instance of {@code CompactHashMap}. */
   CompactHashMap() {
-    init(DEFAULT_SIZE, DEFAULT_LOAD_FACTOR);
+    init(DEFAULT_SIZE);
   }
 
   /**
    * Constructs a new instance of {@code CompactHashMap} with the specified capacity.
    *
    * @param capacity the initial capacity of this {@code CompactHashMap}.
```
jbduncan: Oops, I believe either this `@param` should be renamed to `expectedSize` or the corresponding parameter should be named back to `capacity`. :)

ronshapiro: good catch!

Issue #3432, opened 2019-04-02
Cleanup DoubleUtils.bigToDouble


turbanoff:
Method `DoubleUtils.bigToDouble` can be replaced with direct `BigInteger.doubleValue` call.
[JDK-7131192](https://bugs.openjdk.java.net/browse/JDK-7131192) was fixed in JDK 8.

Klapsa2503:
Do you think that the method `DoubleUtils.bigToDouble` should be removed or just marked as depracated?

netdpb:
Do we know whether that fix was backported to Java 7, or is also evident in the Android SDK?

I don't think it's worth deprecating it, since it's not public. @lowasser, your thoughts?

Klapsa2503:
@netdpb i have deleted this method completely to clean things up.

lowasser:
I would specifically expect that we still need this method on Android.  No real objection to removing it for Java 8, other than the general desire to keep the implementations matching where possible.

umairaziz1:
👤

sananta:
Has this issue been taken care of? If not, I can take a look.

umairaziz1:
☺

kluever:
I think we'd prefer to minimize the diffs between the branches when possible.

PR #3433, opened 2019-04-03
DoubleUtils.bigToDouble replaced with BigInteger.doubleValue

Klapsa2503:


Klapsa2503:
I signed it!

nglorioso:
Per discussion on #3432 I think we'd prefer to keep the implementations the same between the branches if there's no other compelling reason to. Thank you for suggesting the changes though!

PR #3438, opened 2019-04-08
Use optimized for loop indexOf()

manu-p-1:
Added length locally to loop. Ultimately depends on compiler optimization.

### DIFF
--- a/guava/src/com/google/common/primitives/Bytes.java
+++ b/guava/src/com/google/common/primitives/Bytes.java
@@ -116,8 +116,8 @@ public final class Bytes {
     }
 
     outer:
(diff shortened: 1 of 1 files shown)


manu-p-1:
I can go ahead and write a benchmark (why not) and post the result here later on!

JakeWharton:
If you're already going this far, why not compute the subtraction only once?

manu-p-1:
You are right. I'm quite interested in seeing the results

ljacqu:
Guess the results weren‘t good :joy:

manu-p-1:
Super negligible performance diff

PR #3439, opened 2019-04-10
Upgrade error_prone

elharo:
to sync with the version protobuf is now using

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -229,7 +229,7 @@
       <dependency>
         <groupId>com.google.errorprone</groupId>
         <artifactId>error_prone_annotations</artifactId>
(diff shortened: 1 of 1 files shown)


cpovirk:
Thanks. Would going all the way to 2.3.3 be worse for your dependency-agreement efforts?

elharo:
For weird and possibly arbitrary reasons, it's not quite as problematic if Guava jumps ahead of the pack; but it's still better if everyone can agree. I picked 2.3.2 only because protobuf already shipped with that version. I am thinking about how we might get some approximation of ONE_VERSION_ONLY on Github, but it's tricky.

Issue #3044, opened 2018-02-07
Allow RateLimiter to check IF it would acquire the permit immediately


andviane:
Administrators who do load balancing asked me to provide the external REST API to check if my web service is overloaded or not. The "overloaded service" means that its RateLimiter (that is guarding the service from overload) is engaged, actively limiting the traffic. I just call withing request tryAcquire() and simply return 429 (too many requests) if it returns false. 

I could call "tryAcquire" from the overload check URL as well, and may work, but this would indeed acquire the unnecessary permit. Depending on how often the admins wish to ping the overload check endpoint, this may limit the main traffic well below permitted. 

I think I need a simple boolean method that would look like tryAcquire() but would not actually acquire any permits, would just check if it could (immediately). Alternatively, tryAcquire(int numPermits) could accept 0 as a valid parameter, meaning that no permits should be acquired, but the possibility to get one if asked must be checked.

I see the method canAcquire(long, long) in the source code of RateLimiter.java that probably would not require much refactoring to get the functionality needed. If the community agrees, I could also open the pull request.

cpovirk:
We even have this internally. It's not too widely used, but it's probably simple enough to expose it. I'll check with the rest of the team.

StephenFlavin:
Hey I'd love to see this added, I raised a draft PR shown above which exposes the `canAcquire()` method as well as a method which returns the micros until the next permit is available which I would like to use as a comparable.

umairaziz1:
Critical

PR #3446, opened 2019-04-17
RateLimiter: Expose CanAcquire and QueryEarliestAvailable Methods

StephenFlavin:
In reference to https://github.com/google/guava/issues/3044 this PR exposes the methods which are acquire uses internally to check **if** a RateLimiter can be acquired and additionally a method which returns the time until the next permit is available `microsToNextPermit()`.

My reasoning for using these methods is I would like use `microsToNextPermit()` as a way to sort a List of objects that have a RateLimiter e.g.
```
@Override
public int compareTo(T o) {
    return this.rateLimiter.microsToNextPermit() - o.rateLimiter.microsToNextPermit();
}
```


StephenFlavin:
I signed it!

Stephan202:
W.r.t. the suggested use case: the shown `compareTo` is not guaranteed to adhere to the method's general contract. Consider comparing a rate limiter to itself: the result may be 0, but if the clock advances between two invocations of `microsToNextPermit`, the result will be positive.

StephenFlavin:
@Stephan202 good point 😞 
I could potentially add a version of `acquire()` which returns `microsToNextPermit()` to be stored by the calling code?
e.g.

```
public class Foo implements Comparable<Foo>, Runnable {

    private long nextAvailableMili;
    private RateLimiter rateLimiter;

    public Foo(RateLimiter rateLimiter) {
        this.rateLimiter = rateLimiter;
        this.nextAvailableMili = Instant.EPOCH.toEpochMilli();
    }

    @Override
    public void run() {
        nextAvailableMili = Instant.now().plus(rateLimiter.acquire(), ChronoUnit.MICROS).toEpochMilli();
        System.out.println("next available at: " + nextAvailableMili);
    }

    @Override
    public int compareTo(@NotNull Foo o) {
        return Ints.saturatedCast(this.nextAvailableMili - o.nextAvailableMili);
    }

}
```

StephenFlavin:
@Stephan202 what's the merge policy for these changes, I've just rebased on master to fix the build

guava/src/com/google/common/util/concurrent/RateLimiter.java
```diff
@@ -421,6 +421,18 @@ public boolean tryAcquire(int permits, long timeout, TimeUnit unit) {
...
 
+  public long microsToNextPermit() {
+    return queryEarliestAvailable(stopwatch.readMicros());
+  }
+
+  public boolean canAcquire() {
+    return canAcquire(stopwatch.readMicros(), 0);
+  }
+
+  public boolean canAcquire(long timeoutMicros) {
```
Stephan202: Given Google's recent "GoodTime" effort I expect that this method should be replaced with `canAcquire(Duration timeout)`.

StephenFlavin: Finally got around to updating the PR with your suggestions. 😅

StephenFlavin:
I've updated this PR to use Durations however I kept the footprint small and avoided the larger refactor.
I'm happy to spend the time doing the refactor if it's required to get this merged.
I would appreciate your thoughts @cpovirk

guava/src/com/google/common/util/concurrent/RateLimiter.java
```diff
@@ -421,6 +422,18 @@ public boolean tryAcquire(int permits, long timeout, TimeUnit unit) {
     return true;
   }
 
+  public Duration durationToNextPermit() {
+    return Duration.of(queryEarliestAvailable(stopwatch.readMicros()), ChronoUnit.MICROS);
```
MikeDombo, line 426: This time should not be accurate, you need to use 
```
long nowMicros = stopwatch.readMicros();
return Duration.of(max(queryEarliestAvailable(nowMicros) - nowMicros, 0), ChronoUnit.MICROS);
```

StephenFlavin: Ah good spot, I'll update the PR soon and look at adding some unit tests.

StephenFlavin: Code fixed and unit tests added 🙂

MikeDombo (approved):

stunnamaneee:
Thanks sorry i have not answered

wuzm2k05:
I need this PR, anyone can tell me is there a plan to merge this change?

PR #3448, opened 2019-04-27
Add java.util.UUID to ArbitraryInstances

petedmarsh:
I hope this counts as a trivial change. This adds `java.util.UUID` to `ArbitratyInstances` - I have recently been working on a project where we used `UUID` and `NullPointerTester` and we've often had to set a default manually, this should avoid that.


petedmarsh:
The build failure seems to be a problem with the build/travis rather than this change (it happens with other PRs too), but happy to make changes if that's not the case.

PR #3465, opened 2019-05-09
Define project build source encoding as UTF-8

WilliamCollishaw:
At build time the following warning appears:
`[WARNING] Using platform encoding (UTF-8 actually) to copy filtered resources, i.e. build is platform dependent!
`

Defining a project build source encoding removes the warning and can prevent platform dependent issues in the future.

Fix summary by maven [here](http://maven.apache.org/general.html#encoding-warning) and [here](https://maven.apache.org/plugins/maven-resources-plugin/examples/encoding.html)


ronshapiro:
Thanks for all of the fixes!

PR #3471, opened 2019-05-13
Fix "Apache License, Version 2.0" spelling

don-vip:
There are many Java libraries licensed under "Apache License, Version 2.0" that do not use its official spelling.
This causes issues like https://issues.apache.org/jira/browse/MPIR-382: with every library defining its own spelling, it's difficult in large projects to have a clear view of all licenses in use.
This PR changes the license spelling to the official one, as advised by Maven developers.

### DIFF
--- a/android/guava-bom/pom.xml
+++ b/android/guava-bom/pom.xml
@@ -29,7 +29,7 @@
 
   <licenses>
     <license>
-      <name>The Apache Software License, Version 2.0</name>
+      <name>Apache License, Version 2.0</name>
       <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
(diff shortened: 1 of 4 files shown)


ronshapiro:
Thanks! I'll import this

WilliamCollishaw:
This license naming format is used in multiple pom files.
Screen Shot 2019-05-20 at 4 26 05 PM

don-vip:
Thanks! I have amended my commit.

ronshapiro:
Whoops I already submitted this. I'll submit another.

ronshapiro:
As an aside - what if you grouped licenses based on URL and not based on title? Wouldn't that dedupe well?

PR #3472, opened 2019-05-14
change keyToKeyList from HashMap to LinkedHashMap

aikawarazu:
in  constructor method ,keyToKeyList was set as HashMap ,keyToKeyList should keep insertion-order   .
like in readObject()  method do : keyToKeyList = Maps.newLinkedHashMap();


lowasser:
Can you include a test of the public API that fails before and passes after this change?

ronshapiro:
Please make sure to sign the CLA - without that, we cannot accept any contributions.

cgdecker:
The CLA isn't signed and no rationale was provided for making this change, so I'm closing it.

PR #3476, opened 2019-05-17
Upgrade Maven Javadoc Plugin to 3.1.0

WilliamCollishaw:
Upgrade Maven Javadoc Plugin from 3.0.1 to [3.1.0](https://blog.soebes.de/blog/2019/03/04/apache-maven-javadoc-plugin-version-3-dot-1.0-released/)
Additionally added wildcard syntax to the list of excluded package names due to failing without it after the upgrade

Comment from @njhill in PR #3370 notes a javadoc plugin issue with JDK 11 that should be resolved in this most recent version.

### DIFF
--- a/guava/pom.xml
+++ b/guava/pom.xml
@@ -128,7 +128,7 @@
           <sourcepath>${project.build.sourceDirectory}:${project.build.directory}/jdk-sources</sourcepath>
 
           <!-- Passing `-subpackages com.google.common` breaks things, so we explicitly exclude everything else instead. -->
-          <excludePackageNames>com.google.common.base.internal,com.google.thirdparty.publicsuffix,com.oracle,com.sun,java,javax,jdk,org,sun</excludePackageNames>
+          <excludePackageNames>com.google.common.base.internal.*,com.google.thirdparty.publicsuffix.*,com.oracle.*,com.sun.*,java.*,javax.*,jdk.*,org.*,sun.*</excludePackageNames>
 
           <!-- TODO(cpovirk): Move this to the parent after making the package-list files available there. -->
           <!-- We add the link ourselves, both so that we can choose Java 9 over the version that -source suggests and so that we can solve the JSR305 problem described below. -->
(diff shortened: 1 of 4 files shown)


ronshapiro:
thanks! I'll import this shortly!

guava/pom.xml
```diff
@@ -128,7 +128,7 @@
           <sourcepath>${project.build.sourceDirectory}:${project.build.directory}/jdk-sources</sourcepath>
 
           <!-- Passing `-subpackages com.google.common` breaks things, so we explicitly exclude everything else instead. -->
-          <excludePackageNames>com.google.common.base.internal,com.google.thirdparty.publicsuffix,com.oracle,com.sun,java,javax,jdk,org,sun</excludePackageNames>
+          <excludePackageNames>com.google.common.base.internal.*,com.google.thirdparty.publicsuffix.*,com.oracle.*,com.sun.*,java.*,javax.*,jdk.*,org.*,sun.*</excludePackageNames>
```
cpovirk: This PR produces a change in behavior: With this PR applied, the generated Javadoc contains entries for `com.google.common.base.internal`, `com.google.thirdparty.publicsuffix`, and `jdk`. Perhaps we need the `.*` only when that exact package doesn't exist?

Maybe you can find a way that works, but it feels unfortunate to have to further hardcode which packages exist. I guess we may need to cross this bridge eventually, unless we can hope that a future change to the Maven Javadoc Plugin will undo this change?

WilliamCollishaw: Good catch, it looks like I can use the wildcard `:*` instead of `.*` which will exclude those exact packages, [reference](https://maven.apache.org/plugins/maven-javadoc-plugin/examples/exclude-package-names.html).
I will try test this later.

WilliamCollishaw: > Good catch, it looks like I can use the wildcard `:*` instead of `.*` which will exclude those exact packages, [reference](https://maven.apache.org/plugins/maven-javadoc-plugin/examples/exclude-package-names.html).
> I will try test this later.

Please ignore this comment, i misread the documentation

WilliamCollishaw:
Butchered this branch, reopened new PR #3478

guava/pom.xml
WilliamCollishaw: @cpovirk this change in behavior is fixed in #3478

Issue #1872, opened 2014-10-31
InternetDomainName.isValid is not RFC1123 compliant


gissuebot:
_[Original issue](https://code.google.com/p/guava-libraries/issues/detail?id=1872) created by **manderson23** on 2014-10-27 at 10:59 PM_

---

InternetDomainName.isValid returns false for the parameter "8server".

RFC1123 suggests that names should be able to start with a digit. See https://groups.google.com/forum/#!topic/guava-discuss/8Sycya7Fkok on guava-discuss for a discussion of the issue. 

I also asked for clarification on Server Fault at http://serverfault.com/questions/638260/is-it-valid-for-a-hostname-to-start-with-a-digit

My final suggestion on the mailing list for solving this was as follows:

"What about if the final part starts with a digit actually checking if the complete name is an IP address e.g. calling InetAddreses.isInetAddress? The discussion section in RFC 1123 does suggest a full syntactic check to prevent a dotted decimal number getting through."

but there was no response.

tgpfeiffer:
I was directed here in a discussion of https://issues.cloudera.org/browse/DISTRO-638 ("Cannot install Cloudera Manager Agent if hostname starts with a digit") where the origin of that issue is that Docker generates hostnames as random hexadecimal strings, which do therefore quite often start with a digit. (And they refuse to change that, insisting that it is totally RFC-compliant to start with a digit, as long as there is also any normal letter in the hostname: https://github.com/docker/docker/issues/7756, https://github.com/docker/docker/pull/8194)

I would therefore very much like to see a change of behavior from "fail if first character of TLD is a digit" to "fail if all characters of TLD are digits".

frank-dspeed:
tgpfeiffer there is a patch for that you can even patch it easy your self with simply adding a random letter fixed in front of the container name via editing docker src and making a fork

PR #3482, opened 2019-05-25
RFC1123 Compliant InternetDomainName Test Cases

WilliamCollishaw:
RFC1123 compliant InternetDomainName test cases

Add tests to prove URLs with leading digits are valid and containing only digits are valid.
#1872 was resolved at some point from 2014 to now without the issue getting closed. This PR adds test cases proving this issue can be closed. 

Some sample real domains from related to these tests would be [www.9gag.com](https://www.9gag.com/) and [www.9292.nl](https://www.9292.nl/)


WilliamCollishaw:
@cpovirk I believe #1872 can be closed now

frank-dspeed:
LGTM +1

PR #3485, opened 2019-05-26
Clarify behavior of `ImmutableList#copyOf` for null argument.

AnuragPorwal:


ljacqu:
Afaik this is the behavior of all ImmutableCollection types. The class javadoc on ImmutableList explicitly refers to the ImmutableCollection which mentions this at the top. So these changes don‘t seem necessary.

cpovirk:
The factories are tested by other automated null tests, so let's not add more tests.

The existing doc is trying to say that the methods reject null elements, but it's saying it poorly, so thanks for raising this. How about this...?

```
@throws NullPointerException if {@code elements} contains a null element
```

That skips mentioning that `elements` itself might be `null`, but we always skip that, preferring to annotate the parameters that _can_ be `null` with `@Nullable`. (Otherwise, nearly all of our methods would have `@throws NullPointerException if x, y, or z is null`.)

PR #3492, opened 2019-06-04
LocalCache clean up

WilliamCollishaw:
Remove two redundant if statements
Removed some unused imports
Comment grammar fixes


ljacqu:
I don‘t really get the point of these changes. To me it seems that this was just the IDE showing up warnings tbh and that it was fixed without really considering if it makes it easier to read or not

nglorioso:
I appreciate raising this PR.

In the particular circumstances here, I think we'll keep the if conditions as is, as their current structure makes the conditions they intend to check more clear. 

I will fix the typo for 'accommodate' you raised though, thanks!

PR #3497, opened 2019-06-07
Upgrade animal sniffer version to 1.18

WilliamCollishaw:
Version [1.18](https://github.com/mojohaus/animal-sniffer/milestone/5?closed=1) resolves a JDK8 compatibility issue, resource leak and ends support for JDK 7

### DIFF
--- a/android/pom.xml
+++ b/android/pom.xml
@@ -14,7 +14,7 @@
     <!-- Override this with -Dtest.include="**/SomeTest.java" on the CLI -->
     <test.include>%regex[.*.class]</test.include>
     <truth.version>0.44</truth.version>
-    <animal.sniffer.version>1.17</animal.sniffer.version>
+    <animal.sniffer.version>1.18</animal.sniffer.version>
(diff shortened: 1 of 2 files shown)


WilliamCollishaw:
Oops, forgot about the android JDK 7 compilation. Should I remove the upgrade for Android or close PR?

cpovirk:
The JDK 7 failure appears to be a case in which Animal Sniffer is catching a real problem (implicit use of `ReflectiveOperationException`) that it was missing before. That's interesting because I've seen Animal Sniffer catch this problem in the past -- though maybe it was in a different project that was using a newer version?

I'll take care of the fix.

cpovirk:
Oh, wait: I think what we saw was that Animal Sniffer caught the problem _if a method was called on the `ReflectiveOperationException`, which is the only scenario in which we'd seen it cause problems in practice. But we reasoned that other references to `ReflectiveOperationException` could cause problems (maybe for debugging or something) that we just hadn't seen yet, so it's nice that Animal Sniffer detects them now.

PR #3498, opened 2019-06-07
Remove single character string concatenation - EventBus

WilliamCollishaw:
Minor performance improvement. 
[Stack Overflow discussion on this optimization](https://stackoverflow.com/questions/24859500/concatenate-char-literal-x-vs-single-char-string-literal-x)


cgdecker:
The performance impact of this is going to be effectively 0; I don't think it's worth making this change.

PR #3502, opened 2019-06-12
Fixed double slicing ByteSource

### LINKED ISSUE #3501: Slicing a ByteSource twice throws an unexpected IllegalArgumentException
Currently, trying to slice a `ByteSource` twice throws an unexpected `IllegalArgumentException` if the `ByteSource` returned after the first slice has a length smaller than the offset of the subsequent slice. For example, this code throws an `IllegalArgumentException`:

```
import com.google.common.io.ByteSource;

public class GuavaTest {
        ByteSource.concat().slice(0, 3).slice(4, 3);
    }
}
```

That is against the documentation, where it is specified that it should return an empty `ByteSource`.

Maxi17:
Fixes #3501

However, there is still a problem in this method. If `this.offset + this.length > Long.MAX_VALUE`, it will overflow. These are 4 possible approaches:

1. Leave is as it is.
2. Disallow offset + length > Long.MAX_VALUE on slices by adding checks.
3. Allow offsets larger than Long.MAX_VALUE.
4. Assume that no ByteSource has that much elements, and if the offset would overflow, return an empty source.

### DIFF
--- a/guava-tests/test/com/google/common/io/ByteSourceTest.java
+++ b/guava-tests/test/com/google/common/io/ByteSourceTest.java
@@ -221,6 +221,25 @@ public class ByteSourceTest extends IoTestCase {
   }
 
   /**
+   * Tests that slice() works correctly when the source is sliced two times, when the second slice
+   * starts at an offset greater than the length of the previous slice.
+   */
+  public void testSlice_slicingAfterSlicing() throws IOException {
+    // Source of length 10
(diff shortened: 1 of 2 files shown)


Maxi17:
I signed it!

cpovirk:
Somehow we ended up merging https://github.com/google/guava/pull/3829 instead of this one. Sorry about that.

Issue #3503, opened 2019-06-17
Long overflow in cache statistics


timw6n-thg:
We use Guava caches inside a large enterprise application. 

After a few days of runtime calling the `stats()` method on heavily-used caches throws an `IllegalArgumentException` at the equivalent line to [CacheStats.java:87](https://github.com/google/guava/blob/master/guava/src/com/google/common/cache/CacheStats.java#L87), presumably because the total load time has overflowed a long and looped back round to a negative value.

It looks like this is down to the load time stopwatch being started at the point that an entry is submitted for asynchronous refresh, i.e. added to the refresh queue. Our largest queues are drained in batches and end-up with usual sizes in the tens of thousands — so the nanosecond-precision load time increases *very* quickly under these circumstances.

I am aware that it is a design decision for the cache statistics to be [monotonically increasing](https://github.com/ben-manes/caffeine/issues/215#issuecomment-356053115), so a reasonably sane solution might be to simply drop the non-negative validation as part of the CacheStats constructor, or try to somehow wrap back to zero instead of `Long.MIN_VALUE`. Metrics systems using similar to Graphite's `nonNegativeDerivative` should handle that gracefully anyway.

NB: I think this issue also affects Caffeine caches in the same way.

ben-manes:
Thanks, I'll wait to see what the Guava team decides and emulate. If this is not resolved timely, then I'll probably add a `Math.max(0, totalLoadTime.sum())` to the default `StatsCounter`. Since an overflow occurs after 292 years, it should be rarely hit and in Caffeine you can supply your own `StatsCounter`, e.g. to publish negative values if desired, making it easy to workaround. I slightly lean against removing the negative check in `CacheStats`, but I will follow Guava's direction if they choose to.

timw6n-thg:
Thanks @ben-manes. Sounds good.

Just to give a more detailed example of how we're seeing this seemingly very unlikely behaviour:

Imagine a cache with a few hundred thousand entries, with "refresh after write" of 5 minutes so that frequently-requested items are reasonably fresh.

The `reload` method on the loader adds the element to a queue which, to optimise network calls to a slow downstream service, drains slowly in large batches, i.e. with bulk calls.

With an average queue size of 50k, we are effectively accruing 5.7 years (50,000 hours) of load time per hour, well over a hundred years per day. Hence seeing the overflows after only a few days of the application running in production.

ben-manes:
[...]

PR #3519, opened 2019-06-28
Moe Sync

ronshapiro:

Commits:
=====
<p> Add missing description tags to guava-parent poms.

We were previously apparently inheriting them from sonatype-oss-parent, until we removed that. Releases don't work without a description. "Invalid POM: /com/google/guava/guava-parent/28.0-android/guava-parent-28.0-android.pom: Project description missing"

c417618b2be5c127baf8a739789615b3f5c5fe7c

-------

<p> Fix typo in 'stoppingCount' variable name

Fixes #3500

c168ba2f5939102ee0b3e3cbd941f213d3ff8e8a

-------

<p> Split catch() blocks so that javac doesn't generate a reference to their common supertype, ReflectiveOperationException.

ReflectiveOperationException isn't available on the old versions of Android that we support.

As far as we know, there is no actual problem with these references because no methods are called on them. However, the new version of Animal Sniffer we're updating to will flag them anyway, and I think that avoiding them is the safe thing to do.

Prepares for #3497

15c9a9385713910a1ecf9e5a5e6e50e13755fef3

-------

<p> Upgrade animal sniffer version to 1.18

Fixes #3497

afdecf706fa86c5d4760b8211df2b00806e9458c

-------

<p> Document default values of builder setters.

d5dca3613d5a47b87ed88a332e35e2f60a025630

-------

<p> Use saturatedToNanos() in CacheBuilder to avoid overflows.

RELNOTES=Use saturatedToNanos() in CacheBuilder to avoid overflows.

7d04f72bf355e9bac1312debb3af7dca28833077

-------

<p> Tweaks to the list of Guava features.

4dbc7c1a545bdf5dd20a5934ff66c171806957db

-------

<p> Add more @Nullables to variadic parameter lists.

RELNOTES=Add more @Nullables to variadic parameter lists.

fbae3f47239fbea0555b4df2c4aea474f23d4b70

-------

<p> Halve memory consumption when creating a Bloom filter.

AtomicLongArray(long[]) clones its input, which is a waste when we're passing
in a new array.

RELNOTES=Halve VmPeak when creating a Bloom filter.

f4998201486012d050092ebbdbfb10713e9df549

-------

<p> Use LongMath.saturatedAdd/Subtract in CacheStats.

Fixes https://github.com/google/guava/issues/3503

RELNOTES=avoid overflows/underflows in CacheStats

9f3d0481768aa3b0e08de20ab881df4579d6ad18

-------

<p> Fix the travis build

ba4111cb700754c4f6e9be3a5a35ab420610e2eb

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,14 +3,8 @@ sudo: false
 language: java
 
 jdk:
-  - oraclejdk8
+  - openjdk8
   - openjdk11
-
-# https://github.com/travis-ci/travis-ci/issues/3259#issuecomment-130860338
-addons:
-  apt:
-    packages:
-      - oracle-java8-installer
 
 install: mvn -B -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn install -U -DskipTests=true -f $ROOT_POM
 
--- a/README.md
+++ b/README.md
@@ -4,9 +4,8 @@
 [![Build Status](https://travis-ci.org/google/guava.svg?branch=master)](https://travis-ci.org/google/guava)
 
 Guava is a set of core libraries that includes new collection types (such as
-multimap and multiset), immutable collections, a graph library, functional
-types, an in-memory cache, and APIs/utilities for concurrency, I/O, hashing,
-primitives, reflection, string processing, and much more!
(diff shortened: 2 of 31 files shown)


ben-manes:
@kluever I think you missed an overflow check in `SimpleStatsCounter#snapshot()`.

PR #3527, opened 2019-07-09
Transfer buffer reuse optimization

berndhopp:
### Objective
The objective of this PR is to avoid unnecessary allocation of memory to create transfer-buffers to e.g. move data from an InputStream to an OutputStream by introducing a package-local class TransferBuffer to the com.google.io package. 

There had been a similar PR from me before, this is a complete rewrite based on the original idea. This is a purely internal optimization that does not change any public API. 

### Benchmarks

The microbenchmarks show that avoiding buffer allocation yields a significant performance boost

| tested method | bytes copied | runtime (ms) with patch | runtime (ms) without patch | performance increase in % |
| --- | --- | --- | --- | --- |
| ByteStreams.copy | 10 | 105569 | 681130 | 545,2 |
| ByteStreams.copy | 100 | 107411 | 698152 | 549,98 |
| ByteStreams.copy | 10000 | 840354 | 1186307 | 41,17 |
| ByteStreams.readBytes | 10 | 68917 | 633365 | 819,03 |
| ByteStreams.readBytes | 100 | 69802 | 660336 | 846,01 |
| ByteStreams.readBytes | 10000 | 360357 | 934588 | 159,35 |
| ByteSource.contentEquals | 10 | 647140 | 3614650 | 458,56 |
| ByteSource.contentEquals | 100 | 685559 | 647694 | 453,81 |
| ByteSource.contentEquals | 10000 | 5987317 | 9460941 | 58,02 |

### Profiling

To have a deeper look at the memory- and GC-profile, I created another JMH-benchmark [here](https://github.com/berndhopp/guavabenchmark). 

CopyToBenchmark#testCopy() runs 10000 ByteStreams#copyto(inputStream, outputStream)-methods sequentially in a single thread. The first image is the profile on the master-branch, the second on the transfer_buffer_optimization-branch. Memory consumption and gc-activity are both multiple times lower on with transfer-buffer reuse. 

image (1)
image 1: profile of CopyToBenchmark#testCopy on master-branch

image (2)
image 2: profile of CopyToBenchmark#testCopy on transfer_buffer_optimization-branch


berndhopp:
I pushed some small typo- and formatting fixes in TransferBuffer.java in commits I squashed into the original commit, so the git history looks cleaner

cgdecker:
This general idea has come up before, and while it looks good on paper (and in terms of the runtime numbers you've provided) I have a number of concerns with it:

- I feel like holding onto 8k of memory indefinitely is a very unexpected side-effect of a static method, especially since it gives the user no control over that.
- Soft references are nice in theory but not great in practice. Again, I don't like the idea of using them without the user having any control over their use.

Basically, while this certainly looks like a win in terms of JVM runtime numbers, it makes tradeoffs that have unclear costs. The current implementation is at least clear in that it doesn't hold onto extra memory period.

A related idea I've had is creating some class with _instance_ methods for I/O methods like these, which wraps something the user has full control over like a buffer pool. But it's not something I've had a lot of time to think about and is obviously more difficult to use unless you're using dependency injection.

Anyway, I'll try to give this some more thought.

berndhopp:
@cgdecker @kluever any news here? I think the benefits of this are obvious and I couldn't produce any scenario with non-neglible drawbacks in terms of memory consumption or runtime performance. 

best
Bernd

berndhopp:
@cgdecker @kluever how about now?

berndhopp:
@cgdecker any news here?

berndhopp:
@cgdecker  any news here?

cpovirk:
At some point in these past few years, @cgdecker and/or I ended up with some additional concerns about using static buffers:
- Data from one part of a problem can become visible to another part of a program. I would _hope_ that this is not a practical security problem for anyone, but I'd much rather play it safe when it comes to security.
- `ThreadLocal` causes additional problems for Loom users (who may have a very large number of virtual threads, each of which may end up with a buffer).

We've been keeping this issue open on the theory that we might explore [letting users specify their own buffer pool](https://github.com/google/guava/pull/3527#issuecomment-510170014). I think that remains fairly unlikely but not out of the question. Sorry for the lack of communication all this time.

PR #3529, opened 2019-07-10
Add unit tests for com.google.common.hash.LittleEndianByteArray

Braavos96:
I've analysed your codebase and noticed that `com.google.common.hash.LittleEndianByteArray` is not fully tested.
I've written some tests for the methods in this class with the help of [Diffblue Cover](https://www.diffblue.com/opensource).

Hopefully, these tests will help you detect any regressions caused by future code changes. If you would find it useful to have additional tests written for this repository, I would be more than happy to look at other classes that you consider important in a subsequent PR.

### DIFF
--- a/guava-tests/test/com/google/common/hash/LittleEndianByteArrayTest.java
+++ b/guava-tests/test/com/google/common/hash/LittleEndianByteArrayTest.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright (C) 2019 The Guava Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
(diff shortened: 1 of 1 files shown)


Braavos96:
Hi, @ronshapiro  We have a corporate CLA signed. However, when we try to sign it again it fails. I have a copy of the CLA but there is no email address to send it to. Any advice? :)

cpovirk:
It looks like the email address associated with your Git commits is "35978114+Braavos96@users.noreply.github.com." Probably that's not associated with the corporate CLA? Can you rewrite your commit to use a corporate email address, and we can see if that helps?

PR #3531, opened 2019-07-16
Moe Sync

ronshapiro:

Commits:
=====
<p> Clarify that SortedSetMultimap exposes the values for a key in sorted order but does *not* necessarily expose the keys in sorted order.

09bf107763561ae37ddd99cf024d6c5604eedfff

-------

<p> Remove unneeded @SuppressWarnings from @Benchmark methods.

RELNOTES=none

3c3f7c9f7bfd997c5528a4ce7dd0c802c3aeb26f

-------

<p> Rename a method to avoid J2ObjC collisions

Prevent J2ObjC from treating `getAddSuppressed` as a getter by renaming it `generateAddSuppressed`. This silences a -Wobjc-property-implementation warning; soon, such warnings will become errors. See the referenced bug for full context.

cc5a55e487791db6ab2b7a98bcfd1be69d7a3638

### DIFF
--- a/android/guava-tests/benchmark/com/google/common/io/ByteSourceAsCharSourceReadBenchmark.java
+++ b/android/guava-tests/benchmark/com/google/common/io/ByteSourceAsCharSourceReadBenchmark.java
@@ -123,7 +123,6 @@ public class ByteSourceAsCharSourceReadBenchmark {
     data = ByteSource.wrap(string.getBytes(charset));
   }
 
-  @SuppressWarnings("GoodTime") // b/130759882
   @Benchmark
   public int timeCopy(int reps) throws IOException {
     int r = 0;
--- a/android/guava-tests/benchmark/com/google/common/io/CharStreamsCopyBenchmark.java
+++ b/android/guava-tests/benchmark/com/google/common/io/CharStreamsCopyBenchmark.java
@@ -96,7 +96,6 @@ public class CharStreamsCopyBenchmark {
     data = sb.toString();
   }
 
-  @SuppressWarnings("GoodTime") // b/130759882
   @Benchmark
   public long timeCopy(int reps) throws IOException {
     long r = 0;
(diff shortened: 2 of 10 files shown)


Youeatadick:
Hi there Ron

You’ve dealt with me previously regarding this scenario. I need to address a situation where I have been deemed a male with a cervix, specifically an endocervical  surgery that was made by a care aid via St .Pauls Hospital in Vancouver. I now refer to custom as gender but My previous Dr Malcolm Hedgecock never thought to mention this. He did however forget to add his contact info as well as when. 
Otherwise after trying to address the situation which threatens the parties involved as I have a disability and am being taken advantage off.
I have adhd, asburgers and now have been deemed clinically not legitimate by a civil court after trying to address this.
There was adage for a criminalist case but seeing as I was tackled for no reason and not allowed a phone call by the VPD , I don’t feel any of this was valid.
All of this is not being acknowledged and as a 40 year old male with a very hard time articulating and no interpreter. This situation needs to be addressed.
I will call later today. My number is 236 558 4736.
The charge was coming from an ex partner named Sean Sikorsky, which is also where my address with the VPD is listed. Personally this feels like an inheritance scenario but no one will be honest.
Please let me know if you can help.


Adam Hadrian Stone

Previously Adam Hadrian Stone Neal

236 558 4736

My previous address was 
503 -1177 Hornby Street Van BC V9R2M7
Then my partner handed me over to 14-1041 Compx Street Vancouver BC V6E1K1 and never mentioned the care aid. I’m not sure how to address the situation. People have been kind enough to work in the gay film fest to keep the peace I believe but there is research and development involved since 2015 as well.

Thank you 

Sent using the mail.com mail app

jbduncan:
@ronshapiro The comment above above looks like a very strange attempt at spam. Vote to remove it? :)

(If you agree, I'll remove this comment too to prevent confusion.)

Issue #3528, opened 2019-07-09
Missing interval method in Stopwatch


Ashwani001:
Stopwatch is often used to measure performance of code.
Many times we need to measure two things:
1. Time elapsed between sections of code
2. Total time elapsed

I would like to propose that a method be added which returns the interval (time elapsed since last call). Usage will be as such:

        Stopwatch stopwatch = Stopwatch.createStarted();
        someFunctionOne();
        LOG.debug("someFunctionOne took {}  ms", stopwatch.interval(TimeUnit.MILLISECONDS));
        someFunctionTwo();
        LOG.debug("someFunctionTwo took {}  ms", stopwatch.interval(TimeUnit.MILLISECONDS));
        ...
        someFunctionN();
        LOG.debug("someFunctionN took {}  ms", stopwatch.interval(TimeUnit.MILLISECONDS));
        LOG.debug("The whole process took {} ms", stopwatch.elapsed(TimeUnit.MILLISECONDS));

An alternative to having this method will be using two stopwatches. But the problem with this approach is having to reset and start every time after logging. 

        Stopwatch stopwatch = Stopwatch.createStarted();
        Stopwatch stopwatchTotal = Stopwatch.createStarted();
        someFunctionOne();
        LOG.debug("someFunctionOne took {}  ms", stopwatch.elapsed(TimeUnit.MILLISECONDS));
        stopwatch.reset().start();
        someFunctionTwo();
        LOG.debug("someFunctionTwo took {}  ms", stopwatch.elapsed(TimeUnit.MILLISECONDS));
        stopwatch.reset().start();
        ...
        someFunctionN();
        LOG.debug("someFunctionN took {}  ms", stopwatch.elapsed(TimeUnit.MILLISECONDS));
        LOG.debug("The whole process took {} ms", stopwatchTotal.elapsed(TimeUnit.MILLISECONDS));

This approach is is fine for short pieces of code but can get cumbersome for longer pieces. A getAndReset method will solve this but a second stopwatch instance will still be required. Moreover, I feel that an interval method will be more inline with the rest of the API as compared to a getAndReset.

[#2752](https://github.com/google/guava/issues/2752) discusses something similar (6th comment onward), but was closed. Code example was derived one present in the issue.

netdpb:
We've been thinking about adding something that lets you time a `Runnable`:

```
Stopwatch s = Stopwatch.create();
Duration duration = s.time(() -> someFunctionOne());
```

[...]

PR #3543, opened 2019-07-30
Added api to accept block of code and return result and time taken to execute. 

FuncGuy:
Hi,
@kluever @kevinb9n @ogregoire @ogregoire
I've Added API to accept block of code and return result and time taken to execute.

### DIFF
--- a/guava-tests/test/com/google/common/base/StopwatchTest.java
+++ b/guava-tests/test/com/google/common/base/StopwatchTest.java
@@ -16,6 +16,8 @@
 
 package com.google.common.base;
 
+import static com.google.common.base.Stopwatch.measure;
+import static com.google.common.truth.Truth.assertThat;
 import static java.util.concurrent.TimeUnit.MICROSECONDS;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
@@ -25,6 +27,7 @@ import com.google.common.annotations.GwtIncompatible;
 import com.google.common.testing.FakeTicker;
 import java.time.Duration;
 import junit.framework.TestCase;
+import org.junit.Test;
 
 /**
  * Unit test for {@link Stopwatch}.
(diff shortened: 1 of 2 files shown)


FuncGuy:
> @googlebot I signed it!

kluever:
I think we (also) probably want an API that only returns the elapsed `Duration`, where the result of the method invocation can be ignored or is `void`.

FuncGuy:
sorry for invalid indentation I will fix it ASAP.

ogregoire:
I don't like how the exceptions are handled.

We do not know if there was an exception or not. There is currently no difference between an exception thrown by the measured code block and an exception being returned by the same code block.

I don't know what the people would prefer: let the exception pass through and get no result of a timed stopwatch, or on the contrary measure the code block and get the caught exception in a specific returned value.

So I'd say this would require more API:

* `measure` -> `Duration` (ignore result, and let the exception through)
* `measureAndGet` -> `TimedResult` (get the result, but let the exception through)
* `measureAndCatch`? -> `ExceptionallyTimedResult`? (get the result, catch the exception, measure anyways)

Anyways, this should probably be better refined.

FuncGuy:
@kluever  @ogregoire 
I've addressed the review comments by @ogregoire if any further change is needed please mention it.

netdpb:
Looks relevant to #3528.

guava-tests/test/com/google/common/base/StopwatchTest.java
```diff
@@ -211,4 +216,71 @@ public void testToString() {
...
+        assertThat(timedResult.getTimeTaken()).isGreaterThan(Duration.ZERO);
+        assertThat(timedResult.getException()).isInstanceOf(NumberFormatException.class);
+    }
+
+    public void updateExceptionally() {
+      Integer.valueOf("");
+    }
+
+    public void update() {
+    }
```
rahilsh, line 285: why an empty method, `() -> update()` can be replaced with `() -> {}`

PR #3549, opened 2019-08-02
Add blockCancellation method to FluentFuture

njhill:
Maybe there is some reason that this wasn't originally included but it seems like a natural addition and would be handy for me at least.


njhill:
@cpovirk any interest in this? All of the unary `ListenableFuture` transformations in `Futures` have `FluentFuture` analogs apart from this one...

Issue #3555, opened 2019-08-10
Add `Ints.shuffle()`


leventov:
There are very similar arguments applying to this question as to #502 (Add `Ints.reverse()`), the difference is that shuffle algorithm is a bit less trivial than `reverse()`.

netdpb:
I don't see a lot of evidence of people needing to shuffle arrays of ints in Google's codebase. Do you see a lot of use of `Collections.shuffle({Ints,Arrays}.asList(...))` in your codebase?

leventov:
I find myself visiting [this SO question](https://stackoverflow.com/questions/1519736/random-shuffling-of-an-array) to copy the code from time to time. This definitely happened more than three times.

ben-manes:
See `ArrayUtils#shuffle(int[])` in Apache Commons Lang3.

Stupremee:
Can I implement this feature?

sagada:
Hello!
I would like to take on this task as a first challenge. 
Can I start working on it?

netdpb:
Hi @Stupremee  and @sagada. Thanks for your interest in contributing!

Please take a look at [How to Contribute](https://github.com/google/guava/wiki/HowToContribute#code-contributions). We haven't gotten to the point of deciding that this feature is worth adding to Guava, so we're definitely not ready to implement it. And often it's more work for us to accept a PR than it is for us to implement it ourselves. I'm afraid this is likely one of those cases.

PR #3556, opened 2019-08-10
Add shuffle method to Ints

Periecle:
New method for Ints due to #3555


Periecle:
@googlebot I signed it!

PR #3557, opened 2019-08-10
Add shuffle method to Ints due to #3555

Periecle:


guava/src/com/google/common/primitives/Ints.java
```diff
@@ -505,6 +506,35 @@ public static void reverse(int[] array, int fromIndex, int toIndex) {
...
+   * Collections.shuffle(Ints.asList(array).subList(fromIndex, toIndex))}, but is likely to be more efficient.
+   *
+   * @throws IndexOutOfBoundsException if {@code fromIndex < 0}, {@code toIndex > array.length}, or
+   *    {@code toIndex > fromIndex}
+   */
+  public static void shuffle(int[] array, int fromIndex, int toIndex) {
+    checkNotNull(array);
+    checkPositionIndexes(fromIndex, toIndex, array.length);
+    for (int i = fromIndex, j = toIndex - 1; i < j; i++, j--) {
+      int randomPosition = ThreadLocalRandom.current().nextInt(array.length);
```
ljacqu: - The `randomPosition` should probably be between 0 and `toIndex` instead of the array length?
- Are you using `j` only to check the condition `i < j`? It's not used in the code. In general I wonder if this really randomly shuffles all parts of the array the same or if some parts are more prone to staying in the same area...

cgdecker:
I'm closing this because there's already an issue requesting this but, as with most things in Guava, an implementation is not what we need to put this in Guava but rather a strong rationale for adding it (which typically also requires us to find sufficient evidence of the need for it within our codebase). Arguing that it should exist is best done in an issue; providing code before we've decided it's an addition we want to make doesn't really get us anywhere.

PR #3559, opened 2019-08-10
Simplify the code

NB216ers:
Unnecessary 'null' check . isActuallyE(element) method checks 'null' already.


kluever:
Thanks - fixed internally. Will be mirrored out shortly.

PR #3561, opened 2019-08-17
Improve ImmutableList#of Javadoc

wilmol:
- link to the JDK equivalent Collections#singletonList rather than Collections#singleton (which returns a Set)


wilmol:
@googlebot I signed it!

nglorioso:
Thanks! We're going to merge this internally, and it should be synced out tomorrow.

Issue #3570, opened 2019-08-20
Reuse of ImmutableSet.Builder with expected size leading to infinite loop on ImmutableSet.contains


tomasisonfire:
The following snippet produces an infinite loop inside the method ImmutableSet.contains:

```java
ImmutableSet.Builder<Object> builder = ImmutableSet.builderWithExpectedSize(6);
builder.add(0);
builder.build();
ImmutableSet<Object> subject = builder.add(1).add(2).add(3).build();
boolean contains = subject.contains(4);
```

This behavior was first observed in guava 25.1-jre and it still happens in version 28.0-jre.

Analyzing the source code, I think there is a problem in the implementation of ImmutableSet.RegularSetBuilderImpl.review(). As it decides to reduce the _hashTable_, it does not recompute the _expandTableThreshold_ nor _maxRunBeforeFallback_. From this point, I observed two consequences:

1. Given a RegularSetBuilderImpl with D distinct elements. 
After resizing the hashTable to size N, if the (N-D) distinct elements are added, the subsequent build() method would return an instance of ImmutableSet that reproduces the infinite loop on method contains.
2. In the same scenario as above, when the (N+1)th element is added to the builder, the RegularSetBuilderImpl will iterate _maxRunBeforeFallback_ times in a full hashTable, until it falls back to the JdkBackedSetBuilderImpl.

cpovirk:
Thank you for catching this!

tomasisonfire:
You are welcome :)

PR #3571, opened 2019-08-20
Fix https://github.com/google/guava/issues/3570

tomasisonfire:
Fix https://github.com/google/guava/issues/3570 by resetting expandTableThreshold and maxRunBeforeFallback after resizing the hashTable. Unit test included: testReuseBuilderReducingHashTableSizeWithPowerOfTwoTotalElements

### DIFF
--- a/guava-tests/test/com/google/common/collect/ImmutableSetTest.java
+++ b/guava-tests/test/com/google/common/collect/ImmutableSetTest.java
@@ -582,4 +582,12 @@ public class ImmutableSetTest extends AbstractImmutableSetTest {
     }
     return worstCalls;
   }
+
+  public void testReuseBuilderReducingHashTableSizeWithPowerOfTwoTotalElements() {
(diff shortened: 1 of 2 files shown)


tomasisonfire:
@googlebot I signed it!

guava-tests/test/com/google/common/collect/ImmutableSetTest.java
```diff
@@ -582,4 +582,12 @@ private static long worstCaseQueryOperations(Set<?> set, CallsCounter counter) {
     }
     return worstCalls;
   }
+
+  public void testReuseBuilderReducingHashTableSizeWithPowerOfTwoTotalElements() {
+    final ImmutableSet.Builder<Object> builder = ImmutableSet.builderWithExpectedSize(6);
```
lowasser: We don't typically include `final` as you do here, considering it redundant -- it does happen to be final, but that has no real importance to the correctness of the test.

PR #3575, opened 2019-08-23
Moe Sync

cpovirk:

Commits:
=====
<p> Adds constant for the Upgrade-Insecure-Requests header.

https://w3c.github.io/webappsec-upgrade-insecure-requests/#preference

RELNOTES=Adds constant for the Upgrade-Insecure-Requests header.

7e5aa662985c6e54396b1e751149bef17236bc27


ogregoire:
Is there any reason why the javadoc for the User-Agent field has been removed while adding the new field?

cpovirk:
Thanks, looks like a mistake. Will fix.

PR #3580, opened 2019-08-27
Simplify some code

OLPMO:
Remove some redundant 'if' statement.Besides, I use .toArray(new MyClass[0])  replace .toArray(new MyClass[myList.size()]) for improving the  efficiency  of the code.
https://stackoverflow.com/questions/174093/toarraynew-myclass0-or-toarraynew-myclassmylist-size

### DIFF
--- a/guava/src/com/google/common/cache/LocalCache.java
+++ b/guava/src/com/google/common/cache/LocalCache.java
@@ -1764,7 +1764,10 @@ class LocalCache<K, V> extends AbstractMap<K, V> implements ConcurrentMap<K, V> 
     if (expiresAfterAccess() && (now - entry.getAccessTime() >= expireAfterAccessNanos)) {
       return true;
     }
-    return expiresAfterWrite() && (now - entry.getWriteTime() >= expireAfterWriteNanos);
+    if (expiresAfterWrite() && (now - entry.getWriteTime() >= expireAfterWriteNanos)) {
+      return true;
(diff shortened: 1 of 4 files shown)


OLPMO:
@googlebot I signed it!

OLPMO:
@googlebot I signed it!

ljacqu:
-1 on the LocalCache changes, just as commented in PR #3492 this fixes what is presumably reported by an IDE at the expense of readability for no real reason.

OLPMO:
@ljacqu oh,sorry. I prefer the version that I commit because I would do the same thing in my daily coding.I can also understand that different projects have different coding styles. Do you mean that all the changes of 'if' statement is unnecessary? If you think so, I would revert them into the master problem!

OLPMO:
@ljacqu  I think that 'sum == 0' needs to be preserved because 'sum == 0' itself has the meaning of isEmpty. Positive thinking is more conducive to people's understanding than negative thinking.If you do not agree with that. I can also revert other 'if' statement to the master version.
Look forward to your reply.

ljacqu:
Sorry if there was any misunderstanding, I‘m not in any way affiliated/involved with Guava – I just posted a comment as an individual passing by :)

OLPMO:
@ljacqu  That’s all right.I am very glad that you can reply to this pr.Thanks for your suggestion

PR #3582, opened 2019-08-29
extract method from duplicate code

Rookiexu:
Extract method from duplicate code of the same class

### DIFF
--- a/guava/src/com/google/common/collect/TreeMultiset.java
+++ b/guava/src/com/google/common/collect/TreeMultiset.java
@@ -780,14 +780,7 @@ public final class TreeMultiset<E> extends AbstractSortedMultiset<E> implements 
 
         left = initLeft.setCount(comparator, e, expectedCount, newCount, result);
 
-        if (result[0] == expectedCount) {
-          if (newCount == 0 && result[0] != 0) {
-            this.distinctElements--;
-          } else if (newCount > 0 && result[0] == 0) {
-            this.distinctElements++;
-          }
-          this.totalCount += newCount - result[0];
-        }
+        calculatedQuantity(expectedCount, newCount, result);
         return rebalance();
       } else if (cmp > 0) {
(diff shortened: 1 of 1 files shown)


cgdecker:
There are a number of other similar cases in this class. In principal it would be nice to remove the duplication, but the naming such methods would be somewhat awkward (I don't think the name you've used here really describes what it's doing, for example) and even then there's other parallel almost-duplication that we'd theoretically also want to eliminate in an ideal case.

So really I feel like if we were going to do something like this it would require someone making deeper changes, but given that this is low-level code that has been working for quite a while, I don't think it's going to be a high priority for us.

guava/src/com/google/common/collect/TreeMultiset.java
```diff
@@ -801,14 +794,7 @@ public int count(Comparator<? super E> comparator, E e) {
...
-        if (result[0] == expectedCount) {
-          if (newCount == 0 && result[0] != 0) {
-            this.distinctElements--;
-          } else if (newCount > 0 && result[0] == 0) {
-            this.distinctElements++;
-          }
-          this.totalCount += newCount - result[0];
-        }
+        calculatedQuantity(expectedCount, newCount, result);
         return rebalance();
```
lowasser: It looks like instead of pulling out a method, you might as well just pull this out of the if?

Rookiexu: Thank you for your advice!

Rookiexu: I modified my code with reference to your suggestion.

Rookiexu: I modified my code based on your suggestions.

Rookiexu: @lowasser I modified my code based on your suggestions.Thanks

PR #3589, opened 2019-09-02
Allow for names that have labels starting with underscore

mwullink:
This fix allows for names that have labels starting with underscore, this is useful for parsing names that have non-final labels that start with an underscore.

multiple protocols use this naming scheme, such a DKIM, which uses _domainkey labels.
see: https://tools.ietf.org/html/rfc6376


mwullink:
@googlebot I signed it!

mwullink:
@googlebot I signed it!

PR #3600, opened 2019-09-11
Update the comment for toOptional method

dacheng:
Change the description to more clearly state that the method will throw NPE if the only element if null.

The original description made me a bit confused.

That is said, it is just my feeling, and I am not an expert of English,  please feel free to reject if the original description makes more sense.

RELNOTES=n/a


cpovirk:
Without a signed CLA, we aren't allowed to look at the patch. I'll close for now, but if you can sign the CLA (and you want to), then follow the instructions above, and let us know.

PR #3601, opened 2019-09-11
Update the comment for toOptional method

dachengy:
Change the description to more clearly state that the method will throw NPE if the only element if null.

The original description made me a bit confused. 

That is said, it is just my feeling, and I am not an expert of English,  please feel free to reject if the original description makes more sense.

RELNOTES=n/a


dachengy:
@googlebot I signed it! 

I am a googler.

cgdecker:
While I don't think the change proposed here is bad, I think the existing wording is more precise and as such I don't think there's a lot of value in making this change.

PR #3605, opened 2019-09-13
Moe Sync

cgdecker:

Commits:
=====
<p> Combine ImmediateSuccessfulFuture into ImmediateFuture.

ImmediateFuture used to have other implementations, but we switched those to extend TrustedFuture instead.

d780cd5f2b0ccc2610c039c30c04e987771dd0a5

-------

<p> Rollback previous commit.

0812745bec6e345b232154d815ad242cbb4fda37

-------

<p> Roll forward previous rollback, but without making the type final.

4c984c649e1f4665a7dbee5f568cfa25def76727

-------

<p> Fix (and in one case, suppress) low-hanging rawtypes warnings.

It looks like at least some of these may once have been necessary for the Eclipse compiler (used by GWT). In fact, I had to revert one similar change because it broke a GWT build. I'm hoping that that means that the others are safe.

cabc402dbf3805f04e19ba9d72e7f8802ab6bb75

-------

<p> Increase various tests' timeouts.

25842ed29c2c13b949836082552c5fee0cdcb0d5

### DIFF
--- a/guava-tests/test/com/google/common/util/concurrent/AtomicLongMapTest.java
+++ b/guava-tests/test/com/google/common/util/concurrent/AtomicLongMapTest.java
@@ -25,11 +25,6 @@ import com.google.common.testing.SerializableTester;
 import java.util.Map;
 import java.util.Random;
 import java.util.Set;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
-import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicLong;
 import junit.framework.TestCase;
 
 /**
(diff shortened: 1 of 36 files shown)


guava-tests/test/com/google/common/util/concurrent/AtomicLongMapTest.java
```diff
@@ -578,87 +573,4 @@ public void testSerialization() {
...
-                  sum.addAndGet(threadSum);
-                }
-              });
-    }
-
-    threadPool.shutdown();
-    assertTrue(threadPool.awaitTermination(300, TimeUnit.SECONDS));
-
-    assertEquals(sum.get(), map.get(key));
-  }
```
jbduncan: @cpovirk I'm unclear on why this test was removed, and I wondered if you could enlighten me? :thinking:

cpovirk: Hmm, was supposed to have moved to another file. Oh, I see what happened. I'll fix it. Thanks.

jbduncan: Yw. :)

PR #3607, opened 2019-09-14
simplify code

playasim:
Simplified some code for numerical comparison.

### DIFF
--- a/guava/src/com/google/common/math/PairedStatsAccumulator.java
+++ b/guava/src/com/google/common/math/PairedStatsAccumulator.java
@@ -239,9 +239,6 @@ public final class PairedStatsAccumulator {
     if (value >= 1.0) {
       return 1.0;
     }
-    if (value <= -1.0) {
-      return -1.0;
-    }
-    return value;
+    else return Math.max(value, -1.0);
   }
 }
--- a/guava/src/com/google/common/math/PairedStatsAccumulator.java
+++ b/guava/src/com/google/common/math/PairedStatsAccumulator.java
@@ -21,6 +21,7 @@ import static java.lang.Double.isNaN;
 
 import com.google.common.annotations.Beta;
 import com.google.common.annotations.GwtIncompatible;
+import com.google.common.primitives.Doubles;
 
 /**
  * A mutable object which accumulates paired double values (e.g. points on a plane) and tracks some
@@ -236,9 +237,6 @@ public final class PairedStatsAccumulator {
   }
 
   private static double ensureInUnitRange(double value) {
-    if (value >= 1.0) {
-      return 1.0;
-    }
-    else return Math.max(value, -1.0);
+    return Doubles.constrainToRange(value, -1d, 1d);
   }
 }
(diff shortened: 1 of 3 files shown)


playasim:
@googlebot I signed it!

playasim:
@googlebot I signed it!

playasim:
@googlebot I fixed it.

netdpb:
I'm not sure these changes are simpler to understand than what's there. `Math.min` and `Math.max` are better in my opinion when neither argument is a constant. Every time I look at `Math.min` or `Math.max` calls for pinning a range to a constant, I have to think a bit about what it means. And two of these changes mean that we're using simple comparison for one end of a range and min or max for the other.

Can you help me to understand why, in your opinion, these changes are helpful to readers?

Thanks!

playasim:
> I'm not sure these changes are simpler to understand than what's there. `Math.min` and `Math.max` are better in my opinion when neither argument is a constant. Every time I look at `Math.min` or `Math.max` calls for pinning a range to a constant, I have to think a bit about what it means. And two of these changes mean that we're using simple comparison for one end of a range and min or max for the other.
> 
> Can you help me to understand why, in your opinion, these changes are helpful to readers?
> 
> Thanks!

As you said, when two arguments are not constant, `Math.max` or `Math.min` is easy to understand, so the modifications to the `toSafeNanos` method make sense; for the other two methods, I think using `Math.max` method is also very readable, because it is easy to see that our intention is to get the number that is greater than 0 or -1.

guava/src/com/google/common/math/PairedStatsAccumulator.java
```diff
@@ -239,9 +239,6 @@ private static double ensureInUnitRange(double value) {
     if (value >= 1.0) {
       return 1.0;
     }
-    if (value <= -1.0) {
-      return -1.0;
-    }
-    return value;
+    else return Math.max(value, -1.0);
```
netdpb: Could this whole method use `Doubles.constrainToRange()`?

That would be a more simplifying change.

guava/src/com/google/common/util/concurrent/Monitor.java
```diff
@@ -1032,7 +1032,7 @@ private static long toSafeNanos(long time, TimeUnit unit) {
     long timeoutNanos = unit.toNanos(time);
     return (timeoutNanos <= 0L)
         ? 0L
-        : (timeoutNanos > (Long.MAX_VALUE / 4) * 3) ? (Long.MAX_VALUE / 4) * 3 : timeoutNanos;
+        : Math.min(timeoutNanos, (Long.MAX_VALUE / 4) * 3);
```
netdpb: Similarly, what about using `Longs.constrainToRange()` here?

guava/src/com/google/common/math/PairedStatsAccumulator.java
playasim: Really clean and strong change, i will replace both with `constrainToRange()`

netdpb:
OK, this looks fine. I'll copy the change internally, and it'll get mirrored out shortly.

PR #3622, opened 2019-09-26
Some code can be simplified

Scott-YuYan:
1.Anonymous can be replaced with lambda.
2.`for` loop replaceable with `foreach`.

### DIFF
--- a/guava/src/com/google/common/graph/AbstractBaseGraph.java
+++ b/guava/src/com/google/common/graph/AbstractBaseGraph.java
@@ -186,11 +186,21 @@ abstract class AbstractBaseGraph<N> implements BaseGraph<N> {
             Iterators.concat(
                 Iterators.transform(
                     graph.predecessors(node).iterator(),
-                        predecessor -> EndpointPair.ordered(predecessor, node)),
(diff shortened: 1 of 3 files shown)


JiangXi-YuYan:
@googlebot I signed it!

netdpb:
We don't use lambdas in parts of Guava that also appear in Guava-Android, in order to minimize the diff between the two flavors. But we might accept the other changes, although we don't have a process that makes it simple to merge PRs.

Scott-YuYan:
Ok,thanks for your letter and what can i do now？



---Original---

liach:
Just revert the change in `AbstractBaseGraph`. The rest is very fine

Scott-YuYan:
Ok,as you wish and i modified my commit.



---Original---

liach:
uh no you did the wrong thing, you just reverted the changes google will accept

Scott-YuYan:
Ok, I am done,very thankful.



---Original---

amalloy:
Shoot, I tried to make @Scott-YuYan the commit author but I guess I didn't assign it correctly.

PR #3631, opened 2019-10-02
Changed the string concat to append for the testGrowBuffer()

Sanskar95:
Will help in performance issues as appned is much faster in a loop.


Sanskar95:
@googlebot I signed it!

guava-tests/test/com/google/common/escape/UnicodeEscaperTest.java
```diff
@@ -78,7 +78,7 @@ public void testGrowBuffer() { // need to grow past an initial 1024 byte buffer
     StringBuilder expected = new StringBuilder();
     for (int i = 256; i < 1024; i++) {
       input.append((char) i);
-      expected.append("[" + i + "]");
+      expected.append("[").append(i.toString()).append("]");
```
liach: that `toString` call for `i` seems extraneous

cpovirk:
This is definitely a performance improvement but probably isn't worth the cost to readability, at least in a test.

PR #3633, opened 2019-10-02
Code improvement `if` statement instead of a for `statement` to a single item

ze12augusto:
Code improvement, using `if` statement instead of a for `statement` to a single item.

I also found some classes with empty methods that are not being used by the project, is this correct? If these methods can be removed, I'd like to open a pr for this.

ex: `AllEqualOrdering_CustomFieldSerializer.deserialize(SerializationStreamReader reader, AllEqualOrdering instance)`


ze12augusto:
@googlebot I signed it!

cpovirk:
Thanks.

The `CustomFieldSerializer` methods are used through reflection by [GWT](http://www.gwtproject.org) during its GWT-RPC serialization. (We're actually removing support for that at the beginning of next year, but we'll take care of it then.)

Issue #3598, opened 2019-09-08
MoreExecutors.java (w/ & w/o Android) calls the AppEngine currentRequestThreadFactory in JDK8


lesv:
AppEngine no longer requires CurrentRequestFactory to be called for Java 8.  It causes some apps to fail during local development as those API's aren't injected when running locally.

```
java.lang.RuntimeException: Couldn't invoke ThreadManager.currentRequestThreadFactory
        at com.google.common.util.concurrent.MoreExecutors.platformThreadFactory(MoreExecutors.java:759)
        at com.google.common.util.concurrent.MoreExecutors.newThread(MoreExecutors.java:799)
        at com.google.common.util.concurrent.MoreExecutors$Application.addDelayedShutdownHook(MoreExecutors.java:203)
        at com.google.common.util.concurrent.MoreExecutors$Application.getExitingScheduledExecutorService(MoreExecutors.java:189)
        at com.google.common.util.concurrent.MoreExecutors$Application.getExitingScheduledExecutorService(MoreExecutors.java:195)
        at com.google.common.util.concurrent.MoreExecutors.getExitingScheduledExecutorService(MoreExecutors.java:146)
        at com.google.cloud.sql.core.CoreSocketFactory.getDefaultExecutor(CoreSocketFactory.java:149)
```

@ludoch or @eamonnmcmanus will have more context.  This relates to GoogleCloudPlatform/cloud-sql-jdbc-socket-factory#159

The API isn't available in GAE Java 11.

Issue #3606, opened 2019-09-13
Consider deleting MoreExecutors.platformThreadFactory()


cpovirk:
It is `@Beta`, and it is likely about to become equivalent to `Executors.defaultThreadFactory()`: https://github.com/google/guava/issues/3598.

But we could keep it around, either out of inertia or out of concern that we might need it again someday for other purposes.

cpovirk:
Turns out I'm wrong: We still need to call `currentRequestThreadFactory()` in some cases.

PR #3635, opened 2019-10-02
Moe Sync

### LINKED ISSUE #3210: Quantiles.ScaleAndIndexes#computeInPlace throws an unexpected ArrayIndexOutOfBoundsException if the array indexes is empty

cpovirk:

Commits:
=====
<p> Check the array indexes in Quantiles.ScaleAndIndexes is non empty.

Fixes #3210
Fixes #3211

823a432ddaa2d274b539e12ed2bb46e5be0658ee

-------

<p> Simplify some code.

Fixes #3580

e2362def3408478a4a74d24671aef39171a1e93f

-------

<p> Document the JDK 9+ alternative to `Iterators#forEnumeration`

Fixes #3382

e432be553038b4c67f22abb8c27f82ab6d480f3d

-------

<p> Add a MIME type MPEG-DASH (mpd files)

Fixes #2677

RELNOTES=`net`: Added the `MediaType` constant for MPEG-DASH.

579ccf6c0181097be7b895ec145f04502234ef0e

-------

<p> Fix the logic that decides whether to use currentRequestThreadFactory() under App Engine.

While there, also support thread renaming under App Engine.

Fixes https://github.com/google/guava/issues/3598
Relevant to https://github.com/google/guava/issues/3606
(also, the CL in which I experimented with https://github.com/google/guava/issues/3569 before backing it out)

7bd4b2351c1483f34e7192768b4f3e7a10447c33

### DIFF
--- a/android/guava-tests/test/com/google/common/math/QuantilesTest.java
+++ b/android/guava-tests/test/com/google/common/math/QuantilesTest.java
@@ -744,4 +744,13 @@ public class QuantilesTest extends TestCase {
     } catch (IllegalArgumentException expected) {
     }
   }
+
+  public void testScale_indexes_indexes_computeInPlace_empty() {
+    int[] emptyIndexes = {};
+    try {
+      Quantiles.ScaleAndIndexes unused = Quantiles.scale(10).indexes(emptyIndexes);
+      fail("Expected IllegalArgumentException");
+    } catch (IllegalArgumentException expected) {
+    }
+  }
 }
--- a/android/guava/src/com/google/common/cache/LocalCache.java
+++ b/android/guava/src/com/google/common/cache/LocalCache.java
@@ -3815,9 +3815,7 @@ class LocalCache<K, V> extends AbstractMap<K, V> implements ConcurrentMap<K, V> 
         }
         sum -= segments[i].modCount;
       }
-      if (sum != 0L) {
-        return false;
-      }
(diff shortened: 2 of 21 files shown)


cpovirk:
All non-Googler contributions are taken from PRs.

Issue #352, opened 2014-10-31
Add functions to check blank Strings in com.google.common.base.Strings


gissuebot:
_[Original issue](https://code.google.com/p/guava-libraries/issues/detail?id=352) created by **jbaumgarten** on 2010-04-24 at 07:25 PM_

---

Would be useful to have functions to check blank Strings "à la" commons-lang in 
com.google.common.base.Strings.

Could be :
static public boolean isBlank(final String s)
static public boolean isNullOrBlank(@﻿Nullable final String s)

Could also add :
static public @﻿Nullable String blankToNull(@﻿Nullable final String s).

PR #3638, opened 2019-10-02
Add isBlank method to Strings class

slgriff:
Add `isBlank` method to Strings class


lowasser:
See https://github.com/google/guava/issues/352 for extensive discussion of this method.

slgriff:
Thank you @lowasser

PR #3639, opened 2019-10-03
Using optional.ofNullable instead of comparing with 0 and null

Sanskar95:
It shows the possibility of null as a valid response, which is often assumed that it wont be returned.


cpovirk:
We'd prefer to avoid boxing the primitive, as the current code does.

PR #3643, opened 2019-10-04
Skip file.createNewFile in Files.touch if file exists.

### LINKED ISSUE #3612: Files.touch causes unnecessary disk contention
When calling `Files.touch(foo)` on a file that already exists, Guava will call `java.io.File#createNewFile()` and assume the file already exists if the creation operation fails. But because Java guarantees the existence check and creation are atomic, the existence check lives inside a call to `java.io.FileSystem#createFileExclusively`, introducing concurrency penalties that might not be needed.

Instead, Guava should skip `file.createNewFile()` when `file.exists()` returns true.

vVv-AA:
When calling Files.touch(foo) on a file that already exists,
 java.io.File#createNewFile() will should not be called.

Fixes #3612

### DIFF
--- a/android/guava/src/com/google/common/io/Files.java
+++ b/android/guava/src/com/google/common/io/Files.java
@@ -441,8 +441,10 @@ public final class Files {
   @SuppressWarnings("GoodTime") // reading system time without TimeSource
   public static void touch(File file) throws IOException {
     checkNotNull(file);
-    if (!file.createNewFile() && !file.setLastModified(System.currentTimeMillis())) {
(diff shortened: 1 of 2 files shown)


vVv-AA:
@googlebot I signed it!

cpovirk:
It loos like the code no longer calls `setLastModified` if the file exists. Do I have that right? The contract of the method includes updating the last-modified time if the file already exists.

cpovirk:
I think this is what the failed tests are showing.

Before you change this, though, I think I am going to close #3612 altogether. Sorry that we didn't do that initially.

PR #3647, opened 2019-10-08
Add Lists.splitBy

### LINKED ISSUE #3136: Add Lists.partitionBy to split a list into a list of lists
Sometimes you wish to iterate over a list of items in a grouped fashion:

```java
List<Project> trending = Arrays.asList(new Project("guava", "java"),
        new Project("spring", "java"), new Project("node", "js"));

Project current = null;
for (Project p : trending) {
    if (current == null) {
        current = p;
    } else if (!current.getLang().equals(p.getLang())) {
        System.out.println("---");
        current = p;
    }
    System.out.println(p.getName());
}
```

There already is `Lists.partition`, but in this case the sub lists might have different sizes (and we don't know these sizes).

It would be nice to have a standard way to split the list, with only needing to specify the difference between the items:

```java
List<List<Project>> grouped = Lists.partitionBy(trending,
        (p1, p2) -> p1.getLang().equals(p2.getLang()));
for (int i = 0; i < grouped.size(); i++) {
    if (i > 0) {
        System.out.println("---");
    }
    for (Project p : grouped.get(i)) {
        System.out.println(p.getName());
    }
}
```

## Naming

There already is `Lists.partition` and as that also returns `List<List<T>>`, I would suggest a similar name. Possible alternatives would be `splitBy` or `groupBy`. My suggestion would be `partitionBy`.

## Related

- `Lists.partition`, but assumes that each sub list has the same size
- `java.util.stream.Collectors.groupingBy` (and very similar `Collectors.partitioningBy`) only expect a predicate (not a `BiPredicate`), so the caller cannot compare two items directly. They also return a `Map`, which is not desired in some cases.

## Summary

Add new method `List<List<T>> Lists.partitionBy(List<T>, BiPredicate<T, T>)`.

I would be happy to implement it and if there is agreement on this, I can create a PR.
**Thanks for reading, and if you have any questions, please ask!**

pascalgn:
Fixes #3136

### DIFF
--- a/guava-tests/test/com/google/common/collect/ListsTest.java
+++ b/guava-tests/test/com/google/common/collect/ListsTest.java
@@ -20,6 +20,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.collect.testing.IteratorFeature.UNMODIFIABLE;
 import static com.google.common.truth.Truth.assertThat;
 import static java.util.Arrays.asList;
+import static java.util.Collections.emptyList;
 import static java.util.Collections.singletonList;
 
 import com.google.common.annotations.GwtCompatible;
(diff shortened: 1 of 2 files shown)


nglorioso:
Hi there! We can't look at this PR until you sign the CLA, per `googlebot`'s  comments.

However, I think it might be better to wait for more discussion on the other issue (#3647) before advancing with the code (see https://github.com/google/guava/wiki/HowToContribute#new-features)

pascalgn:
@googlebot I signed it!

pascalgn:
@nglorioso Ah, sorry about this! I thought I could just sign the CLA before the PR will be merged, but you're right of course! Also sorry about opening this, I read the HowToContribute when I created the issue, but in the meantime forgot about it and thought having some code helps explaining the feature better!

nglorioso:
No worries, don't feel sorry for opening it, we appreciate the enthusiasm! I think there's still some discussion we're going to have about whether this feature is worth adding above and beyond things like Multimaps.index(...). I just wouldn't want you to have to spend a lot of time polishing this up with tests, documentation, etc. etc. while the original discussion hasn't gotten to a point where we'd be ready to accept it.

PR #3649, opened 2019-10-09
Make animal sniffer annotations optional

elharo:
Per https://www.mojohaus.org/animal-sniffer/animal-sniffer-annotations/ this should not be inherited by dependents

@cpovirk

### DIFF
--- a/android/guava/pom.xml
+++ b/android/guava/pom.xml
@@ -46,6 +46,7 @@
       <groupId>org.codehaus.mojo</groupId>
       <artifactId>animal-sniffer-annotations</artifactId>
       <version>${animal.sniffer.version}</version>
+      <optional>true</optional>
     </dependency>
     <!-- TODO(cpovirk): does this comment belong on the <dependency> in <profiles>? -->
     <!-- TODO(cpovirk): want this only for dependency plugin but seems not to work there? Maven runs without failure, but the resulting Javadoc is missing the hoped-for inherited text -->
--- a/guava/pom.xml
+++ b/guava/pom.xml
@@ -46,6 +46,7 @@
       <groupId>org.codehaus.mojo</groupId>
       <artifactId>animal-sniffer-annotations</artifactId>
       <version>${animal.sniffer.version}</version>
+      <optional>true</optional>
     </dependency>
     <!-- TODO(cpovirk): does this comment belong on the <dependency> in <profiles>? -->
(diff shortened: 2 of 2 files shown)


elharo:
The purpose of animal sniffer is for compile time checking of the project with animal sniffer. That's it. These annotations don't have any use post-compile.

elharo:
I also think it's important that a fundamental library such as Guava minimize its dependency tree, ideally to the JDK, full stop. However between animal sniffer, error prone, and others Guava has slowly been accumulating a lot of additional dependencies that are transitively appearing in other projects. Every dependency is a liability.

cpovirk:
They have class retention, though, so they'll be in the output. Maybe they shouldn't be, but arbitrary tools don't know that :( Maybe file a bug against Animal Sniffer?

elharo:
Good idea. Done: https://github.com/mojohaus/animal-sniffer/issues/73

We'll see what the animal sniffer devs say, but I'm guessing the plugin is analyzing the .class files rather than the .java files, which would explain why they want class retention with an optional dependency. The linkage checker operates similarly. We read .class files, not .java files.

cpovirk:
Ah, you're probably right.

I found the following in [some docs](http://www.mojohaus.org/animal-sniffer/animal-sniffer-annotations/index.html):

> Animal sniffer also allows you to specify your own custom annotation if needed. Check the configuration of the relevant task for more information.

So maybe we can provide our own.

cpovirk:
http://www.mojohaus.org/animal-sniffer/animal-sniffer-maven-plugin/check-mojo.html#annotations

cpovirk:
I think that the impact of annotation-only packages is fairly small as liabilities go. Or at the very least, if it's safe to omit the packages entirely, then it's likely to be safe to have the wrong version present.

Still, I think it would be fine to migrate our one usage of `@IgnoreJRERequirement` to a custom, package-private implementation, configuring Animal Sniffer to recognize that, even if that feels like primarily a symbolic victory to me.

elharo:
Should that approach (moving to a package private custom implementation) start in google3 or here?

cpovirk:
google3 is marginally easier for us, but I'm fine with whatever you prefer.

elharo:
It seems the pom.xml only lives here so probably easier to work from github.

cpovirk:
A search like `f:guava/pom.xml` should turn it up in `google3`, but again, here is fine if that's easier.

PR #3652, opened 2019-10-10
switch to an internal annotation for ignoring JRE requirements

elharo:
@cpovirk This completely removes the dependency on animal-sniffer annotations


cpovirk:
(Thanks!)

PR #3663, opened 2019-10-18
fix #3662: ByteStreams.fixUpTo avoids unnecessary large allocations

### LINKED ISSUE #3662: ByteStreams.skipUpTo makes large, unnecessary, allocations
skipUpTo helpfully falls back from skip invocations which return zero to invoke a read into a buffer.

The current implementation has two inefficiencies which cause unexpected performance degradation in tight loops, especially when a small number of bytes are requested:

1. In many cases skipSafely is able to skip all the requested bytes, there's no need to allocate a read buffer
2. The caller only requests a few bytes are skipped. Calling `ByteStreams.skipFully(stream, 32)` always allocates a 8192 byte buffer.

Proposed solution:
Defer buffer allocation until it is needed, using the minimum of the requested size and BUFFER_SIZE. This reduces allocation pressure in all cases of small skip values, as well as streams that report nonzero bytes available (assuming skip returns a positive value).

carterkozak:
The fallback read buffer is allocated only when fallback reads
are used. The buffer size is bounded by the remaining bytes
to skip, allowing small skip requests which do require a fallback
read to avoid an 8k buffer allocation.

### DIFF
--- a/guava/src/com/google/common/io/ByteStreams.java
+++ b/guava/src/com/google/common/io/ByteStreams.java
@@ -816,7 +816,8 @@ public final class ByteStreams {
    */
   static long skipUpTo(InputStream in, final long n) throws IOException {
     long totalSkipped = 0;
-    byte[] buf = createBuffer();
+    // A buffer is allocated if skipSafely does not skip any bytes.
+    byte[] buf = null;
 
     while (totalSkipped < n) {
       long remaining = n - totalSkipped;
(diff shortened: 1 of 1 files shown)


carterkozak:
@googlebot I signed it!

guava/src/com/google/common/io/ByteStreams.java
```diff
@@ -816,7 +816,8 @@ public static void skipFully(InputStream in, long n) throws IOException {
    */
   static long skipUpTo(InputStream in, final long n) throws IOException {
     long totalSkipped = 0;
-    byte[] buf = createBuffer();
+    // A buffer is allocated if skipSafely does not skip any bytes.
+    byte[] buf = null;
```
carterkozak: It may be worth considering a singleton skip-buffer instead of allocation in this case. Using that model, it's possible to write a stream to snoop on other skip calls sharing the JVM which may be reading sensitive data, so I opted for an approach more similar to the original implementation.

cgdecker: Yes, that's exactly why we aren't using a singleton buffer for this.

cgdecker:
This has been submitted internally and will be synced out soon.

PR #3666, opened 2019-10-22
Moe Sync

cgdecker:

Commits:
=====
<p> Fix ImmutableList#of Javadoc

Fixes #3561

a14d317807e63770c5e57a4307cd1e69ea233ba0

-------

<p> Attempt to fix inheriting Javadoc from the JDK again.

It's currently missing:
https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/primitives/UnsignedInteger.html#equals-java.lang.Object-

But I haven't managed to reproduce the problem locally.

I'm trying this on the theory that java.home might be pointing to the JDK (the top-level directory in which src.zip lives) rather than the JRE (a subdirectory).

Note that JDK11 has src.zip in a lib/ subdirectory. I tried adding it to a Java 11 build of Guava, but things blew up because the JDK contains module declarations, which naturally are incompatible with our -source 8. Possibly the right fix there is to avoid -source 8 in favor of the approach I used for jimfs in CL 272937179 -- except that maybe I already tried that and it wasn't sufficient? I'm happy to punt on that for now, especially since I just tried removing -source 8 and got a bunch of "too many module declarations found" errors :)

4c83832db7e6ae8d0cd351fac7fe28f4116ef0de

### DIFF
--- a/android/guava/pom.xml
+++ b/android/guava/pom.xml
@@ -173,7 +173,7 @@
   </build>
   <profiles>
     <profile>
-      <id>srczip</id>
+      <id>srczip-parent</id>
       <activation>
         <file>
           <exists>${java.home}/../src.zip</exists>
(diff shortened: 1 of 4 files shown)


cgdecker:
Verifying CLA because the commit comes from PR #3561.

PR #3671, opened 2019-10-25
Upgrade checker-qual to v2.11.1

adriangonz:
The `checker-qual` version that `guava` is using (`v2.8.1`) is a few releases behind. It would be good to upgrade to the latest version `v2.11.1`. There doesn't seem to be any breaking changes on the [changelog](https://github.com/typetools/checker-framework/releases) and tests are passing after the change.

## Changelog

- Upgrade `org.checkerframework:checker-qual` from v2.8.1 to v2.11.1.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -247,7 +247,7 @@
       <dependency>
         <groupId>org.checkerframework</groupId>
         <artifactId>checker-qual</artifactId>
-        <version>2.10.0</version>
+        <version>2.11.1</version>
       </dependency>
       <dependency>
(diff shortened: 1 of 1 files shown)


adriangonz:
@googlebot I signed it!

adriangonz:
Thanks for that clarification @cgdecker! We are currently processing this so that I can get added to the company's CLA whitelist.

cgdecker:
Got this submitted internally and merged out in #3769. Thanks!

PR #3673, opened 2019-10-25
Update 'adding Guava to your build using Gradle' snippet

jjohannes:
Hello from Gradle 👋 

I noticed that the 'adding Guava to your build using Gradle' section in the README could use an update. So here it is.

The notation with brackets - () - and double quotes - " - is valid in both Gradle's Groovy and Kotlin DSL. The keyword (the 'configuration') determines the scopes in which  the dependency is visible. "compile" is already discouraged since some time and officially deprecated with Gradle 6. "implementation" and "api" are both possible, but the initial recommendation for using a
library should be "implementation".

https://docs.gradle.org/6.0-rc-1/userguide/java_library_plugin.html#sec:java_library_recognizing_dependencies

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,11 +40,24 @@ To add a dependency using Gradle:
 
 ```gradle
 dependencies {
-  compile 'com.google.guava:guava:28.1-jre'
-  // or, for Android:
-  api 'com.google.guava:guava:28.1-android'
+  // Pick one:
+
+  // 1. Use Guava in your implementation only:
+  implementation("com.google.guava:guava:28.1-jre")
+
+  // 2. Use Guava types in your public API:
+  api("com.google.guava:guava:28.1-jre")
+
+  // 3. Android - Use Guava in your implementation only:
+  implementation("com.google.guava:guava:28.1-android")
+
+  // 4. Android - Use Guava types in your public API:
+  api("com.google.guava:guava:28.1-android")
 }
 ```
+
+For more information on when to use `api` and when to use `implementation`, consult the
+[Gradle documentation on API and implementation separation](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation).
 
 ## Snapshots
 


cpovirk:
Thanks! It sounds like we should definitely be moving off `compile` at this point.

In the past, though, we've heard people (@tbroyer, @JakeWharton) suggest that `api` makes a safer default than `implementation`:
- https://github.com/google/guava/pull/3154
- https://github.com/google/auto/issues/620

(The Auto discussion touches on [especially controversial issues around annotations](https://github.com/google/auto/pull/765#discussion_r330655031), which might be a distraction from the more common case -- though Guava does include annotations, too!)

Even if we do give `api` as a "default" recommendation, I can certainly see _mentioning_ `implementation`, ideally by saying something like:
> If you can make _certain guarantees_ \[link\], then you can change `api` to `implementation` to let Gradle skip work in some cases.

...

OK, and now I actually went and read the docs you linked to :) (Thanks!) I see how `implementation` has its advantages beyond build speed.

I'll revise what I said above to something more like "It's complicated" :) I'm now not sure it makes sense to give _either_ recommendation as a "default."

I would be tempted to put:

> <pre>
> dependencies {
>   <i>some-configuration</i>("com.google.guava:guava:28.1-jre")
>   // or, for Android:
>   <i>some-configuration</i>("com.google.guava:guava:28.1-android")
> }
> </pre>
> ...where <i>some-configuration</i> is either `implementation` or `api`, as discussed [in the Gradle docs](https://docs.gradle.org/6.0-rc-1/userguide/java_library_plugin.html#sec:java_library_recognizing_dependencies).

Thoughts on that?

jjohannes:
Thanks for the quick response @cpovirk. I wasn't aware that there is such a history in discussing this little snippet. :) And of course it depends a lot on whom you ask what the more sensible default is. From our side, we prefer recommending "implementation" because the user will have to make the explicit decision to expose it, if they are developing a library/module used by others. But I can also understand the "some users just want it to work as before/as in Maven" argument.

The downside of the "some-configuration" would be that it looses it's nice "copy&paste" character. Also in theory it makes sense as there are even more configurations you can use.

How about adding both?
```
dependencies {
  // Use Guava in your implementation
  implementation("com.google.guava:guava:28.1-jre")
  // Use Guava types in your public API
  api("com.google.guava:guava:28.1-jre")
  // Android: Use Guava in your implementation
  implementation("com.google.guava:guava:28.1-android")
  // Android: Use Guava types in your public API
  api("com.google.guava:guava:28.1-android")
}
```
 The comments are vastly simplifying the effect, but give a rough guidance for users who absolutely have no idea.

cpovirk:
I am encouraged enough by tbroyer's thumbs-up to move on to tweaking wording :)

> ```gradle
> dependencies {
>   // Use Guava in your implementation only:
>   implementation("com.google.guava:guava:28.1-jre")
>   // OR: Use Guava types in your public API:
>   api("com.google.guava:guava:28.1-jre")
> ```
> 
> Or, for Android:
> 
> ```gradle
> dependencies {
>   // Use Guava in your implementation only:
>   implementation("com.google.guava:guava:28.1-android")
>   // OR: Use Guava types in your public API:
>   api("com.google.guava:guava:28.1-android")
> }
> ```

But maybe that separates the Android and non-Android cases too widely? Plus, we'd probably want to do the same for Maven, which is more verbose, so we'd end up with 4 code blocks, some of them large :\

> ```gradle
> dependencies {
>   // Use Guava in your implementation only:
>   implementation("com.google.guava:guava:28.1-jre")
>   // OR: Use Guava types in your public API:
>   api("com.google.guava:guava:28.1-jre")
> }
>
> // Or, for Android:
> 
> dependencies {
>   // Use Guava in your implementation only:
>   implementation("com.google.guava:guava:28.1-android")
>   // OR: Use Guava types in your public API:
>   api("com.google.guava:guava:28.1-android")
> }
> ```

That would mean only 2 code blocks (Gradle and Maven), and maybe we can get away with keeping the Maven one as it is?

I don't know, or:

> ```gradle
> dependencies {
>   // Pick one:
>
>   // 1. Use Guava in your implementation only:
>   implementation("com.google.guava:guava:28.1-jre")
>
>   // 2. Use Guava types in your public API:
>   api("com.google.guava:guava:28.1-jre")
>
>   // 3. Android - Use Guava in your implementation only:
>   implementation("com.google.guava:guava:28.1-android")
>
>   // 4. Android - Use Guava types in your public API:
>   api("com.google.guava:guava:28.1-android")
> }
> ```

Maybe I like that one?

(Whatever we do, we should link to the nice Gradle docs.)

jjohannes:
I also like the last suggestion. With the newlines in between it's more readable than what I suggested. I think it's good to keep having one block for Maven and one for Gradle. Then it's easier to spot that each is about one tool.

Below the listing we can add a sentences like:

For more information on when to use `api` and when to use `implementation`, consult the [Gradle documentation on  API and implementation separation](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_separation)

cpovirk:
Sounds good to me. Want to update the PR to that, and I can import it?

jjohannes:
Done! Thanks @cpovirk!

PR #3674, opened 2019-10-28
Moe Sync

cgdecker:

Commits:
=====
<p> Rollback

*** Reason for rollback ***

Didn't work: https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/primitives/UnsignedInteger.html#equals-java.lang.Object-

Attempt to fix inheriting Javadoc from the JDK again.

It's currently missing:
https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/primitives/UnsignedInteger.html#equals-java.lang.Object-

But I haven't managed to reproduce the problem locally.

I'm trying this on the theory that java.home might be pointing to the JDK (the top-level directory in which src.zip lives) rather than the JRE (a subdirectory).

Note that JDK11 has src.zip in a lib/ subdirectory. I tried adding it to a Ja...

***

5b6537baac6ae0569b9411e578c8d37f75cc716a

-------

<p> Update 'adding Guava to your build using Gradle' snippet

The notation with brackets - () - and double quotes - " - is valid
in both Gradle's Groovy and Kotlin DSL.

The "compile" configuration is already discouraged since some time and
officially deprecated with Gradle 6.

Fixes #3673

f4866c6d50df1551a75986b1bb5d34b9245634d1

-------

<p> Format README.md.

c850b5742a68be52b5d08a7fdf544627088beb4d

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,11 +40,25 @@ To add a dependency using Gradle:
 
 ```gradle
 dependencies {
-  compile 'com.google.guava:guava:28.1-jre'
-  // or, for Android:
-  api 'com.google.guava:guava:28.1-android'
+  // Pick one:
+
+  // 1. Use Guava in your implementation only:
+  implementation("com.google.guava:guava:28.1-jre")
+
+  // 2. Use Guava types in your public API:
+  api("com.google.guava:guava:28.1-jre")
+
+  // 3. Android - Use Guava in your implementation only:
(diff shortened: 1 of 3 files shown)


cgdecker:
Manually setting `cla: yes` because the commit in question comes directly from the author's PR #3673.

PR #3676, opened 2019-10-30
Precautions for null pointer exception

Sangaibisi:


Sangaibisi:
@googlebot I signed it!

ogregoire:
I believe that the `NullPointerException` is intended: the package is annotated with `@ParametersAreNonNullByDefaut` so I actually would expect a `NullPointerException` instead of the `IllegalArgumentException` that would be thrown instead.

netdpb:
The package `com.google.common.base.internal` isn't annotated with `@ParametersAreNonNullByDefault`, but if anything we'd prefer to throw `NullPointerException` in that case anyway.

This method is called (reflectively) only in [`FinalizableReferenceQueue`](https://github.com/google/guava/blob/master/guava/src/com/google/common/base/FinalizableReferenceQueue.java), and with a class literal for that argument.

Thank you for this PR, but I don't think it's necessary.

PR #3677, opened 2019-10-31
MOE Sync 2019-10-31

netdpb:

Commits:
=====
<p> Fix toString exception handling change from 957c37116fb2cc52b2f8b405871b65a84917f2ca.

This is not an ideal code fix but it restores the previous functionality and adds a test that verifies the exception handling logic.

82b7589dfbfc5f2b6f68d6a6a1bb34e4633f4f3c

-------

<p> Handle StackOverflowError from both pendingToString and appendUserObject.

Adds a bunch of tests to validate error handling in toString code
- Verify that a future completing during the toString call results in a done formatted string
- Verify that an exception thrown by pendingToString doesn't cause toString to fail
- Verify that cycles don't cause toString to fail
- Verify that deep chains of SetFuture don't cause toString to fail

RELNOTES=Catch StackOverflowError in AbstractFuture.toString to prevent long chains of futures from failing toString calls.

29981d416d6b4214e65360ad3d932ba0881a4c8f

-------

<p> Update Documentation

Fixes #3507

8a0d6bb345341fc0ba91781070da5b522d4b714b

### DIFF
--- a/android/guava-tests/test/com/google/common/util/concurrent/AbstractFutureTest.java
+++ b/android/guava-tests/test/com/google/common/util/concurrent/AbstractFutureTest.java
@@ -895,6 +895,21 @@ public class AbstractFutureTest extends TestCase {
     assertThat(orig.toString()).contains("[status=SUCCESS, result=[this future]]");
   }
 
+  public void testSetFutureSelf_toStringException() {
+    SettableFuture<String> orig = SettableFuture.create();
+    orig.setFuture(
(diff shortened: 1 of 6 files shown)


netdpb:
@gokaco Can you please leave a comment that contains only `@googlebot I consent.`

netdpb:
Adding CLA label because 8a0d6bb345341fc0ba91781070da5b522d4b714b came from @gokaco's PR #3507.

PR #3534, opened 2019-07-24
Upgrade maven jar plugin from 3.0.2 to 3.2.0 - Reproducible Builds

WilliamCollishaw:
Upgrade maven jar plugin from 3.0.2 to the latest [3.2.0](https://blogs.apache.org/maven/entry/apache-maven-jar-plugin-version3)
Maven source and jar plugins both introduced [build reproducibility](https://maven.apache.org/guides/mini/guide-reproducible-builds.html) (Maven source plugin upgrade in #3681) in 3.2.0

Versions bypassed:

- [3.1.0](https://blogs.apache.org/maven/entry/apache-maven-jar-plugin-version)
- [3.1.1](https://blogs.apache.org/maven/entry/apache-maven-jar-plugin-version1)
- [3.1.2](https://blogs.apache.org/maven/entry/apache-maven-jar-plugin-version2)

PR #3681, opened 2019-11-06
Upgrade maven source plugin from 3.1.0 to 3.2.0 - Reproducible Builds

WilliamCollishaw:
Upgrade maven source plugin from 3.1.0 to [3.2.0](https://maven.apache.org/guides/mini/guide-reproducible-builds.html) 
Maven source and jar plugins both introduced [build reproducibility](https://maven.apache.org/guides/mini/guide-reproducible-builds.html) (Maven jar plugin upgrade in #3534) in 3.2.0


WilliamCollishaw:
Closed in #3686

Issue #3688, opened 2019-11-07
Suppliers.memoizeWithExpiration should take a duration supplier function


chriskilding:
The `Suppliers.memoizeWithExpiration` function currently takes a fixed duration as an argument. I have a use case in a Jenkins plugin where the memoized Supplier’s cache duration needs to be user-configurable, in the Jenkins global configuration object.

It would be particularly helpful if there was a variant of this function that accepted a duration supplier function. Something like...

```java
memoizeWithExpiration(Supplier<T> delegate, Supplier<Duration> duration)
```

Which in my use case would let me do something like...

```java
var credentialSupplier =
    Suppliers.memoizeWithExpiration(<delegate>, () -> PluginConfiguration.getInstance().getCacheDuration());
```

It would be nice if the duration supplier could return a `java.time.Duration`, but I am aware that issue #3169 is not yet resolved. If this is not possible then a custom Guava-specific substitute would also work for my use case.

PR #3691, opened 2019-11-08
Add Duration and Supplier<Duration> overloads for Supplier#memoizeWithExpiration

mfboulos:
Directly addresses #3688 

I don't want to step on any toes since I saw that `long, TimeUnit` overloads are in the cards for several classes, but I also see the use in a `Duration` supplier, so I went and wrote this up real quick so we can at least have it in the works.

One thing I noticed while writing this is that `com.google.common.util.concurrent.Internal#toNanosSaturated` is package-scoped internal, but the only method in there is applicable to all legacy API's that require a `long, TimeUnit`, and there are several scattered in `base`, `cache`, and `collect` under `com.google.common`. It might be worth considering gauging the need for it to public-scoped, if at some point the human race figures out how to prolong life indefinitely.

Also first contribution here, hope all looks good :)

### DIFF
--- a/guava/src/com/google/common/base/Suppliers.java
+++ b/guava/src/com/google/common/base/Suppliers.java
@@ -20,6 +20,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import com.google.common.annotations.GwtCompatible;
 import com.google.common.annotations.VisibleForTesting;
 import java.io.Serializable;
+import java.time.Duration;
 import java.util.concurrent.TimeUnit;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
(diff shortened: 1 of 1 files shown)


mfboulos:
@googlebot I signed it!

jbduncan:
Hi @mfboulos, welcome to Guava. :)

I think it's worth mentioning that it's unlikely this PR will be merged, because the Guava team prefer to do API reviews internally (which is apparently the hard part) and then write the code themselves to match whatever API they've agreed on.

See the [contribution guidelines](https://github.com/google/guava/blob/master/CONTRIBUTING.md) for more information.

jbduncan:
That being said, I'm sure that this code would be a great start for @chriskilding's needs, who opened #3688, if you're happy for them to take the code as-is. :)

kluever:
> It might be worth considering gauging the need for it to public-scoped

FWIW, we have a copy of this in our (still internal) `com.google.common.time` package, which we hope to get around to releasing at some point.

mfboulos:
Understandable! My intuition was to put this PR in there just to have the code handy for you guys, but I understand there's more discussion behind it since it's essentially an extension to the existing API.

cpovirk:
We added the `Duration` overload recently, including [under Android](https://guava.dev/releases/snapshot-android/api/docs/com/google/common/base/Suppliers.html#memoizeWithExpiration(com.google.common.base.Supplier,java.time.Duration)). I don't know that we'll end up adding a `Supplier<Duration>` overload at all, but we can come back to this PR if we someday resume the discussion on https://github.com/google/guava/issues/3688 and decide to go for it.

PR #3694, opened 2019-11-09
Removed some deprecated methods in FilesTest and replaced with suggested. 

playasim:
I removed some deprecated methods in FilesTest.java and replaced them with suggested


kluever:
These tests are intended to test the deprecated APIs and they should stay as-is.

PR #3709, opened 2019-11-18
WIP +eventbus Add EventDispatchTask to help retrieve the original event.

hepin1989:
I need this to retrieve the original event


hepin1989:
@googlebot I signed it!

nglorioso:
Hi there! We're unlikely to add features to EventBus anymore. 

I note that the subscribers receive the event directly, so you may be able to log or deal with the event without having the executor that runs the event subscriber callback be aware of the event being called-back.

hepin1989:
@nglorioso Are you sure? If I am using an AsyncEvent bus, and the executor is using abort policy, then how can the subscriber know?

hepin1989:
@nglorioso Ping

PR #3713, opened 2019-11-19
Rejuvenate log levels

saledouble:
We appreciate your previous feedback and it's very helpful! Here's a reissue of https://github.com/google/guava/pull/3435 with a new version of our tool. The tool made many fewer transformations. Again, we'd appreciate any feedback and are willing to make further changes if you wish to incorporate our PR into your project.

Settings
---
We have several analysis settings. We can vary these settings and rerun if you desire. The settings we are using in this pull request are:

Treat CONFIG/WARNING/SEVERE levels as a category and not a traditional level, i.e., our tool ignores these log levels.
Never lower the logging level of logging statements within catch blocks.
Never lower the logging level of logging statements within if statements.
Never lower the logging level of logging statements containing certain (important) keywords.
Never raise the logging level of logging statements without particular keywords in their messages.
Avoid log wrapping by disregarding logging statements contained in if statements mentioning log levels.
The greatest number of commits from HEAD evaluated: 1000.
The head at the time of analysis was: 71de4064ac005e1d283b8a6d4153aab834903720

### DIFF
--- a/guava-testlib/src/com/google/common/collect/testing/PerCollectionSizeTestSuiteBuilder.java
+++ b/guava-testlib/src/com/google/common/collect/testing/PerCollectionSizeTestSuiteBuilder.java
@@ -61,7 +61,7 @@ public abstract class PerCollectionSizeTestSuiteBuilder<
     Set<Feature<?>> features = Helpers.copyToSet(getFeatures());
     List<Class<? extends AbstractTester>> testers = getTesters();
 
-    logger.fine(" Testing: " + name);
+    logger.finer(" Testing: " + name);
 
     // Split out all the specified sizes.
     Set<Feature<?>> sizesToTest = Helpers.<Feature<?>>copyToSet(CollectionSize.values());
@@ -72,7 +72,7 @@ public abstract class PerCollectionSizeTestSuiteBuilder<
     sizesToTest.retainAll(
         Arrays.asList(CollectionSize.ZERO, CollectionSize.ONE, CollectionSize.SEVERAL));
 
-    logger.fine("   Sizes: " + formatFeatureSet(sizesToTest));
+    logger.finer("   Sizes: " + formatFeatureSet(sizesToTest));
 
     if (sizesToTest.isEmpty()) {
       throw new IllegalStateException(
--- a/guava-testlib/src/com/google/common/collect/testing/FeatureSpecificTestSuiteBuilder.java
+++ b/guava-testlib/src/com/google/common/collect/testing/FeatureSpecificTestSuiteBuilder.java
@@ -173,12 +173,12 @@ public abstract class FeatureSpecificTestSuiteBuilder<
   public TestSuite createTestSuite() {
     checkCanCreate();
 
-    logger.fine(" Testing: " + name);
-    logger.fine("Features: " + formatFeatureSet(features));
+    logger.finest(" Testing: " + name);
+    logger.finest("Features: " + formatFeatureSet(features));
 
     FeatureUtil.addImpliedFeatures(features);
 
-    logger.fine("Expanded: " + formatFeatureSet(features));
+    logger.finest("Expanded: " + formatFeatureSet(features));
 
     // Class parameters must be raw.
     List<Class<? extends AbstractTester>> testers = getTesters();


guava-testlib/src/com/google/common/collect/testing/PerCollectionSizeTestSuiteBuilder.java
```diff
@@ -61,7 +61,7 @@ public TestSuite createTestSuite() {
     Set<Feature<?>> features = Helpers.copyToSet(getFeatures());
     List<Class<? extends AbstractTester>> testers = getTesters();
 
-    logger.fine(" Testing: " + name);
+    logger.finer(" Testing: " + name);
```
netdpb: This method is logging roughly the same information as the method it's overriding, but your tool suggested "finer" here and "finest" in the overridden method. How come?

More generally, is there a way to describe this PR meaningfully, rather than just as the result of running a tool? For instance, what benefit will Guava developers or Guava users get from this particular change?

As before, I'm curious whether this tool is designed to be appropriate for stable shared libraries (and even more, for _testing_ libraries like these) — or is it more appropriate for application code?

saledouble:
The CSV files: [CSV_files.tar.gz](https://github.com/google/guava/files/3870255/CSV_files.tar.gz)

guava-testlib/src/com/google/common/collect/testing/PerCollectionSizeTestSuiteBuilder.java
khatchad: @netdpb, thank you for your response. Please see inline below:

> This method is logging roughly the same information as the method it's overriding,

Yes, we tweaked the approach based on some feedback we received from various PRs.

> but your tool suggested "finer" here and "finest" in the overridden method. How come?

The tool deemed the surrounding code of these statements to be not as "interesting" as other statements as they may have not be edited recently. However, this PR includes a setting that only adjusts INFO and below. We will issue another one that does the full range of levels as well.

> More generally, is there a way to describe this PR meaningfully, rather than just as the result of running a tool? For instance, what benefit will Guava developers or Guava users get from this particular change?

Yes. We believe that there is a certain class of logging statements that are correlated with a project's features. Over time, certain features become more "interesting" than others, e.g., ones that are being enhanced or require bug fixes, and vice-versa. As such, we would like to bring logging statements associated with more active features to the forefront of developers' attention and likewise push statements not associated with active features to the background.

In this specific case, the tool decreased the level of logging statements in test setup code. This could allow statements pertaining to actual test cases (non-boiler plate) to be displayed more prominently.

> As before, I'm curious whether this tool is designed to be appropriate for stable shared libraries (and even more, for _testing_ libraries like these) — or is it more appropriate for application code?https://github.com/google/guava/pull/3435#issuecomment-480320723

Yes, we did consider [your comment from before](https://github.com/google/guava/pull/3435#issuecomment-480320723) (thank you again). We think that this approach still works for library/framework code as it evolves (i.e., developers may want to see logging statements more specific to their own tests rather than test setup).

All that being said, @saledouble is currently running the tool with different settings as mentioned before. We'll post that momentarily. Thank you again for your consideration!

khatchad: Alternate PR: https://github.com/google/guava/pull/3716. It has all of these transformations as well as some new ones that are related to exceptions.

netdpb: >> but your tool suggested "finer" here and "finest" in the overridden method. How come?
>
> The tool deemed the surrounding code of these statements to be not as "interesting" as other statements as they may have not be edited recently. However, this PR includes a setting that only adjusts INFO and below. We will issue another one that does the full range of levels as well.

My comment was about the fact that your tool changed "fine" to "finer" in one case and "finest" in what looks like an identical case, and in fact one case is in a method that overrides the other. How does that help the reader of this code or the person running the test?

khatchad:
You're correct. Parts of the library that are important to library developers may not match the interest level of application developers using the library. This seems to touch on a broader issue of log "composition," i.e., the log intentions of application developers may not coincide with the application's dependencies in general.

khatchad:
I think that this problem will be more apparent in library components that are exposed to application developers.

guava-testlib/src/com/google/common/collect/testing/PerCollectionSizeTestSuiteBuilder.java
khatchad: Sorry, I misread your original question. The DOI model is supposed to propagate interest across hierarchies. That is, if `M` is "interesting" and `M'` overrides `M`, then `M'` is also "interesting" as vice-versa. But, the *degree* of "interestingness" may not be consistent, i.e., `M'` is "interesting" but not necessarily to the degree of `M`.

In fact, I bet the overriding method `createTestSuite()` in this case was the more interesting one and that internestness propagated to overridden method. We'll take a look to be sure.

In general, it seems we have a kind of behavioral subtyping problem here. There should be consistency with the recommendations of overriding methods. I've filed https://github.com/ponder-lab/Logging-Level-Evolution-Plugin/issues/250. Thank you for your feedback.

khatchad: This is what I've found. My hypothesis about the overriding DOI propagation seems plausible but we haven't recorded whether DOI is increased due to a direct code manipulation in the git history or through relationship propagation. We'l need to do some further investigation, but I do think that the issue I filled is relevant either way:

type FQN | enclosing method | DOI value
-- | -- | --
com.google.common.collect.testing.FeatureSpecificTestSuiteBuilder | createTestSuite() | 0.6066328
com.google.common.collect.testing.PerCollectionSizeTestSuiteBuilder | createTestSuite() | 8.482215

netdpb:
Thanks for showing us this tool and its results. I'm going to close this PR (and the other one), as I don't think it's generating useful changes for our library. Best of luck on developing your tools in the future.

khatchad:
Thanks for your feedback @netdpb!

Issue #2730, opened 2017-01-31
RateLimiter.create called with zero warmup fails with real clock


greggdonovan:
We recently hit an issue where calling `RateLimiter.create(16.0, 0, TimeUnit.SECONDS)` failed to rate limit after an upgrade from Guava 17 to 21. I tried to reproduce this using the `FakeStopwatch` but could only get it to reproduce with a real clock.

I'll be happy to submit this as PR if the change below is useful.

Bisecting tells me this changed in [this commit](https://github.com/google/guava/commit/0b4e7e5df7f985026f731f10f2c37a44543a5355).

`bisect.sh`
```bash
#!/usr/bin/env bash

function back() {
  cd ~/development/guava
  git co .
}
trap back EXIT

JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk1.7.0_80.jdk/Contents/Home

# I can't get guava-gwt to compile :-(
sed -i '' 's_<module>guava-gwt</module>__' pom.xml

mvn clean install -Dmaven.test.skip=true || exit 125
cd guava-tests
{ mvn test -Dtest=com.google.common.util.concurrent.RateLimiterZeroWarmupTest && exit 0; } || exit 127
```

`guava-tests/test/com/google/common/util/concurrent/RateLimiterZeroWarmupTest.java`
```java
package com.google.common.util.concurrent;

import java.util.concurrent.TimeUnit;
import junit.framework.TestCase;

public class RateLimiterZeroWarmupTest extends TestCase {

  public void testWarmUpWithZeroWarmupAndRealClock() {
    RateLimiter limiter = RateLimiter.create(16.0, 0, TimeUnit.SECONDS);
    int totalPermits = 1000;
    double totalSleep = 0;
    for (int i = 0; i < totalPermits; i++) {
      totalSleep += limiter.acquire();
    }
    assertFalse("totalSleep should be more than 60s, not 0", totalSleep == 0.0d); // <-- this fails
  }
}
```

Using the `FakeStopwatch` fails to reproduce:
```
diff --git a/guava-tests/test/com/google/common/util/concurrent/RateLimiterTest.java b/guava-tests/test/com/google/common/util/concurrent/RateLimiterTest.java
index b143bd94a..0c4de6c86 100644
--- a/guava-tests/test/com/google/common/util/concurrent/RateLimiterTest.java
+++ b/guava-tests/test/com/google/common/util/concurrent/RateLimiterTest.java
[...]

PR #3724, opened 2019-11-28
Do not allow warmupPeriod set as zero

liufuyang:
In order to solve issues like https://github.com/google/guava/issues/2730

Or simply see this issue by running following code as a test:
```
@Test
  public void t() {
   RateLimiter rateLimiter =
        RateLimiter.create(1, 0, TimeUnit.SECONDS);
    Semaphore semaphore =
        new Semaphore(4);

    while (true) {
      rateLimiter.acquire();
      try {
        semaphore.acquire();
        CompletableFuture.completedStage("Done")
            .thenApply(s -> {System.out.println(s); return s;})
            .whenComplete(
                (result, throwable) -> {
                  semaphore.release();
                });
      } catch (InterruptedException e) {
        System.out.println("Ignored");
      }
    }
  }
```

When setting `warmupPeriod=0` will trigger some downstream slope calculation with some value divide 0:  https://github.com/google/guava/blob/master/guava/src/com/google/common/util/concurrent/SmoothRateLimiter.java#L229
image


liufuyang:
@cpovirk Perhaps you can give a review of this?

liufuyang:
@cgdecker Perhaps you can take a look as well? :)

liufuyang:
@ronshapiro @kluever Pinging you two here as I saw you changed this ratelimiter recently 😄

PR #3725, opened 2019-11-30
 Re-expose constructor and method that were made private at Guava 22.0

sbabcoc:
This set of revisions restores the constructor and method of the
SimpleTimeLimiter class that were introduced in Guava 19.0 and made
private in Guava 22.0. This enables clients of the UrlChecker class of
Selenium 2 to upgrade to the latest Guava release to fix a security
defect.


sbabcoc:
You can see the affected class here: [UrlChecker](https://oss.sonatype.org/service/local/repositories/releases/archive/org/seleniumhq/selenium/selenium-remote-driver/2.53.1/selenium-remote-driver-2.53.1-sources.jar/!/org/openqa/selenium/net/UrlChecker.java)

sbabcoc:
@googlebot I signed it!

cgdecker:
The constructor/methods in question were `@Beta` APIs that were deprecated in Guava 22 and removed in 23, in accordance to our `@Beta` policy. No library should have been using `@Beta` APIs, and while it's unfortunate that some libraries have over the years, we're not planning to re-add APIs we removed because of that.

sbabcoc:
I understand that the decision of the Selenium implementers to use a `@Beta` interface was ill-advised, but development on Selenium 2 shut down before switching to the final form of `TimeLimiter`. Selenium 3 doesn't use this `@Beta` interface, but folks like me who still need to support Java 7 can't upgrade to Selenium 3. The constructor/method are still present in the SimpleTimeLimiter class... they're just `private`, and the corresponding variant of `callWithTimeout` has been removed from the `TimeLimiter` interface.
Because the Selenium 2 browser drivers rely on SimpleTimeLimiter, I currently can't upgrade Guava to any release newer than 19.0. This leaves me vulnerable to the unbounded memory allocation issue that was resolved in later Guava releases. Is there any chance you'd make an exception to your deprecation/removal policy for `@Beta` interfaces in this case?

sbabcoc:
@cgdecker Any chance for to be reconsidered?

Issue #3721, opened 2019-11-23
MutableNetwork.hasEdgeConnecting throw IllegalArgumentException when node are missing


fathzer:
According to the javadoc, the hasEdgeConnecting method is equivalent to nodes().contains(nodeU) && successors(nodeU).contains(nodeV).
But, unlike this expression, it throws an exception if the nodes are unknown.
Here is a simple demonstration
```java
import com.google.common.graph.MutableNetwork;
import com.google.common.graph.NetworkBuilder;

public class Bug {
	public static void main(String[] args) {
		final MutableNetwork<String, String> graph = NetworkBuilder.undirected().allowsParallelEdges(false).build();
		// Foolowing line prints false
		System.out.println(graph.nodes().contains("1") && graph.successors("1").contains("2"));
		// Next should print false but throws an IllegalArgumentException
		System.out.println(graph.hasEdgeConnecting("1", "2"));
	}
}
```
version guava-28.1-jre

nglorioso:
It looks like a bit subtle, but `successors(nodeU).contains(nodeV)` throws an `IllegalArgumentException` when `nodeU` is not in the network (which is the actual behavior you're seeing here).

However, you're right that the short-circuiting behavior of `&&` isn't replicating here. That seems like an important part of the documented behavior :). FWIW, it appears to also throw an IllegalArgumentException even if `"1"` is in the network but `"2"` isn't.

jrtom:
This is just a straight-up bug on our parts; I'm coding up a fix now.

The problem originates in the fact that `hasEdgeConnecting()` is actually (in effect) calling `edgesConnecting().isEmpty()`.  This seems reasonable on the face of it, but `edgesConnecting(u, v)` has the contract that it throws if either `u` or `v` are not in the graph; this is consistent with the contracts for the other `Set`-returning accessors, but it's not what we want for boolean-returning queries like `hasEdgeConnecting()`.

Sorry about that; thanks for the catch.  :)

jrtom:
I've submitted an internal fix; this should be propagating externally soon.

jrtom:
This is fixed in v28.2: https://github.com/google/guava/releases/tag/v28.2

PR #3726, opened 2019-12-01
MOE Sync 2019-12-01

nglorioso:

Commits:
=====
<p> Make incidentEdges() return edges in insertion order

df2c5527d5fe4a50135239f0b6b516058a37c5fa

-------

<p> AbstractNetwork: fix bug in AbstractNetwork.hasEdgeConnecting() causing it to throw if either endpoint was not in the graph.

RELNOTES=Fix bug in AbstractNetwork.hasEdgeConnecting() causing it to throw if either endpoint was not in the graph.  Originally reported as GitHub issue #3721.

8e0f821430acaa5b77c5c2c05380e3d29f18e410

-------

<p> Refactor AbstractGraphTest to allow for tests with ImmutableGraph.

Reason: ImmutableGraph implementations are undertested compared to MutableGraph implementations. The current tests didn't catch a bug I deliberately introduced in [] Also, it would be nice to be able to share the incident edge order tests.

Note about design: I started out by making an AbstractGraphTest subclass for mutable graphs. However, that would lead to a duplication of all 7 subclasses and most of their tests. The issue is that directed/undirected and mutable/immutable are orthogonal and we also want to test allowSelfLoops=true/false and incidentEdgeOrder=unordered/stable. The proposed solution is somewhat unconventional, but at least allows us to share much more code between the tests.

a6d0db50f6d678839f17c82a76341bb0d3d60b29

-------

<p> Merge AbstractXGraphTest and ConfigurableSimpleXGraphTest, thus removing one layer from the AbstractGraphTest tree.

Reason: Nobody in [] is using these abstract test classes (even though they are designed for general purpose graphs). However, even though they aren't used, these classes add a layer of complexity to an already complex class tree that is going to become more complex with the addition of ImmutableGraph tests and stable incident edge order tests.

ce50502b9bd6fbd909abed1afd1de2976246eef9

-------

<p> Make the AbstractGraphTest tree more sane with only leaf nodes non-abstract and merge the tests for with and without self-loops.

f5ac328e048606d0b40c7e0b1b55bccd6ff177eb

-------

<p> Merge StandardMutableDirectedGraphTest and its variant with stable order.

This is in preparation for adding an Immutable version that also needs tests for the stable order.

a0ebd7b6d2789a168df5e3e0a224513dfe1257ee

-------

<p> Add StandardImmutableDirectedGraphTest

033535490cf36040b4c2aba6a94e9b0f3e59564e

-------

<p> Annotate benign race in AbstractTable so as to avoid TSAN issues.

a4c3bcd7d1c8438d02d368a18c46c1b39f27e80d


jbduncan:
Great work on improving the graph tests @nymanjens! Looking forward to seeing it merged in. 👏

nymanjens:
> Great work on improving the graph tests @nymanjens! Looking forward to seeing it merged in.

Thanks!

PR #3727, opened 2019-12-03
Update to errorprone 2.3.4

dhatanian:
Errorprone 2.3.3 was missing classes such as `DoNotMock` . This didn't affect Guava, but it led to dependency conflicts for people using both Guava and Lombok.

Version 2.3.4 adss `DoNotMock` back, which solves this issue.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -252,7 +252,7 @@
       <dependency>
         <groupId>com.google.errorprone</groupId>
         <artifactId>error_prone_annotations</artifactId>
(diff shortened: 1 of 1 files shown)


dhatanian:
@googlebot I signed it!

cgdecker:
Thanks for the PR! It looks like we already have a commit in that makes this change; it should be synced out soon.

cgdecker:
(4eea0f79c27864039a653ec4ca09ef7585e439fc)

PR #3730, opened 2019-12-05
fix CaseFormat

zhixiangyuan:
The extra space requested here is for the new separator, so you need to use the passed separator.


zhixiangyuan:
@googlebot I fixed it.

cgdecker:
Thanks for the PR; it's been submitted internally and should be synced out soon.

PR #3735, opened 2019-12-10
Documentation: format the Javadoc of ByteStreams.read with @throw tags.

Prodigysov:
Formated the descriptions of possible exceptions of `ByteStreams.read` into `@throw` tags.


kevinb9n:
We will document the IOOBE.  As for NPE, though, it is implied for every Guava parameter not marked `@Nullable`, and we think it's too much noise to document every case.

Prodigysov:
Sounds good to me, thanks!

PR #3745, opened 2019-12-23
Refactor Lambda can be replaced method reference

wangzhaoning:


wangzhaoning:
@googlebot I signed it!

PR #3746, opened 2019-12-23
MOE Sync 2019-12-23

kluever:

Commits:
=====
<p> Make Compact(Linked)HashMap fall back to a java.util.(Linked)HashMap if hash flooding is detected, just as the immutable collections do.

40d75c94d568935159ce838c61940430fce54249

### DIFF
--- a/android/guava/src/com/google/common/collect/CompactHashMap.java
+++ b/android/guava/src/com/google/common/collect/CompactHashMap.java
@@ -24,6 +24,7 @@ import com.google.common.annotations.GwtIncompatible;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Objects;
 import com.google.common.base.Preconditions;
+import com.google.common.primitives.Ints;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import com.google.j2objc.annotations.WeakOuter;
 import java.io.IOException;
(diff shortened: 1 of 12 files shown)


android/guava/src/com/google/common/collect/CompactHashMap.java
```diff
@@ -100,14 +103,39 @@
...
 
   /**
-   * The hashtable. Its values are indexes to the keys, values, and entries arrays.
-   *
-   * <p>Currently, the UNSET value means "null pointer", and any positive value x is the actual
-   * index + 1.
+   * Maximum allowed false positive probability of detecting a hash flooding attack given random
+   * input.
+   */
+  @VisibleForTesting(
```
szarnekow: This formatting seems to be slightly off, here.

android/guava/src/com/google/common/collect/CompactHashMap.java
```diff
@@ -194,6 +222,36 @@ int allocArrays() {
...
+  @NullableDecl
+  Map<K, V> delegateOrNull() {
+    if (table instanceof Map) {
+      return (Map<K, V>) table;
+    }
+    return null;
+  }
+
+  Map<K, V> createHashFloodingResistantDelegate(int tableSize) {
+    return new LinkedHashMap<>(tableSize, 1.0f);
```
szarnekow: Should this be a HashMap according to the comment in line 114 and the override in CompactLinkedHashMap?

cpovirk: Thanks. The other one should be `LinkedHashMap`, too. I'll fix it.

android/guava/src/com/google/common/collect/CompactHashMap.java
cpovirk: This is from some automated stripping of an internal-only line. Maybe someday we'll update our export process to clean things like this up.

cpovirk:
Will redo this with a fix for the problems with our tests under Maven's test runner.

PR #3749, opened 2019-12-23
Adding travis_wait to mvn install

FanJups:
Why ?
Building such big APIs as guava   can take a lot of time.  **It's possible to get a time out**.
[Here](https://github.com/4hwc/4HWCAutonomousCar/commit/96c6396d444f0364150241c5bed3e0b9fc7ba5c8), I got a time out due to generating javadoc. Then, I decided to add travis_wait and it works pretty well.

Official docs : https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received


cpovirk:
We have `travis_wait` on the `script` line below. As far as I know, we haven't seen similar problems with `install`. If that changes, we should make this change, but until then, let's keep it simple.

Issue #3037, opened 2018-01-22
user-provided Ticker for RateLimiter


bdupras:
Internally, RateLimiter code can already take a provided SleepingStopWatch, which in turn can take a provided Ticker. The the Ticker instance used is hard coded one that reads from the system clock. 

I would like to supply my own time source, so it would be great to have a public `create` method that allows the caller to provide a customer Ticker. (Use-case: simulation environment with a time clock controlled by the sim, including ability to pause and fast-forward.)

If accepted, I can supply a PR for the code change if so desired.

findepi:
Another use-case, not mentioned in the issue description, is the ability to test code that uses `RateLimiter`.
Currently such needs need to `Thread.sleep`, which is obviously very far from ideal.

@cpovirk

findepi:
(Same request is also tracked by https://github.com/google/guava/issues/1774)

netdpb:
This is a duplicate of #1774. Closing this issue in favor of that one.

Issue #1774, opened 2014-10-31
Allow building RateLimiter instances with Ticker


gissuebot:
_[Original issue](https://code.google.com/p/guava-libraries/issues/detail?id=1774) created by **itt...@wix.com** on 2014-06-03 at 05:39 AM_

---

I'd like to be able to use rateLimiter and so I need to be able to test it.
I see you have a factory method which accepts a ticker but that is internal only; Such a method would help me.

thecoop:
I've also run into this issue, currently you can't use a RateLimiter in a unit test using a fake Ticker.

kluever:
We allow customer `Ticker` instances to be provided for other APIs, so this seems reasonable.

PR #3759, opened 2020-01-03
Allow user-provided Ticker in RateLimiter

findepi:
Fixes https://github.com/google/guava/issues/3037. Fixes https://github.com/google/guava/issues/1774

As mentioned in https://github.com/google/guava/issues/3037#issuecomment-570493413, this is needed to remove `Thread.sleep` from unit tests of classes that internally use `RateLimiter` (eg https://github.com/prestosql/presto/blob/cb62e11aa72116c13bcac50ddcfb8e0f01bbb02a/presto-hive/src/test/java/io/prestosql/plugin/hive/util/TestThrottledAsyncQueue.java#L105).

### DIFF
--- a/guava-tests/test/com/google/common/util/concurrent/RateLimiterTest.java
+++ b/guava-tests/test/com/google/common/util/concurrent/RateLimiterTest.java
@@ -25,6 +25,7 @@ import static java.util.concurrent.TimeUnit.SECONDS;
 import com.google.common.collect.ImmutableClassToInstanceMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Lists;
+import com.google.common.testing.FakeTicker;
 import com.google.common.testing.NullPointerTester;
 import com.google.common.testing.NullPointerTester.Visibility;
 import com.google.common.util.concurrent.RateLimiter.SleepingStopwatch;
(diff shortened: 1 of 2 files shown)


findepi:
I'd use advice how to deal with https://github.com/google/guava/blob/d08dd756cce43c3c27714e2d99286de7d08ff709/guava/src/com/google/common/util/concurrent/RateLimiter.java#L482

findepi:
@netdpb can you suggest me the correct way to deal with https://github.com/google/guava/pull/3759#issuecomment-570513834?

netdpb:
I'm not sure. That's what makes this task nontrivial.

findepi:
@netdpb @kluever 
Got it. At the same time, it's also a good thing to have.
(not being able to unit test code _because_ it uses Guava is so lame)

I propose that we introduce a `SleepingTicker` interface, with two methods.

* One for getting current time (Ticker)
* and another for sleeping no longer than until current time becomes x
  * for system clock-based implementation this will be Thread.sleep or "sleep uninterruptibly" (as current code uses the latter)

We don't need to provide a fake implementation (just as FakeTicker is not part of Guava API).

findepi:
I am closing this PR. However, if you decide you want to review my contribution (along what i proposed or otherwise), let me know.

PR #3761, opened 2020-01-05
Fix off-by-one error in AsciiDigits.asciiDigits initialization

perceptron8:
`AsciiDigits.digit('[')` should return -1 instead of 36.

I changed both loops to consistently use strict inequalities. Now it's clear that there are 10 + 26 = 36 digits / iterations in total.

Fortunately this bug doesn't affect anything other than `Longs.tryParse` where it is - by accident - harmless.

https://github.com/google/guava/blob/v28.2/guava/src/com/google/common/primitives/Longs.java#L405
https://github.com/google/guava/blob/v28.2/guava/src/com/google/common/primitives/Longs.java#L414

Note that `Longs.tryParse('[', 36)` and `Longs.tryParse('0[', 36)` currently return `null` because of `digit >= radix` condition, not `digit < 0` as it should be.

### DIFF
--- a/android/guava/src/com/google/common/primitives/Longs.java
+++ b/android/guava/src/com/google/common/primitives/Longs.java
@@ -328,10 +328,10 @@ public final class Longs {
     static {
       byte[] result = new byte[128];
       Arrays.fill(result, (byte) -1);
-      for (int i = 0; i <= 9; i++) {
+      for (int i = 0; i < 10; i++) {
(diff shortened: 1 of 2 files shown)


perceptron8:
I said about '[' = 'A' + 26 = 'Z' + 1, but I forgot to mention about '{' = 'a' + 26 = 'z' + 1. Just for the record.

PR #3778, opened 2020-01-22
Add more context to the guava.dev landing page

CydeWeys:
This change makes it clear how widely used Guava actually is; it's not just another throw-away library, it's essential to Java development with Google and many other companies.

---
This change is [Reviewable](https://reviewable.io/reviews/google/guava/3778)

### DIFF
--- a/index.md
+++ b/index.md
@@ -3,9 +3,11 @@ layout: default
 title: Guava
 ---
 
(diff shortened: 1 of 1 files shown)


CydeWeys:
Ping?

PR #3788, opened 2020-01-29
Make AbstractFuture compatible with ForkJoinPool by catching exceptions from property retrieval

ElliottBrossard:
Context: https://github.com/google/guava/issues/3784

PR #3795, opened 2020-02-06
DO NOT MERGE: Make AbstractFuture compatible with ForkJoinPool (taken over from #3788)

cpovirk:


cpovirk:
Will make this change internally.

PR #3798, opened 2020-02-11
arXiv.org e-Print archive

85u2e8:
`e:print = ('' arxiv.org'')`


nglorioso:
Hi there. I think this change isn't intended to be merged into Guava (is it for a different project?)

PR #3800, opened 2020-02-15
Simplify redundant `Collection.addALl()` call

zhangt2333:
Simplify redundant `Collection.addALl()` call and Optimize performance.

### DIFF
--- a/guava-testlib/src/com/google/common/collect/testing/CollectionTestSuiteBuilder.java
+++ b/guava-testlib/src/com/google/common/collect/testing/CollectionTestSuiteBuilder.java
@@ -91,8 +91,7 @@ public class CollectionTestSuiteBuilder<E>
   }
 
   private static Set<Feature<?>> computeReserializedCollectionFeatures(Set<Feature<?>> features) {
-    Set<Feature<?>> derivedFeatures = new HashSet<>();
-    derivedFeatures.addAll(features);
+    Set<Feature<?>> derivedFeatures = new HashSet<>(features);
     derivedFeatures.remove(SERIALIZABLE);
     derivedFeatures.remove(SERIALIZABLE_INCLUDING_VIEWS);
     return derivedFeatures;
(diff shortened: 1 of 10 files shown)


zhangt2333:
@googlebot  I signed it!

cpovirk:
A couple quick thoughts, since I sadly doubt that I'll be able to dedicate a lot of time to this:

- I somewhat like to create the collection on one line and add to it on another _if_ I am then going to also _remove_ from it (or add more to it), like [here](https://github.com/google/guava/pull/3800/files#diff-f23f8f184b38c696a270e2e0ae0114eeL172). The multi-line process is a hint that there's going to be more to the collection than just its initial contents. That is, we're not just making a defensive copy. However, this is a minor point, and others might not see it that way. Plus, we're usually modifying the collection _immediately_ after creating, so there's not much danger of confusion.

- For cases in which we add objects and then add more objects, I'm not sure that we necessarily improve performance by presizing to the size of the _first_ set of objects, like [here](https://github.com/google/guava/pull/3800/files#diff-c0a860430c2be8897c12525322b7fda5R491). It might actually be better to accept the default size.

The unfortunate thing is that this code does need some performance tuning and general cleaning. I just fear that it would be a big investment.

zhangt2333:
@cpiotr 
Thanks for your code review. This PR was submitted for the following reasons : 
* Using the constructor with parameter, it can save one time array copy.
* Improving readability and polishing the code.

But I agree with your thoughts : 
* For the contextual semantic unity in a block of code.
* The performance improvement may be small.
* More time can be allocated for implementing features and fixing bugs

I respect your decision, so what should I do?

cpovirk:
Thanks. I'm going to skip `DerivedCollectionGenerators.java`, since I'm not sure if that will help or hurt performance. (I notice that that code could probably be simplified even further, but I left it alone.) I'm looking to merge the rest.

PR #3816, opened 2020-03-06
Add UTs for base64Url() decode in BaseEncoding

dyrone:
I use base64Url in code if Gerrit to make the base64 URL safe.

But found that it doesn't cover some basic cases.

### DIFF
--- a/guava-tests/test/com/google/common/io/BaseEncodingTest.java
+++ b/guava-tests/test/com/google/common/io/BaseEncodingTest.java
@@ -19,6 +19,7 @@ import static com.google.common.io.BaseEncoding.base16;
 import static com.google.common.io.BaseEncoding.base32;
 import static com.google.common.io.BaseEncoding.base32Hex;
 import static com.google.common.io.BaseEncoding.base64;
+import static com.google.common.io.BaseEncoding.base64Url;
 import static com.google.common.truth.Truth.assertThat;
 
 import com.google.common.annotations.GwtCompatible;
@@ -105,6 +106,11 @@ public class BaseEncodingTest extends TestCase {
     testDecodes(base64(), "Zg====", "f");
   }
 
+  public void testBase64Url() {
+    testDecodesByBytes(base64Url(), "_zzz", new byte[] {-1, 60, -13});
+    testDecodesByBytes(base64Url(), "-zzz", new byte[] {-5, 60, -13});
+  }
+
   public void testBase64InvalidDecodings() {
     // These contain bytes not in the decodabet.
     assertFailsToDecode(base64(), "A\u007f", "Unrecognized character: 0x7f");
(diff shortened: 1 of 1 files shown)


guava-tests/test/com/google/common/io/BaseEncodingTest.java
```diff
@@ -51,6 +54,34 @@ public void testSeparatorsExplicitly() {
     testEncodes(base32().withSeparator("*", 4), "foobar", "MZXW*6YTB*OI==*====");
   }
 
+  public void testBase64UrlWithPlus() {
```
nglorioso: I would turn these into members in `testBase64InvalidDecodings()`

guava-tests/test/com/google/common/io/BaseEncodingTest.java
```diff
@@ -51,6 +54,34 @@ public void testSeparatorsExplicitly() {
...
+
+  public void testBase64UrlWithSlash() {
+    try {
+      base64Url().decode("/zzz");
+      fail("Expected IllegalArgumentException");
+    } catch (IllegalArgumentException expected) {
+    }
+  }
+
+  public void testBase64UrlWithUnderline() {
```
nglorioso: Can you use the testDecodes() helper method for these?

guava-tests/test/com/google/common/io/BaseEncodingTest.java
dyrone: Yeah

 Your suggestions is exactily right.

guava-tests/test/com/google/common/io/BaseEncodingTest.java
dyrone: testDecodes() is use UTF-8 set to encoding, this may limit the scope of BASE64 decoding and make it failed.

I added a new test function names testDecodesByBytes() to compare decode bytes directly.

guava-tests/test/com/google/common/io/BaseEncodingTest.java
nglorioso: Hi There! It looks like you've added these to `testBase16InvalidDecodings` (whereas base64 would be more appropriate).

Given that `base64Url()` is a different decoder than `base64()`,  I would consider adding a number of new unit test functions, situated nearby the block of base64 tests, something like `testBase64Url()` and `testBase64UrlInvalidDecodings()`.

dyrone: Thanks, Nick 
Sorry for my negligence.
I pushed the amend change with your suggestions

dyrone: @nglorioso  Hi

Can you help to review the new path?

Thank you very much.

nglorioso: Thanks for your patience! I'm going to make a copy of this commit internally (attributed to you), then vend that back out, hopefully by tomorrow, if not Monday.

dyrone: Thank you.

Best regards

PR #3822, opened 2020-03-13
Change maximum capacity for Maps as per JDK 8

kkraja555:
If initialCapacity is greater than Ints.MAX_POWER_OF_TWO then JDK 8 uses Ints.MAX_POWER_OF_TWO as MAXIMUM_CAPACITY.
So, Even if we return Integer.MAX_VALUE as MAXIMUM_CAPACITY, JDK gonna use Ints.MAX_POWER_OF_TWO as MAXIMUM_CAPACITY

Snippet from JDK 8 Source:
public HashMap(int initialCapacity, float loadFactor) {
if (initialCapacity < 0)
throw new IllegalArgumentException("Illegal initial capacity: " +
initialCapacity);
if (initialCapacity > MAXIMUM_CAPACITY)
initialCapacity = MAXIMUM_CAPACITY; // which is Ints.MAX_POWER_OF_TWO

### DIFF
--- a/guava-tests/test/com/google/common/collect/MapsTest.java
+++ b/guava-tests/test/com/google/common/collect/MapsTest.java
@@ -35,6 +35,7 @@ import com.google.common.base.Predicates;
 import com.google.common.collect.Maps.EntryTransformer;
 import com.google.common.collect.Maps.ValueDifferenceImpl;
 import com.google.common.collect.SetsTest.Derived;
+import com.google.common.primitives.Ints;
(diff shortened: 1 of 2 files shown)


cpovirk:
As I understand it, this is not a behavior change. Additionally, if a `HashMap` implementation someday supports a larger internal capacity, we would want to take advantage of it.

kkraja555:
Makes sense. Thanks

Issue #3790, opened 2020-01-31
NoSuchElementException when using Lists.partition and remove


amine-java:
Hello, 

It seems that there's a bug in the implementation of the Lists.partition. Here's the problem; assume that I have partitionned a List of Integers, if you perform a loop over the partitionned list, and for each subList you perform a removeAll() function, you'll get an error (NoSuchElementException)

Here's the code: 

![image](https://user-images.githubusercontent.com/7299923/73552339-5275b000-4448-11ea-8d7a-79090f68eaf2.png)

  It seems like whenever the subList is emptied, the empty subList is **REMOVED** from the partitionned list and therefore causes the exception mentioned above, the exception is thrown from the next() method inside AbstractList.java

dimo414:
This is an interesting failure mode 🙂 ultimately I suspect the caveat in the docs about mutations applies:

> The inner lists are ... subject to all the usual caveats about modification as explained in [`List.sublist()`].

That said, this is definitely odd behavior. Consider:

```
jshell> var partitioned = Lists.partition(Lists.newArrayList(1), 2)
partitioned ==> [[1]]

jshell> var iter = partitioned.iterator()
iter ==> java.util.AbstractList$Itr@28c4711c

jshell> iter.hasNext()
$5 ==> true

jshell> iter.next().removeAll(Lists.newArrayList(1))
$6 ==> true

jshell> iter.hasNext()
$7 ==> true

jshell> iter.next()
|  Exception java.util.NoSuchElementException
|        at AbstractList$Itr.next (AbstractList.java:377)
|        at (#8:1)
```

`partitioned` only starts with one element! Why would `iter.hasNext()` return `true` the second time it's called?

We can trigger the same behavior with an even simpler example that doesn't involve `Lists.partition()`:

```
jshell> var ls = Lists.newArrayList(1)
ls ==> [1]

jshell> var iter = ls.iterator()
iter ==> java.util.ArrayList$Itr@534df152

jshell> iter.hasNext()
$11 ==> true

jshell> iter.next()
$12 ==> 1

jshell> ls.remove(0)
$13 ==> 1

jshell> iter.hasNext()
$14 ==> true
```

That's not right! Calling `.next()` will similarly fail, though with a slightly more informative `ConcurrentModificationException`. Either way though, this directly violates the [`Iterator` spec](https://docs.oracle.com/en/java/javase/13/docs/api/java.base/java/util/Iterator.html#hasNext()), which says:

[...]

PR #3828, opened 2020-03-18
Fixed NoSuchElementException when using Lists.partition and remove #3790

Buddy119:
NoSuchElementException occurs because of Iterator in AbstractList call hasNext() as below
```
public boolean hasNext() {
            return cursor != size();
        }
```

But after execute remove operation on List, the result of calling size() will change. So I make size as a final int listSize.

### DIFF
--- a/guava/src/com/google/common/collect/Lists.java
+++ b/guava/src/com/google/common/collect/Lists.java
@@ -664,9 +664,11 @@ public final class Lists {
   private static class Partition<T> extends AbstractList<List<T>> {
     final List<T> list;
     final int size;
+    final int listSize;
 
     Partition(List<T> list, int size) {
       this.list = list;
+      this.listSize = IntMath.divide(list.size(), size, RoundingMode.CEILING);
       this.size = size;
     }
 
@@ -680,7 +682,7 @@ public final class Lists {
 
     @Override
     public int size() {
-      return IntMath.divide(list.size(), size, RoundingMode.CEILING);
+      return this.listSize;
     }
 
     @Override


Buddy119:
I test my commit locally. But do not add test in commit. If needed, i will add test in the future.

cpovirk:
From the discussion on #3790, it sounds like, if anything, we would want to make `hasNext()` throw a `ConcurrentModificationException`.

Buddy119:
@cpovirk 
Hi, I have reviewed the discussion. But I think the expected behavior should not throw an exception.

According to the discussion, they are confused on why **iter.hasNext()** is true. Actually, it is because class **Partition** use **Iterator** from superclass **AbstractList**. And the **hasNext()** function in class **AbstractList** is not perfect. They have been optimized in subclass **ArrayList**.Plese review below

This is **hasNext** in **AbstractList** class. It call **size()** function
```
public boolean hasNext() {
    return cursor != size();
} 
```
		
This is **hasNext** in **ArrayList** class. It use **value size**. And the documentation said it is *An optimized version of AbstractList.Itr.* I see this in JDK1.8

```
public boolean hasNext() {
    return cursor != size;
}
```	

when use fuction **size()**, every time propram call **hasNext()**. The **hasNext()** will call **size()**, then the **value size** will be recalculated. Back to the code from #3570, when the propram execute **removeall()**, the recalculate size will be 0, but the cursor is 1. so the **hasNext()** return true.

example code in #3570
```
for(List<Integer> l : ll){
   l.removeAll(elementExclude);
}
```

So I introduce **listsize** to save the value to make sure it only be calculated once.

I also tested below code. I use ArrayList as container and then do the same **removeAll** operation. No exception is throw. So, I think the behavior of **partition** should be same

```
        List<Integer> list = new ArrayList();
        list.add(1);
        List<Integer> elementExclude = new ArrayList();
        elementExclude.add(1);
        
        List<List<Integer>> ll = new ArrayList();
        
        ll.add(list);
        for(List<Integer> l : ll){

            l.removeAll(elementExclude);
        }
```

Buddy119:
Typo 3790 not 3570

cpovirk:
Let's consolidate the discussion on the other thread. We can reopen this one (or a new PR) if there is agreement on what to do.

Buddy119:
@cpovirk  ok, i will discuss my thoughts in #3790 with them. Anyway thanks

PR #3843, opened 2020-04-03
Fix unnecessary method access level escalation

aczaplinski:
The overridden methods are protected in superclasses, and there is no need do make them public in subclasses. Why this change doesn't break anything? Because the objects containing the changed methods are passed as HashFunction implementations, which hides the methods anyway.


aczaplinski:
@googlebot I fixed it.

cgdecker:
Thanks, this is being submitted internally and will be synced out soon.

Issue #3610, opened 2019-09-16
Multimap iterator remove leads to ConcurrentModificationException


bennofs:
Simple example:

```
class Scratch {
    public static void main(String[] args) {
        final Multimap<String, String> map = HashMultimap.create();
        map.put("test", "x");
        map.put("xx", "z");
        for (Iterator<Map.Entry<String, Collection<String>>> it = map.asMap().entrySet().iterator(); it.hasNext(); ) {
            Map.Entry<String, Collection<String>> entry = it.next();
            Iterator<String> it2 = entry.getValue().iterator();
            it2.next();
            it2.remove();
        }
        System.out.println(map);
    }
}
```

I think this happens because `it2.remove()` will detect that the collection is empty, which triggers a `.remove` call on the underlying map that is not done through the iterator. The correct thing here would be to call `.remove` on the underlying map iterator as well.

cpovirk:
This is definitely unfortunate. But I think it's unfixable in the general case, at least without significant tradeoffs.

I haven't tested this, but I'd expect problems with a case like the following:

```java
map.put("a", "a");
map.put("b", "b");
map.put("c", "c");
Iterator<Map.Entry<String, Collection<String>>> it = map.asMap().entrySet().iterator();
Map.Entry<String, Collection<String>> first = it.next();
Map.Entry<String, Collection<String>> second = it.next();
Iterator<String> it2 = first.next().getValue().iterator();
it2.remove();
it.next();
```

By the time we call `it2.remove()`, we've already called `it.next()` again, so we can't use `it.remove()` to perform the removal. So we end up with the same problem on the next `it.next()` call. (We could have special handling for the more common case that you've described, but that's probably more complexity than we'd prefer, both in the implementation and in the API.)

There might be more clever solutions we could employ, like making the underlying map key not the `String` directly but rather a `Holder<String>`, which we could then null out on a call to `remove()` instead of fully removing the key. (This would use some extra storage temporarily, but hopefully it would get cleared out eventually when users perform some other structural modification, which we can detect through the view collections. And there's also the ongoing overhead (including indirection) from the `Holder` itself, though maybe that can be avoided with a custom hash table -- which we can do (and already do in places), but it's extra code (bad for Android) and more complexity.)

[...]

PR #3844, opened 2020-04-03
Issue 3610 - documented exception when last element of a value collection is removed

apostrophe:


cgdecker:
I'm not sure this is worth documenting, at least not in the main documentation of `Multimap` itself. It's standard for the Java collections framework that unless a collection documents otherwise (e.g. the concurrent collections), modifications to its contents during iteration may cause `ConcurrentModificationException`.

cpovirk:
I feel a little bad for [having suggested documenting this](https://github.com/google/guava/issues/3610#issuecomment-532282138). I see how there's not a great place to do so :\

PR #3845, opened 2020-04-03
Make singletons' constructors private

aczaplinski:


cky:
I'm not on the Guava team. However, I've looked at Guava code for a long time, and generally, Guava leans heavily on implicit trust for package-private code. In this instance, for example, the Guava team probably trusts that there won't be any internal code that would try to instantiate those classes directly.

Besides, if the Guava team did want to make this change for some reason, the usual Effective Java pattern for singletons is to use a one-constant enum.

aczaplinski:
The one-constant enum singleton implementation can't be applied in these cases - the classes extend some other classes, which enums in Java can't do.

cgdecker:
While in general it's preferable to have a private constructor for singletons, this doesn't really buy us anything in this case.

Issue #3669, opened 2019-10-24
Inheriting Javadoc from the JDK is broken on Travis


cpovirk:
This means that our snapshots are missing doc for methods like `equals`:

https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/primitives/UnsignedInteger.html#equals-java.lang.Object-

Note that our releases are fine. This suggests that there's something insufficient about our setup specifically on Travis. (e.g., maybe it omits `src.zip` there to improve performance.)

I made one attempt to fix this (based on a different theory) in https://github.com/google/guava/commit/4497e2a620892a0540a7d739519cdf77612539fc, but it didn't work, and I've rolled it back internally (to be mirrored out shortly).

We could probably play around with submitting PRs that run `ls -lR $JAVA_HOME` or something, but I don't plan to look into it at the moment.

developerhb:
Hello @cpovirk 

I am Hem Bhagat. I am in 3rd year of Bachelor in Computer Engineering. I am interested in open source projects and want to contribute in them. I chose this project because I am quite interested in Java programming language and want to make a career in it. 

I would like to get detail information about what the problem is?

Please also provide me information about the prerequisites that are necessary for me to solve this problem.

Thank you.

developerhb:
@cpovirk I am waiting for your reply.

cpovirk:
Hello,

The main thing we need to do is figure out if `src.zip` (a zip file containing the source `.java` files for the JDK) is available on Travis for openjdk11.

One thing that you could do is create a pull request that edits [our Travis configuration](https://github.com/google/guava/blob/31999ae6f56e71c341333afde90fcf66c13ce657/.travis.yml#L12). Instead of running our tests, you could tell it to run commands like `ls -lR` on the directory that contains the JDK. When you submit the PR, Travis will run, and you could look for a `src.zip` in its output.

It's possible that [`src.zip` exists but is a symlink pointing to the wrong location](https://bugs.launchpad.net/ubuntu/+source/openjdk-lts/+bug/1791219), or maybe [we need to install an additional package to get the sources](https://askubuntu.com/a/1104463).

If we can't find a JDK 11 `src.zip` at all, then another alternative is to see if we can install the `src.zip` from an older version of Java and use that instead.

developerhb:
Hello @cpovirk 

I had a look at the travis configuration and travis documentation for using different configuration. From that understanding we can have following configuration to list all the files in the JDK folder.

```
deploy:
       provider: script
[...]

PR #3853, opened 2020-04-09
MOE Sync 2020-04-09

cpovirk:

Commits:
=====
<p> Try yet again to inherit docs from the JDK when building Javadoc for snapshots.

I am optimistic: See https://github.com/google/guava/issues/3669#issuecomment-611658339

4b55cc246fc8c8660540f2e3b1617777333c0154

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,12 @@ jdk:
   - openjdk8
   - openjdk11
 
+addons:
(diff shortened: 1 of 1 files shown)


cpovirk:
Need to fix internally to match the extra commit I added externally. Will open a new PR.

Issue #3859, opened 2020-04-10
Fallback to getMethods() if getDeclaredMethods() causes SecurityException in SubscriberRegister


vrozov:
`getDeclaredMethods()` requires  "accessDeclaredMembers" runtime permission that may not be granted to the library. As `@Subscribe` methods are expected to be public, it is possible to fallback to `getMethods()` when `getDeclaredMethods()` causes `SecurityException`.

netdpb:
`getMethods()` is also subject to security management, no?

vrozov:
It is still subject to security management but with fewer restrictions (does not require "accessDeclaredMembers" permission).

PR #3860, opened 2020-04-10
Filter out java.lang.Object from checked types and fallback to getMethods() if getDeclaredMethods() causes SecurityException.

### LINKED ISSUE #3858: SubscriberRegistry should exclude java.lang.Object from checked types.
As `java.lang.Object` can't have `@Subscriber` annotation, it can be safely excluded from checked types in `SubscriberRegistry`.

vrozov:
fixes #3858 and #3859

### DIFF
--- a/guava/src/com/google/common/eventbus/SubscriberRegistry.java
+++ b/guava/src/com/google/common/eventbus/SubscriberRegistry.java
@@ -37,9 +37,9 @@ import com.google.j2objc.annotations.Weak;
 import java.lang.reflect.Method;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.ConcurrentMap;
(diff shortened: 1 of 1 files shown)


guava/src/com/google/common/eventbus/SubscriberRegistry.java
```diff
@@ -175,27 +175,35 @@ void unregister(Object listener) {
...
+    Set<MethodIdentifier> identifiers = new HashSet<>();
+    return supertypes
+      .stream()
+      .filter(cls -> !Object.class.equals(cls))
+      .flatMap(supertype -> {
+        Method[] methods;
+        try {
+          methods = supertype.getDeclaredMethods();
+        } catch (SecurityException e) {
+          methods = supertype.getMethods();
```
Qpm-wen, line 187: using `getMethods` will call anothor bug I think.
`getMehods` will get all the methods from subclass, include its superclass.  so when code like that:
```
class ListenerA {
  @Subscribe void listenA(Event event) {}
} 

class ListenerB extends ListenerA {
  @Subscribe void listenB(Event event) {}
}
```

if we use `ListenerB.class.getMethods`，we will get `ListenerA.listenA` as well. it will be a bug, in this code, we will get 2 method obj of `ListenersA.listenA`, one froms ListenerA it self, another froms ListenerB

ljacqu:
Just as a viewer passing by—

Conceptually: if you fall back to `getMethods` it's a little excessive to call it for every supertype since you get all parent methods immediately. Might be worth checking first whether you can call `getDeclaredMethods` and if not to use `clazz.getMethods` as your sole source of methods.

Stylistically: I find the two streams within each other to be really awkward. The previous version with regular loops was much more clear IMHO. Streams work out nicely when you have a small _filter-and-transform_ thing going on but nesting them makes it difficult to remember where one is located.
I also find it awkward that the Set `identifiers` is just used for a uniqueness check (`filter` on L205) when really at the end you could just return that `identifiers` mapped in an ImmutableList, just like before. Also just for reference the filter call on L192 that always returns true could be turned into a `peek`.

guava/src/com/google/common/eventbus/SubscriberRegistry.java
vrozov: Please explain why that will be a bug?

Qpm-wen: sorry, I ignored the `MethodIdentifier filter` you writed at last, I had make a test for my comment!  : )
```
identifiers = HashSet();
.filter(method -> identifiers.add(new MethodIdentifier(method)))
```
and here is my test code.
```java
  public void testSuperAnsSubSubscribe() {
    SubClass subClass = new SubClass();
    registry.register(subClass);
    assertEquals(2, Iterators.size(registry.getSubscribers(Integer.valueOf(1))));
  }

  private static class SuperClass {
    @Subscribe
    public void superHandle(Integer i) {}
  }

  private static class SubClass extends SuperClass {
    @Subscribe
    public void subHandle(Integer i) {}
  }
```

Tizzit-max:
afaaf1e3bd9e93aedaa9762573b2e0e1cc3108d2

Tizzit-max:
Conflicting files
guava/src/com/google/common/eventbus/SubscriberRegistry.java

vrozov:
Any plan to pull this into a release?

vrozov:
@netdpb the PR is open for almost a year, are there any plan to merge it?

Issue #3742, opened 2019-12-19
Sets.cartesianProduct gives undocumented IllegalArgumentException due to its size


Meijuh:
The following piece of code throws an undocumented `IllegalArgumentException`:

    final Set<String> products = Sets.newHashSet("coffee", "tea");
    final List<Set<String>> productOf = Collections.nCopies(31, products);    
    assert Sets.cartesianProduct(productOf).size() == Integer.MAX_VALUE;

The fact that `Sets::cartesianProduct` will throw an `IllegalArgumentException` when there would be more than 2^31-1 elements in the resulting set is not documented as far as I can see.

I would like to see one of two solutions:

1. Document why this Exception is thrown (there may be an implementation detail that prevents *even defining* such large sets).
2. Fix it, because `java.util.Set::size()` specifies that if the size would be more than `Integer.MAX_VALUE` it would simply be `Integer.MAX_VALUE`.

I like to see this fixed (solution 2), because now I can't even define such a Cartesian product. Iterating over so many elements is of course a different issue, but I have a piece of code that may very quickly find an `element` in the product, because I know it is in there and due to some ordering on `element`.

developerhb:
@Meijuh @kluever I think we should fix this(solution 2). I will have a look at it.

PR #3870, opened 2020-04-21
Issue #3742 Changing the axesSizeProduct entries to Integer.MAX_VALUE instead of throwing exception

HelenaGregg:
#3742 said that they would like to be able to create a `Sets::cartesianProduct` object that would have more than 2^31-1 elements in the resulting set. Currently, it throws an `IllegalArgumentException` when this happens. @Meijuh said that since "java.util.Set::size() specifies that if the size would be more than Integer.MAX_VALUE it would simply be Integer.MAX_VALUE" this should be fixed.
So the code change is that instead of throwing an exception, the remaining lower parts of the axesSizeProduct integer array are set to Integer.MAX_VALUE so the size function (which returns axesSizeProduct[0]) will now return Integer.MAX_VALUE.


HelenaGregg:
@googlebot I signed it!

HelenaGregg:
I was having issues with CLA so opened a new request.

PR #3871, opened 2020-04-22
Issue #3742 Changing the axesSizeProduct entries to Integer.MAX_VALUE instead of throwing exception

HelenaGregg:
#3742 said that they would like to be able to create a `Sets::cartesianProduct` object that would have more than 2^31-1 elements in the resulting set. Currently, it throws an `IllegalArgumentException` when this happens. @Meijuh said that since "java.util.Set::size() specifies that if the size would be more than Integer.MAX_VALUE it would simply be Integer.MAX_VALUE" this should be fixed.
So the code change is that instead of throwing an exception, the remaining lower parts of the axesSizeProduct integer array are set to Integer.MAX_VALUE so the size function (which returns axesSizeProduct[0]) will now return Integer.MAX_VALUE.

### DIFF
--- a/guava/src/com/google/common/collect/CartesianList.java
+++ b/guava/src/com/google/common/collect/CartesianList.java
@@ -57,10 +57,10 @@ final class CartesianList<E> extends AbstractList<List<E>> implements RandomAcce
         axesSizeProduct[i] = IntMath.checkedMultiply(axesSizeProduct[i + 1], axes.get(i).size());
       }
     } catch (ArithmeticException e) {
-        // size which returns axesSizeProduct[0] will now return Integer.MAX_VALUE
-        for (; i >= 0; i--) {
-          axesSizeProduct[i] = Integer.MAX_VALUE;
-        }
+      // size which returns axesSizeProduct[0] will now return Integer.MAX_VALUE
+      for (; i >= 0; i--) {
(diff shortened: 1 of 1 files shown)


lowasser:
Several other methods are defined in terms of `axesSizeProduct` and make assumptions about how it works, so please add tests for them; I suspect that several of them are broken in the current version.
 In particular, `iterator` and `contains`.  (Further thinking about this brought up #3873 and #3874 .)  

In particular:
 * I'm not sure `List.indexOf` is well-defined for a list of size greater than `Integer.MAX_VALUE`, so we'd have to make a decision what to do there.  I prefer throwing `IndexOutOfBoundsException`, if only because `Objects.equal(list.get(list.indexOf(o)), o)` should always be true if `index` is not -1, and `list.indexOf(o) >= 0` should be as equivalent as possible to `list.contains(o)`.  (Note that this change also affects `Lists.cartesianProduct`, so `indexOf` is a method that needs to be supported.
* The iterator needs to return the next element after you've called `next` `Integer.MAX_VALUE` times.  Unfortunately, I'm not sure that can be tested without being very slow, and I don't know how to solve that.
* `contains` definitely needs to be tested, including on elements whose "index" would be greater than `Integer.MAX_VALUE`.  That probably means it needs a new implementation separate from `indexOf`, which would ideally also address #3873.

Issue #3874, opened 2020-04-22
Lists.cartesianProduct.lastIndexOf should be as "smart" as indexOf


lowasser:
The implementation should be pretty much equivalent, but the asymptotic complexity difference is huge.

lowasser:
Closed by https://github.com/google/guava/commit/48c7cfd5e38c8492e4356aab5c633254d1ce43be

PR #3878, opened 2020-04-27
Issue #3874: Improve cartesianProduct.lastIndexOf 

Koellewe:
Implemented `lastIndexOf()` in `CertesianList` and added appropriate tests.

### DIFF
--- a/guava-tests/test/com/google/common/collect/ListsTest.java
+++ b/guava-tests/test/com/google/common/collect/ListsTest.java
@@ -629,6 +629,15 @@ public class ListsTest extends TestCase {
     assertEquals(actual.indexOf(list(1, 1, 1)), -1);
   }
 
+  public void testCartesianProduct_lastIndexOf() {
+    List<List<Integer>> actual = Lists.cartesianProduct(list(1, 1), list(2, 3));
+    assertEquals(actual.lastIndexOf(list(1, 2)), 2);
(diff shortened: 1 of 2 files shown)


guava-tests/test/com/google/common/collect/ListsTest.java
```diff
@@ -629,6 +629,15 @@ public void testCartesianProduct_indexOf() {
     assertEquals(actual.indexOf(list(1, 1, 1)), -1);
   }
 
+  public void testCartesianProduct_lastIndexOf() {
+    List<List<Integer>> actual = Lists.cartesianProduct(list(1, 1), list(2, 3));
+    assertEquals(actual.lastIndexOf(list(1, 2)), 2);
```
lowasser, line 634: `assertEquals` actually expects its arguments in the other order -- expected, then actual.  Could you use Truth `assertThat(...).isEqualTo(...)` to avoid that confusion?

guava-tests/test/com/google/common/collect/ListsTest.java
```diff
@@ -629,6 +629,15 @@ public void testCartesianProduct_indexOf() {
...
+    assertEquals(actual.lastIndexOf(list(1, 3)), 3);
```
lowasser, line 635: Could you add a test that a list of the correct length does not have an index?

PR #3883, opened 2020-04-28
Move comment to a more appropriate location

szarnekow:
Signed-off-by: Sebastian Zarnekow <sebastian.zarnekow@gmail.com>


szarnekow:
@googlebot I signed it!

cpovirk:
Thanks!

A quick question in case you happen to have immediate thoughts on it. If not, no problem.

Is it suspicious that the `putCoalescing` implementation for `subRangeMap` has the very kind of `range.isEmpty()` check that this comment seems to warn against?
https://github.com/google/guava/blob/bc7248bcb0a987e18cf7a4e7ef92ed458086f4b0/guava/src/com/google/common/collect/TreeRangeMap.java#L511

szarnekow:
Good catch. Test and fix is in https://github.com/google/guava/pull/3886

PR #3884, opened 2020-04-28
Smarter implementation of CartesianList#lastIndexOf

the-thing:
As per https://github.com/google/guava/issues/3874 suggestion implemented a smarter version of `com.google.common.collect.CartesianList#lastIndexOf`.


the-thing:
I just realised there is already one pull request opened https://github.com/google/guava/pull/3878

PR #3886, opened 2020-04-28
Fixed TreeRangeMap.SubRangeMap.putCoalescing

szarnekow:
The short circuiting on an attempt to put an empty range caused the contract of putCoalescing to be violated.

Signed-off-by: Sebastian Zarnekow <sebastian.zarnekow@gmail.com>


cpovirk:
Thanks!

Issue #3890, opened 2020-05-03
PowerSet.equals depends on set's iteration order


RyanSkonnord:
If `com.google.common.collect.Sets.powerSet` is called with two different sets that contain the same elements but in a different iteration order, the returned power sets will also contain the same elements but in a different iteration order. Therefore, `Set.equals` on those power sets is `true`, but `com.google.common.collect.Sets.PowerSet.equals` incorrectly returns `false`.

cpovirk:
Wow, almost 7 years old! Nice find.

RyanSkonnord:
Thanks! :grin:

mattysday:
👍👍👍👍😀😀

PR #3891, opened 2020-05-03
Fix equals when comparing two PowerSet instances with different order

RyanSkonnord:
Fixes https://github.com/google/guava/issues/3890


nglorioso:
Thanks for the improvement! That certainly does look like a bug, and I appreciate the reproducing test.

I'll mirror this change internally, and you should see it added back out later today or tomorrow.

RyanSkonnord:
Cool; thank you! :grin: I'm very glad to have helped.

PR #3892, opened 2020-05-04
MOE Sync 2020-05-04

nglorioso:

Commits:
=====
<p> Fix PowerSet.equals() when comparing to another PowerSet whose items are the same, but in a different iteration order.

RELNOTES:
  Fix issue where PowerSet.equals(PowerSet) would erroneously return
  false if the PowerSet's underlying Sets were equal, but in a different
  iteration order.

Fixes #3891, #3890

befd5ced9edc93a65c75201eb0698a571398005f


nglorioso:
CLA:yes from #3891's acceptance

Issue #3895, opened 2020-05-08
More rounding methods: {log, BigInteger, BigDecimal} -> double


lowasser:
https://stackoverflow.com/q/61669813/869736 asked for long-to-double with nonstandard rounding modes.  Seems potentially worth providing -- not a lot of API surface area, hard to do, fits well with the existing utilities -- though `BigDecimalMath` would be new.

PR #3902, opened 2020-05-14
More rounding methods: {long, BigInteger, BigDecimal} -> double

dhruvdeveloper:
Created new class BigDecimalMath for More rounding methods: {long, BigInteger, BigDecimal} -> double. 
Issue: #3895


lowasser:
We're going to be doing this with an API surface analogous to the other rounding methods in Guava, all of which accept a `java.math.RoundingMode` parameter.  This work has already started in https://github.com/google/guava/commit/d01a196043e7d1620f2880da5d35a31da9acd898.

PR #3920, opened 2020-06-04
Reuse ThreadFactory in MoreExecutors

dennyac:
Currently we create a new ThreadFactory instance whenever we call ```MoreExecutors.newThread()```

Just changing it to reuse the same instance. The only usage of ```platformThreadFactory()``` is in ```MoreExecutors.newThread()``` and they set the name of the thread in that method.

### DIFF
--- a/guava/src/com/google/common/util/concurrent/MoreExecutors.java
+++ b/guava/src/com/google/common/util/concurrent/MoreExecutors.java
@@ -63,6 +63,9 @@ import java.util.concurrent.TimeoutException;
  */
 @GwtCompatible(emulated = true)
 public final class MoreExecutors {
+
+  private static final ThreadFactory DEFAULT_THREAD_FACTORY = Executors.defaultThreadFactory();
+
   private MoreExecutors() {}
 
   /**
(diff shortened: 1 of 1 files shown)


netdpb:
The instance returned by `Executors.defaultThreadFactory()` can be functionally different after `System.setSecurityManager()` is called, and in any case the thread names are different for different returned objects.

While maybe rare, these are not insignificant bits of configuration.

What's the value of making this change, compared to that not well known but nonzero cost? I don't think anyone's worried about performance aspects of creating thread factories.

dennyac:
@netdpb - The threadName is always overwritten in ```MoreExecutors.newThread()``` which is the only place it is used.

The ```SecurityManager``` is only used to ```getThreadGroup()``` which by default, returns ```Thread.currentThread().getThreadGroup()``` which is what is used in ```Executors.defaultThreadFactory()```

But, if someone sets the ```SecurityGroup``` and overrides ```getThreadGroup()```, then functionality will change for threads created after that. Not sure how common that is.

netdpb:
`platformThreadFactory()` is public. It's not only used from `newThread()`. This would change its behavior for all users.

dennyac:
Woah, I totally missed that! Feel free to close this PR.

PR #3934, opened 2020-06-15
Replace anonymous class with lambda expression

bhawna94:


bhawna94:
@netdpb @cpovirk 
please review the PR and let me know your thoughts.

nick-someone:
Hello! Thanks for looking to contribute.

However, Guava has two builds (`android/` and `guava/`), the first of which requires Java 7: https://github.com/google/guava/blob/e839f94fb4eb1240f10763586aa92a7f701caa89/android/pom.xml#L119

We tend to avoid changes that create unnecessary drift between the Java 7 and Java 8 versions, which this would fall under (see the Java 7 version here: https://github.com/google/guava/blob/e839f94fb4eb1240f10763586aa92a7f701caa89/android/guava/src/com/google/common/base/Optional.java#L295). So, for now, I will close this PR.

Issue #3467, opened 2019-05-10
ImmutableTable.get results in copyOf with actual copy being invoked


victornoel:
I faced a strange performance situation with `ImmutableTable`.

Here is a stacktrace of what's happening (it's a profiling trace from [async-profiler](https://github.com/jvm-profiling-tools/async-profiler)):
![image](https://user-images.githubusercontent.com/160975/57542430-f7e47300-7351-11e9-83a6-10f8c42191b9.png)

Basically, it seems that calling `ImmutableTable.get()` results in `ImmutableMap.copyOf()` being called, and even though there is a comment in `rowMap()` saying that `copyOf` is called only for casting (https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/SparseImmutableTable.java#L97), in practice, there is an actual copy via `toArray()`.

I believe this is not meant to happen, is it?

victornoel:
For the record, the `ImmutableTable` was created using `ImmutableTable.toImmutableTable`

samukce:
Hello, I was looking at this and I could understand why this is happening for `get` method. Basically, the classes that:
1. Inherits from AbstractTable;
2. It doesn't override the `V get` method;
3. It has in `rowmap()` implementation the use of `copyOf` or `ImmutableMap.of`

So based on those criteria I could find those classes that have this same behavior of creating news maps internally:
- SparseImmutableTable
- SingletonImmutableTable
- DenseImmutableTable

So, trying to summarize my founds: every use of rowMap()(+columnMap()) method in AbstractTable class it's a potential point of creating extra objects that it seems not necessary for such methods implementations such as: `containsRow, containsColumn, containsValue, contains, get` for ImmutableTable implementations;

My initial thought about it is that the `rowMap` and `columnMap` method it has been burdened with responsibilities: 
1. to provide external data (which can be mutable or immutable data) 
2. to provide support internally to access the data.

That said I can see this possible approach: To have an internal way to access the `rowMap/columnMap` in a protected class scope from ImmutableTable class so we could have those methods been implemented by this one, avoiding create unnecessary objects; 

Something like 
```java
public abstract class ImmutableTable<R, C, V> extends AbstractTable<R, C, V> {
....
  protected abstract ImmutableMap<R, ImmutableMap<C, V>> immutableRowMap();

  @Override
  public boolean containsRow(@Nullable Object rowKey) {
    return Maps.safeContainsKey(immutableRowMap(), rowKey);
  }
...
}

[...]

PR #3940, opened 2020-06-20
Add tests to guarantee no need copyOf map

samukce:
I added tests to guarantee that unnecessary maps are not created given some possibles performance issues mentioned here #3467. Though I was not able to create a test scenario to simulate the issue. So, opening this PR to have at least some explicit tests narrowing this behavior.


samukce:
hello @cpovirk, do you think that is worth to merge those tests? otherwise, we can close the PR.

samukce:
@cpovirk do you think it's worthy to add those tests or should we close it?

PR #3974, opened 2020-07-25
Removed some redundant 'else's using early return

pradipta:


Archish27 (approved):
LGTM 🚀

netdpb:
Hi @pradipta. Many of these look good, and we'll push out a commit that incorporates them, attributed to you, soon. We're not taking the ones where the else branch is of roughly the same size as the if branch, because those seem to have more parallelism.

Issue #3992, opened 2020-08-20
Document that primitive event types are not supported, and throw an exception when they're encountered


netdpb:
#3989 raises the issue that `@Subscribe` methods with primitive parameter types can be registered but they will never be called because the `post()` method takes `Object` and so autoboxes primitives.

`EventBus` should throw an exception preventing registering any `@Subscribe` methods with primitive parameters and document the restriction.

PR #3998, opened 2020-08-25
MOE Sync 2020-08-25

netdpb:

Commits:
=====
<p> Throw if @Subscribe is applied to a method that takes a primitive parameter.

Fixes https://github.com/google/guava/issues/3992.

RELNOTES=Prevent @Subscribe being applied to a method that takes a primitive, as that will never be called.

9fe4ecdc84e628c7da0aa68fc6257932dc9dbced

### DIFF
--- a/guava/src/com/google/common/eventbus/SubscriberRegistry.java
+++ b/guava/src/com/google/common/eventbus/SubscriberRegistry.java
@@ -16,6 +16,7 @@ package com.google.common.eventbus;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Throwables.throwIfUnchecked;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.MoreObjects;
@@ -31,6 +32,7 @@ import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Multimap;
+import com.google.common.primitives.Primitives;
 import com.google.common.reflect.TypeToken;
 import com.google.common.util.concurrent.UncheckedExecutionException;
 import com.google.j2objc.annotations.Weak;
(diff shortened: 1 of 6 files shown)


guava/src/com/google/common/eventbus/SubscriberRegistry.java
```diff
@@ -183,11 +190,20 @@ void unregister(Object listener) {
...
               parameterTypes.length);
 
+          checkArgument(
+              !parameterTypes[0].isPrimitive(),
+              "@Subscribe method %s's parameter is %s. "
+                  + "Subscriber methods cannot accept primitives. "
+                  + "Consider changing the parameter to %s.",
+              method,
+              parameterTypes[0].getName(),
+              Primitives.wrap(parameterTypes[0]).getSimpleName());
```
kilaka, line 205: Hi :)

Can we somehow postpone the string concatenation and param wrapping for only when the argument is illegal?

For example, what log4j do.
Or by using a lambda (I know lambdas aren't welcome on Guava. See one more option at the bottom), like:

```
Object o = lazyToString(() -> someLongCalculatedString());

public static Object lazyToString(Supplier<String> supplier)
{
	return new Object()
	{
		@Override
		public String toString()
		{
			return supplier.get();
		}
	};
}

```

Another option is to just check the argument inline and call a specific method to throw the `IllegalArgumentException`.
Example:
```
if (parameterTypes[0].isPrimitive()) {
  throwIllegalArgumentException(
          "@Subscribe method %s's parameter is %s. "
              + "Subscriber methods cannot accept primitives. "
              + "Consider changing the parameter to %s.",
          method,
          parameterTypes[0].getName(),
          Primitives.wrap(parameterTypes[0]).getSimpleName());
}
```

What do you think?
@netdpb 

Cheers.

kilaka: Come to think about it, 3rd option looks even more readable than `checkArgument()`.

cpovirk:
I would have to look at bytecode to confirm, but I don't think there's any allocation. Concatenation of literally happens at compile time, and checkArgument has non-varargs overloads: https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/base/Preconditions.html#checkArgument-boolean-java.lang.String-java.lang.Object-java.lang.Object-java.lang.Object-

cpovirk:
(It remains a fair point that we use checkArgument habitually, even in some cases where a manual check and throw would be better.)

cpovirk:
(I do hope that we'll one day add methods like illegalArgument("foo %s", bar).)

kilaka:
I like the "shortness" of `illegalArgument("foo %s", bar).)`.
I'd like it to be clear that it throws an exception.

I think something of the sorts looks very readable:
```
if (checkFails) {
  throw illegalArgument("foo %s", bar);
}
```
`illegalArgument` just builds the exception with the string and whatever formatting logic it applies.

@cpovirk

PR #4008, opened 2020-09-08
MOE Sync 2020-09-08

nick-someone:

Commits:
=====
<p> Minor internal change

3a85bdca141a1c5c2bb29f06fa8f99c706e46ee8


cpovirk:
This may pass normal tests, but it will fail the GWT tests (which we run only at release time). The right thing for us to do will be to edit the MOE config to add this new GWT inherits line to the list of those to be stripped.

nick-someone:
Sounds good to me. I'll pull back this PR and send you the change internally.

cpovirk:
Thanks. (Also: I am not "in the office" yet, but I saw that some bugs came in on Friday, and I'll triage those soon if you haven't already.)

Issue #4002, opened 2020-08-30
UnsupportedOperationException when modify TreeRangeSet.subRangeSet(Range)


zhanhb:
Unable to modify subRangeSet returned by TreeRangeSet.subRangeSet(Range).
IMO, it should be availiable to do so as specified in the document.

```java
import com.google.common.collect.Range;
import com.google.common.collect.TreeRangeSet;
import org.junit.jupiter.api.Test;

@SuppressWarnings("UnstableApiUsage")
public class SubRangeSetTest {

    @Test
    public void testAdd() {
        TreeRangeSet<Integer> set = TreeRangeSet.create();
        Range<Integer> range = Range.closedOpen(0, 5);
        set.subRangeSet(range).add(range);
    }

    @Test
    public void testReplaceAdd() {
        TreeRangeSet<Integer> set = TreeRangeSet.create();
        Range<Integer> range = Range.closedOpen(0, 5);
        set.add(range);
        set.subRangeSet(range).add(range);
    }
}
```

```
java.lang.UnsupportedOperationException
	at java.base/java.util.AbstractMap.put(AbstractMap.java:209)
	at com.google.common.collect.TreeRangeSet.replaceRangeWithSameLowerBound(TreeRangeSet.java:266)
	at com.google.common.collect.TreeRangeSet.add(TreeRangeSet.java:216)
	at com.google.common.collect.TreeRangeSet$SubRangeSet.add(TreeRangeSet.java:894)
	at SubRangeSetTest.testAdd(SubRangeSetTest.java:12)
```
```
java.lang.UnsupportedOperationException
	at com.google.common.collect.UnmodifiableIterator.remove(UnmodifiableIterator.java:46)
	at com.google.common.collect.Iterators.clear(Iterators.java:985)
	at com.google.common.collect.Maps$IteratorBasedAbstractMap.clear(Maps.java:3562)
	at com.google.common.collect.TreeRangeSet.add(TreeRangeSet.java:214)
	at com.google.common.collect.TreeRangeSet$SubRangeSet.add(TreeRangeSet.java:894)
	at SubRangeSetTest.testReplaceAdd(SubRangeSetTest.java:20)
```

cpovirk:
Thanks, that sounds like a bug to me, too.

I see a slightly different stack trace (with both guava-29.0-jre.jar and guava-29.0-android.jar), so the exact failure may depend on the Guava version:

```
# -jre
Exception in thread "main" java.lang.UnsupportedOperationException
[...]

PR #4019, opened 2020-09-15
Issue #4002 UnsupportedOperationException when modify TreeRangeSet.subRangeSet(Range).

Koooooo-7:
The same implementation as `remove`.


Zenix27 (approved):

Kraisornbiinthi (commented):
Hell

PR #4021, opened 2020-09-15
Fixed Typos

nishantc1527:
I found multiple typos in comments and class names, so I fixed it.


lowasser:
According to dictionary.com, it is in fact "interruptible," not "interruptable."

nishantc1527:
Thanks, I changed it back.

Issue #3679, opened 2019-11-04
Various cast() methods that work around JDK-6558557 are no longer necessary


aepshteyn:
https://github.com/google/guava/blob/49f5a6332a63737dff70cf77472f9867bc7ca6eb/guava/src/com/google/common/collect/Collections2.java#L355-L358

It's probably not worth the effort to refactor all the callers of this method, but I just wanted to bring to your attention that [JDK-6558557](https://bugs.java.com/bugdatabase/view_bug.do?bug_id=6558557) was fixed in Java 1.7, so if that's the only reason for that method to exist, it's no longer necessary.  

That's all 🙂

netdpb:
Thanks!

Saucistophe:
I'm having trouble with the CLA, but the PR is here.

cpovirk:
Reopening, as I've noticed that we still have similar methods in classes other than `Collections2`:
https://github.com/search?q=6558557+repo%3Agoogle%2Fguava+path%3A%2Fguava&type=Code&ref=advsearch&l=&l=

tanmauec:
As discussed with @cpovirk, I am taking this up.

PR #4034, opened 2020-09-19
google#1409 Add poll() and offer() to Uninterruptibles class

tanmauec:


tanmauec:
Any updates?

cpovirk:
I still probably won't have a chance for a while, but it's on my to-do list.

tanmauec:
> I still probably won't have a chance for a while, but it's on my to-do list.

Works bro, I don't have slightest clue how full is your plate 👍

tanmauec:
@cpovirk Can you please review and possibly close this!

cpovirk:
It's still on my list, just still behind some other things :(

tanmauec:
@cpovirk  Sorry again, I was waiting for my first PR accepted, that has happened so I am fine!

Please suggest some feature development problems to work on..

cpovirk:
Sorry for the delay on the other one, which was simple enough that I should have accepted it more quickly.

I'll let you know if I think of some more ideas.

tanmauec:
Sure Ack.

tanmauec:
> Sorry for the delay on the other one, which was simple enough that I should have accepted it more quickly.
> 
> I'll let you know if I think of some more ideas.

@cpovirk any issues/features you suggest?

cpovirk:
The one thing that comes to mind is a Dependabot config (like [this one](https://github.com/google/compile-testing/blob/master/.github/dependabot.yml)) for https://github.com/google/caliper. But before doing that, we'd really like to update all our dependencies (`mvn org.codehaus.mojo:versions-maven-plugin:2.8.1:update-properties org.codehaus.mojo:versions-maven-plugin:2.8.1:use-latest-releases`, similar to [this commit](https://github.com/google/auto/commit/14bb4c20c792b0ea16dde838ae10e9329f05cda4)) and all our plugins (which may be easiest to do [by hand](https://stackoverflow.com/a/34032422/28465)).

cpovirk:
Also https://github.com/google/guava/issues/3679

tanmauec:
@cpovirk Sure will take up #3679 first as it seems quick fix.

cpovirk (commented):
Sorry for taking so long on this. The main thing that these implementations will need to do is to track how much time has passed. Currently, they always pass the original timeout to the delegate. But when that call is interrupted, they should pass only how much time is left. You can see how to set that up here;

https://github.com/google/guava/blob/ce71b12f50a1de6b4fbaec67b02c911e3cef335e/android/guava/src/com/google/common/util/concurrent/Uninterruptibles.java#L81

tanmauec:
@cpovirk I have updated offer, poll functions adding timeout update as required. Please check.

PR #4036, opened 2020-09-23
Remove unnecessary parens

T45K:
This PR is related to PR #3191.

In #3191, unnecessary parens in `PairedStats` were omitted, but its similar class, `Stats` was not touched.
I found unnecessary parens in the class, so removed.


T45K:
@googlebot I signed it!

Issue #3680, opened 2019-11-05
Remove GWT-RPC support


cpovirk:
In January, [we announced that we'd remove it](https://groups.google.com/d/topic/guava-announce/zHZTFg7YF3o/discussion) at the beginning of 2020. We've already laid some groundwork for that.

I'm filing this issue as an additional way of raising awareness. I'll try to also mention it from relevant CLs.

cpovirk:
This is coming in the next release (in a commit to be mirrored out later this week).

cpovirk:
(https://github.com/google/guava/commit/5214a10b1eac7a1f7267ee6ed5cf82c9cd76ad0b)

PR #5267, opened 2020-10-08
MOE Sync 2020-10-08

netdpb:

Commits:
=====
<p> Run all zero of our GWT tests during presubmit/CI.

Unfortunately, this does *not* detect any bogus <inherits> lines in *test* .gwt.xml files (neither the individual testModule.gwt.xml files nor the top-level GuavaTests.gwt.xml).
(And bogus <inherits> lines in *prod* .gwt.xml files were already detected even before this change.)

As a result, this CL appears to be very nearly a no-op -- maybe just a verification that the no-op GWT module and entry point exist. And maybe I'll end up rolling it back if we reenable the tests externally someday. Still, for now it seems like a reasonable thing to clean up, if only to prevent me from trying to clean it up again 6 months from now :)

(Followup after https://github.com/google/guava/issues/3680)

9230b14efefc11d24c3c969f2dbae7f8595ef469

-------

<p> Check the generated PublicSuffixPatterns.java into Piper.

Copybara can't export generated files.

Also, reformat the template (and then the file) so that autoformatting the file doesn't produce diffs.

eebb75ecc2beb4ff5c95cca974d3529505e31a60

-------

<p> Omit TestModuleEntryPoint from open-source repo.

Since I've disabled GWT tests in the open-source repo, these no longer serve a purpose. (They're still listed in our testModule.gwt.xml files, but as noted in a previous commit, the GWT test infrastructure doesn't seem to care what we put in those files anymore.)

We could still include them, but they're generated during our build, so that complicates exporting them with Copybara.

5948a59b546c02d61ac9c2f590483fb4a7271ddd


guava-gwt/pom.xml
```diff
@@ -375,7 +373,7 @@
             </goals>
             <configuration>
               <module>com.google.common.GuavaTests</module>
-              <includes>${gwt.test.include}</includes>
+              <includes>**/GwtTestSuite.java</includes>
               <logLevel>${gwt.logLevel}</logLevel>
               <mode>htmlunit</mode>
               <htmlunit>FF38</htmlunit>
```
cpovirk, line 379: Oops: The move of `TestModuleEntryPoint` below (which I can't seem to comment on directly) shouldn't happen: This commit should include only the `pom.xml` changes.

The `TestModuleEntryPoint` move is happening because MOE is reading the latest copy of its config from head, and that version of the config no longer has special handling for `TestModuleEntryPoint`. Ideally, MOE would use the version of its config at CL 335869875 for this change and use the new version of its config only for later CLs.

I suspect that `guava-testlib` won't actually build successfully at this specific commit. If you're feeling ambitious, you could amend the commit to undo the `TestModuleEntryPoint` changes. But then you might have to amend the later commit that removes `TestModuleEntryPoint` to remove the copies in their corrected location.... And probably no one will ever care: It's not like we're going to cut a release at this specific commit. So I am mainly just noting this in case someone does discover the problem later.

Someone should migrate us to Copybara :) (which hopefully doesn't have this problem...).

netdpb: So you're saying this commit moves files to a bad location, but two commits later the files are deleted anyway?

I shall choose to look the other way, then.

PR #5268, opened 2020-10-09
MOE Sync 2020-10-09

### LINKED ISSUE #3679: Various cast() methods that work around JDK-6558557 are no longer necessary

### LINKED ISSUE #3682: MultiMap concurrency unclear Javadoc
From https://github.com/google/guava/issues/135#issuecomment-472966705

The [javadoc for HashMultimap](https://google.github.io/guava/releases/27.0-jre/api/docs/com/google/common/collect/HashMultimap.html)  states:  

> This class is not threadsafe when any concurrent operations update the multimap. Concurrent read operations will work correctly.

Does this mean that it is thread-safe if only one thread writes to the map (while any number of threads may read simultaniously) ? The javadoc for [other MultiMap implementations](https://google.github.io/guava/releases/27.0-jre/api/docs/com/google/common/collect/ArrayListMultimap.html) is the same and i think it should be clearer on that.

netdpb:

Commits:
=====
<p> Clarify `HashMultimap` docs about concurrency.

Closes #5266.
Fixes #3682.

e6b523481b34b2c2007046ea987bf62df92eb0f8

-------

<p> Removed all references to now useless workaround

Fixes #3679
Fixes #5265

26da0949c1cf54eb730578e75fe087f67110ee71

### DIFF
--- a/android/guava/src/com/google/common/collect/Collections2.java
+++ b/android/guava/src/com/google/common/collect/Collections2.java
@@ -329,11 +329,6 @@ public final class Collections2 {
     return new StringBuilder((int) Math.min(size * 8L, Ints.MAX_POWER_OF_TWO));
   }
 
-  /** Used to avoid http://bugs.sun.com/view_bug.do?bug_id=6558557 */
-  static <T> Collection<T> cast(Iterable<T> iterable) {
-    return (Collection<T>) iterable;
-  }
-
   /**
    * Returns a {@link Collection} of all the permutations of the specified {@link Iterable}.
    *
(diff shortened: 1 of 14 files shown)


netdpb:
These commits were derived from the PRs by @Saucistophe, so I am marking the CLA as signed.

PR #5271, opened 2020-10-12
Adding ppc64le architecture support on travis-ci

kishorkunal-raj:
Hi,
I have added support for ppc64le architecture support on travis-ci in the branch. I believe it is ready for the final review and merge.
The travis-ci build log can be tracked on the link : https://travis-ci.com/github/kishorkunal-raj/guava/builds/189450395
Please have a look on it

Thanks !!


cpovirk:
Thanks. I haven't looked in detail at the logs, but I see a timeout in the Travis run here. Maybe we need `travis_wait` for the `install` command, similar to what we have here...?
https://github.com/google/guava/blob/be1c0d8a4fd00e6c5cc362cc0021db5e9e3d4fef/.travis.yml#L17

cpovirk:
(We've wanted a convenient way to test on PPC a few times in the past, so this could be handy.)

cpovirk:
Sorry, I'm not sure that I realized that this PR had gotten updated after the earlier discussion.

Nowadays, it would have to be updated for GitHub CI. I don't know if anyone is interested in giving that a shot, but we should be able to accept a change if someone is.

cpovirk (approved):
I'm going to approve this as a test, even though we'd need to rework it to use GitHub CI if we wanted to merge it.

PR #5287, opened 2020-10-17
#3951 - Joiner implements Collector

### LINKED ISSUE #3951: Provide Joiner as Collector
The main advantage of Guava `Joiner` over JDKs `Collectors.joining(...)` or `StringJoiner` that it can handle arbitrary objects and has a convenient builder api to express e.g. `useForNull(nullText)`.

Thus allow the usage of Joiner when streams come into play and get rid of cumbersome expressions like `filter(Ojects::nonNull).map(Object::toString)`.

E.g. with factory method _joiningOn(...)_:
```
objectStream.collect(joiningOn(", ").skipNulls());
objectStream.collect(joiningOn(", ").useForNull("<N/A>"));
```

Alternatively `Joiner` could implement Collector interface.
```
objectStream.collect(Joiner.on(", ").skipNulls());
objectStream.collect(Joiner.on(", ").useForNull("<N/A>"));
```

Saucistophe:
Fixes #3951

### DIFF
--- a/guava-tests/test/com/google/common/base/JoinerTest.java
+++ b/guava-tests/test/com/google/common/base/JoinerTest.java
@@ -26,11 +26,10 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.testing.NullPointerTester;
 import java.io.IOException;
-import java.util.Arrays;
-import java.util.Iterator;
-import java.util.Map;
+import java.util.*;
 import java.util.Map.Entry;
-import java.util.Set;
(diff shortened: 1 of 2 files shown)


nick-someone:
Thank you for providing an implementation of Joiner as a Collector! However, we're still not sure that this is a good idea yet (e.g.: should any arbitrary Joiner be a Collector (or should we provide different collector interfaces), how do we handle Java 8 vs Android, etc.).

I'm going to close this PR for now, as there likely to be a number of design choices to work out that may obviate most of the code here.

PR #5288, opened 2020-10-17
#5284 - Fixed different result when closing encoding stream twice

### LINKED ISSUE #5284: BaseEncoding.encodingStream().close() is non-idempotent
[`Closeable.close`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/Closeable.html#close()) says:
> If the stream is already closed then invoking this method has no effect.

Code:
```
StringWriter w = new StringWriter();
OutputStream out = BaseEncoding.base64().encodingStream(w);
out.write(0);
out.close();
System.out.println(w);
out.close();
System.out.println(w);
```
Output:
```
AA==
AA==A===
```

Saucistophe:
Fixes #5284

### DIFF
--- a/android/guava/src/com/google/common/io/BaseEncoding.java
+++ b/android/guava/src/com/google/common/io/BaseEncoding.java
@@ -620,6 +620,7 @@ public abstract class BaseEncoding {
             int charIndex = (bitBuffer << (alphabet.bitsPerChar - bitBufferLength)) & alphabet.mask;
             out.write(alphabet.encode(charIndex));
             writtenChars++;
+            bitBufferLength = 0;
             if (paddingChar != null) {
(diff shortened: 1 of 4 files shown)


android/guava/src/com/google/common/io/BaseEncoding.java
```diff
@@ -627,6 +627,7 @@ public void close() throws IOException {
               }
             }
           }
+          bitBufferLength = 0;
```
nick-someone, line 630: I'd move this just after the increment to writtenChars (so, just in case the write of the padding character fails and someone tries to re-close the stream anyway, we get it correctly)

Saucistophe: Done!

CertifiedRice (approved):

Issue #3349, opened 2018-12-21
ClassPath.getTopLevelClasses() does not return a top-level class with $ character


suztomo:
@netdpb and I found that [ClassPath.getTopLevelClasses()](https://google.github.io/guava/releases/26.0-jre/api/docs/com/google/common/reflect/ClassPath.html#getTopLevelClasses--) does not return top-level class with $ character in its name. Example class: `com.google.cloud.bigquery.$AutoValue_Labels` in google-cloud-bigquery-1.56.0.jar.

Test case
```
  @Test
  public void testGuavaTopLevelClass() throws Exception {
    // google-cloud-bigquery-1.56.0.jar contains com.google.cloud.bigquery.$AutoValue_Labels
    URL jarFileUrl =
        new URL(
            "file:///home/suztomo/.m2/repository/com/google/cloud/google-cloud-bigquery/1.56.0/google-cloud-bigquery-1.56.0.jar");
    URL[] jarFileUrls = new URL[] {jarFileUrl};

    URLClassLoader classLoaderFromJar = new URLClassLoader(jarFileUrls, null);

    com.google.common.reflect.ClassPath classPath =
        com.google.common.reflect.ClassPath.from(classLoaderFromJar);

    ImmutableList<String> allClassNames =
        classPath.getAllClasses().stream()
            .map(classInfo -> classInfo.getName())
            .collect(toImmutableList());

    // This code does not return class com.google.cloud.bigquery.$AutoValue_Labels
    // Guava only checks the existence of '$'
    // https://github.com/google/guava/blob/master/guava/src/com/google/common/reflect/ClassPath.java#L85
    ImmutableList<String> topLevelClassNames =
        classPath.getTopLevelClasses().stream()
            .map(classInfo -> classInfo.getName())
            .collect(toImmutableList());

    String class$AutoValue_Labels = "com.google.cloud.bigquery.$AutoValue_Labels";
    System.out.println("In all classes? " + allClassNames.contains(class$AutoValue_Labels)); // true
    System.out.println(
        "In top-level classes? " + topLevelClassNames.contains(class$AutoValue_Labels)); // false. It should be true.
  }

```

[JLS 3.8](https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.8) states that dollar sign ('$') is a valid Java letter.

AutoValue has `@Memoized` annotation, which creates class `$AutoValue_XXXX` and `AutoValue_XXXX extends $AutoValue_XXXX` in the same package.

[...]

PR #5293, opened 2020-10-24
MOE Sync 2020-10-24

nick-someone:

Commits:
=====
<p> Add ClassInfo.isTopLevel().

Seems like a useful method? I can't remember why I didn't add it in the first place. Perhaps just an oversight.

Partially addresses https://github.com/google/guava/issues/3349

RELNOTES=`reflect`: Added `ClassInfo.isTopLevel()`.

14173d1bd5997fab0da2e8636e3c3a78f8e7cd61

### DIFF
--- a/android/guava/src/com/google/common/reflect/ClassPath.java
+++ b/android/guava/src/com/google/common/reflect/ClassPath.java
@@ -78,14 +78,6 @@ import org.checkerframework.checker.nullness.compatqual.NullableDecl;
 public final class ClassPath {
   private static final Logger logger = Logger.getLogger(ClassPath.class.getName());
 
-  private static final Predicate<ClassInfo> IS_TOP_LEVEL =
-      new Predicate<ClassInfo>() {
-        @Override
-        public boolean apply(ClassInfo info) {
-          return info.className.indexOf('$') == -1;
-        }
-      };
-
   /** Separator for the Class-Path manifest attribute value in jar files. */
   private static final Splitter CLASS_PATH_ATTRIBUTE_SEPARATOR =
       Splitter.on(" ").omitEmptyStrings();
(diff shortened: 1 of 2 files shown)


guava/src/com/google/common/reflect/ClassPath.java
```diff
@@ -324,6 +328,18 @@ public String getName() {
       return className;
     }
 
+    /**
+     * Returns true if the class name "looks to be" top level (not nested), that is, it includes no
+     * '$' in the name, This method may return false for a top-level class that's intentionally
```
Stephan202, line 333:
```suggestion
     * '$' in the name. This method may return false for a top-level class that's intentionally
```

guava/src/com/google/common/reflect/ClassPath.java
```diff
@@ -324,6 +328,18 @@ public String getName() {
...
+     * named with the '$' character. If ths is a concern, you could use {@link #load} and then check
```
jbduncan, line 334:
```suggestion
     * named with the '$' character. If this is a concern, you could use {@link #load} and then check
```

android/guava/src/com/google/common/reflect/ClassPath.java
```diff
@@ -324,6 +328,18 @@ public String getName() {
       return className;
     }
 
+    /**
+     * Returns true if the class name "looks to be" top level (not nested), that is, it includes no
+     * '$' in the name, This method may return false for a top-level class that's intentionally
+     * named with the '$' character. If ths is a concern, you could use {@link #load} and then check
```
ineuwirth, line 334: Typo in `ths`.

nick-someone:
Thanks for the comments! I'll fix this up internally and it'll end up on the next push out tomorrow.

PR #5304, opened 2020-11-07
Issue 1673 adding upper hyphen type case format

malasemi:
Adding upper hyphen type case format


malasemi:
@googlebot I signed it!

netdpb:
Thanks! Do you have a particular use case for upper-hyphen?

PR #5306, opened 2020-11-08
Add `X-Request-ID` to `HttpHeaders`

TiagoGrosso:

PR #5307, opened 2020-11-10
MOE Sync 2020-11-10

cgdecker:

Commits:
=====
<p> Refactor ClassPath: replace Scanner with Location class.

75c2c543c6ac64943037630cfa40945feff10f42

-------

<p> Fix documentation for HashCode.toString() that incorrectly claimed it returned in big-endian order.

The implementation of toString() goes through the bytes in the same order that asBytes() returns, so if the order of the underlying bytes is little-endian (as is the case for LongHashCode), then the string will also be little-endian, contradicting what the documentation says.

841129afbd77f8c78bb469de2ece758b418d322b

-------

<p> Optimize Hashing.crc32c() by eliminating two xors for every byte of input.

Benchmarks suggest a modest speedup around 8%.

b685bf7068044cfef0c73163408a0ccebacef203

-------

<p> Add constant for Origin Isolation header

Spec: https://github.com/WICG/origin-isolation

RELNOTES=Adds constant for the Origin-Isolation header.

f3d79119d1e1f9937f332a8c73ed1690dfbb8ffb

-------

<p> Remove outdated reference in docs

3ac6e42a8622be4b122dfd4b9eef989148141148

-------

<p> Add additional test for `Crc32cHashFunction`.

4f7e925397e6ca2b17ce3d563ba1968e94959bd1

-------

<p> Verify maximumWeight before setting

The maximumWeight builder method is the only method that does the argument verification before the assignment.
It looks like a typo, and does not affect anyone in practice, but it is still better to have the implementation uniform.

Fixes #5298

f1d40c3dd5243c5a29fe683fb42f7ff61b2a3068

-------

<p> Do 16-byte strides in CRC32C in c.g.c.hash, resulting in roughly a 3.3x speedup to 1.5 GB/s.

This is probably as good as it gets until Java 11, when we'll have access to the intrinsics that use CRC hardware when available.

6d9e9e33c6a784f86b51c3fac785532351374271

-------

<p> Rollback

Do 16-byte strides in CRC32C in c.g.c.hash, resulting in roughly a 3.3x speedup to 1.5 GB/s.

This is probably as good as it gets until Java 11, when we'll have access to the intrinsics that use CRC hardware when available.

***

916807371db167b99bb8b591a22c9ec07fe2bac2

-------

<p> Backport `combine` to java7 for `ImmutableList` & `ImmutableSet`.

This is so that Java8 code relying on the Java7 branch of Guava can still efficiently collect Guava collections.

a275458c2601c33bfb5ba6ec54e2458feeeff06a

-------

<p> Fix `LoadingCache` minor documentation typo.

Fixes #5503.

59a9556d82fc233acf8c0a3ad00de14e02384f75

-------

<p> Added submitAsync to ClosingFuture.

RELNOTES="Added `submitAsync(AsyncClosingCallable)` to `ClosingFuture`."

187db6d240a896be6efb6fc98f014b9d4690fe30

-------

<p> Add `X-Request-ID` to `HttpHeaders`.

Fixes #5306.

RELNOTES=Add `X-Request-ID` to `HttpHeaders`.

0f5a31bbc213f61404189d777b2039842210cab5

-------

<p> Fix Android test failure in ClassPathTest.

Reproduced the error and verified that this change fixes it.

c1f6ac185f0e72998ed30ae72a0983804c5a4f5a


cgdecker:
Setting `cla: yes` since all commits with different authors come directly from PRs by those authors.

PR #5324, opened 2020-11-19
Fix #4011: Files::createTempDir security vulnerability

### LINKED ISSUE #4011: CVE-2020-8908: Files::createTempDir local information disclosure vulnerability
# IMPORTANT NOTE

**Updating to Guava 30.0 does not fix this security vulnerability**. The method is merely deprecated. There currently exits no fix for this vulnerability. 

https://github.com/google/guava/issues/4011#issuecomment-765672282

---

Since the fix for this vulnerability is now disclosed by this commit (https://github.com/google/guava/commit/fec0dbc4634006a6162cfd4d0d09c962073ddf40) and it was closed internally by google as 'Intended Functionality' I figure I'll disclose the vulnerability fully.

## Vulnerability

```java
File guavaTempDir = com.google.common.io.Files.createTempDir();
System.out.println("Guava Temp Dir: " + guavaTempDir.getName());
runLS(guavaTempDir.getParentFile(), guavaTempDir); // Prints the file permissions -> drwxr-xr-x
File child = new File(guavaTempDir, "guava-child.txt");
child.createNewFile();
runLS(guavaTempDir, child); // Prints the file permissions -> -rw-r--r--
```

On the flip side, when using `java.nio.file.Files`, this creates a directory with the correct file permissions.
```java
Path temp = Files.createTempDirectory("random-directory");
System.out.println("Files Temp Dir: " + temp.getFileName());
runLS(temp.toFile().getParentFile(), temp.toFile()); // Prints the file permissions -> drwx------
Path child = temp.resolve("jdk-child.txt");
child.toFile().createNewFile();
runLS(temp.toFile(), child.toFile()); // Prints the file permissions -> -rw-r--r--
```

## Impact

The impact of this vulnerability is that, the file permissions on the file created by `com.google.common.io.Files.createTempDir` allows an attacker running a malicious program co-resident on the same machine can steal secrets stored in this directory. This is because by default on unix-like operating systems the `/temp` directory is shared between all users, so if the correct file permissions aren't set by the directory/file creator, the file becomes readable by all other users on that system.

## Workaround

This vulnerability can be fixed by explicitly setting the `java.io.tmpdir` system property to a "safe" directory when starting the JVM.

## Resolution

The resolution by the Google team was the following:

> The team decided to document the behavior, as well as deprecate the method as other alternatives exist.

This completely makes sense to me, and I think is appropriate. The open question that exists in my mind is whether or not this issue warrants a CVE number issued.

melloware:
I am submitting this PR because...

- this method is marked both `Beta` and `Deprecated`.
- This change still has the code work correctly it just uses NIO which does not have the vulnerability
- Allows security companies like Sonatype and Snyk to stop reporting this as vulnerable

### DIFF
--- a/guava/src/com/google/common/io/Files.java
+++ b/guava/src/com/google/common/io/Files.java
@@ -421,25 +421,14 @@ public final class Files {
   @Beta
   @Deprecated
   public static File createTempDir() {
-    File baseDir = new File(System.getProperty("java.io.tmpdir"));
-    @SuppressWarnings("GoodTime") // reading system time without TimeSource
-    String baseName = System.currentTimeMillis() + "-";
-
-    for (int counter = 0; counter < TEMP_DIR_ATTEMPTS; counter++) {
-      File tempDir = new File(baseDir, baseName + counter);
-      if (tempDir.mkdir()) {
-        return tempDir;
-      }
+    try {
+      @SuppressWarnings("GoodTime") // reading system time without TimeSource
+      String baseName = System.currentTimeMillis() + "-";
+      java.nio.file.Path temp = java.nio.file.Files.createTempDirectory(baseName);
+      return temp.toFile();
+    } catch (IOException ioex) {
+      throw new IllegalStateException("Failed to create temporary directory.", ioex);
     }
-    throw new IllegalStateException(
-        "Failed to create directory within "
-            + TEMP_DIR_ATTEMPTS
-            + " attempts (tried "
-            + baseName
-            + "0 to "
-            + baseName
-            + (TEMP_DIR_ATTEMPTS - 1)
-            + ')');
   }
 
   /**


melloware:
@googlebot I signed it!

jbduncan:
As an external user of Guava (not part of the team), this looks good to me. 👍

cpovirk:
Thanks. We're hoping that we'll be able to merge this. That said, we won't be able to make the change in our Android versions (since they target a version of Android without `java.nio.file`). So I hope that Sonatype and Snyk will be able to configure their tools to recognize that `35.0-jre`, etc. are safe even when they consider `35.0-android`, etc. to be vulnerable.

melloware:
@cpovirk Yes I did not touch the Android version on purpose in this PR and I can speak for Sonatype it does know the -jre version of the JAR in its detection methods so it will be able to mark this resolved for Guava 31.0-jre if that is the version it ends up in!

cpovirk:
Oh, great, thank you. I was worried that they might depend on being able to specify the vulnerable versions as Maven ranges. But if they can say "all `-android` versions are vulnerable" (or just list all the existing ones individually, or maybe even preemptively list likely future versions), then that should do the trick.

cgdecker:
My current inclination is that we should not make this change.

This change is a behavioral change that has the _potential_ to break existing users if they're intentionally accessing temp directories created with the method from a separate process (communicating the path to another process via some means, say). My expectation is that 99%+ of users are just using temp dirs from the process that creates them, but I don't feel like I can rule out the possibility of this breaking someone.

That wouldn't necessarily be reason enough not to do this, except that I think the current state of things is not that bad. The method is deprecated and has a clear and easy replacement for most users. Given that, I'd want to feel certain that we wouldn't be breaking anyone in order for this change to feel worthwhile.

melloware:
@cgdecker the fact that its still marked Beta means users should not expect it not to change between versions if they are expecting some legacy behavior they can update their code not to use this deprecated method.  

IMHO the Security risk alone warrants this change for all the reasons I stated both in the ticket and above.  Just my two cents.

cpovirk:
A couple other observations from some quick internal testing:

- This fails at least one internal test. That test sets `java.io.tmpdir` to `/invalid` and then expect the `createTempDir` call to fail. That said, I doubt that any actual production code cares about that behavior. Rather, the test was probably just verifying that the correct type of exception was thrown if the operation _did_ fail. Still, it's _possible_ that there's a behavior change here. I didn't look into whether `java.nio.file` was creating the missing `/invalid` parent directory (probably unlikely, since I doubt it has permissions?) or failing to respect the programmatic change to `java.io.tmpdir`. If the latter, someone _might_ care.

- This breaks some of our [J2ObjC](https://developers.google.com/j2objc) users. That said, I was able to mark the method as `@J2ObjcIncompatible`, and tests passed again. In theory, if a J2ObjC user _were_ using `createTempDir`, this would be a breaking change. But that may well never come up.

Still, the points stand that this is `@Beta` and `@Deprecated`.

It's hard to know what would be least disruptive to users :(

JLLeitschuh:
CC. So you can update the Syky advisory
 - Benji @benjifin
 - Leeya @leeyashalti

melloware:
@cpovirk When it comes to security, disruption should be least of concerns.  See SolarWinds hack 2020...

cpovirk:
To try to merge the various discussions about this security vulnerability as a whole into one place, I'm closing this in favor of the discussion on https://github.com/google/guava/issues/4011#issuecomment-769975829. As noted above and in that thread, this PR solves the problem only for the `-jre` variant, and we expect to solve the problem for both variants by removing the method entirely.

melloware:
Awesome lets remove it!

PR #5350, opened 2020-12-09
checkWeightWithWeigher ,Simplify if else

loveincode:
Simplify if else  ,Streamline the code


loveincode:
@googlebot I signed it!

PR #5351, opened 2020-12-09
checkWeightWithWeigher ,Simplify if else

loveincode:
Simplify if else  ,Streamline the code


kluever:
I don't think this PR makes the branching logic any easier to understand.

Issue #5362, opened 2020-12-21
Add checkstyle as dependency


AlbertoSinigaglia:
Hi,  
I was reading the contributions rules, and i came across this:

_Files should be formatted according to [Google's Java style guide](https://google.github.io/styleguide/javaguide.html)._

I looked at the main `pom.xml`, but I can't find any kind of static code analysis, for example to be sore that that rule is complied.

So my question is, are there any specific reason why you are not using [something like this](https://checkstyle.sourceforge.io/google_style.html) to check that constraint?

cpovirk:
Thanks. My guess is that we won't want to set this up, but the reasons are a bit weird:

First: All our changes are submitted to an internal repository first and then synced out to GitHub automatically. The internal repository already has a formatting presubmit in place.

Additionally, there's a complication that comes up when our syncing process makes automated changes to some of our code. For example, we have internal code like the following...

```java
/* MOE:insert private */
void foo() {}
```

...which gets rewritten to...

```java
private
void foo() {}
```

That would make more sense on one line, but if we're going to _require_ it that way, we'll have to clean up some existing code (not that that's too hard), and the process will introduce more diffs between the versions (which makes it slightly harder to import pull requests -- also not a big deal, but the benefit of joining the line in that case is also small). \[edit: It would also require a formatting step during the sync to GitHub, and GitHub sync+CI is already the long pole for most of our submissions.\]

Next, there's your point about existing violations on #5363: Either Checkstyle or its "Google style" config is frequently incorrect. Some of this looks like simple bugs in the config. For example, we don't follow the rule that "Single-line Javadoc comment should be multi-line." Others (related to indentation level) suggest a deeper bug. That problem, at least, could be solved by using Spotless and [its support for google-java-format](https://github.com/diffplug/spotless/blob/main/lib/src/main/java/com/diffplug/spotless/java/GoogleJavaFormatStep.java).

We could probably get something set up adequately, but I'm still not sure it would be a net win: For Git in particular, it's easy for commit hooks to interfere with work when people are testing changes locally, and of course the same goes for test-time checks. That goes for when they succeed (since that takes time) as well as when they fail. And then I'd expect to have to maintain it over time.

Again, these are all fairly small concerns, but given that we get most of what we need from the internal checks, additional external checks would likely have more costs than benefits.

[...]

PR #5363, opened 2020-12-21
Added checkstyle to force the code style to be Google's Java Style compliant

AlbertoSinigaglia:
Hi,
I've added to the main `pom.xml` the `checkstyle` plugin, in order to enforce the code to be Google's Java Style as specified on the file [CONTRIBUTING.md](https://github.com/google/guava/blob/master/CONTRIBUTING.md)

This however cause "a awful lot" of warnings in the `guava` sub-project, which if you wan i can take care, but it would be a very big PR, and so i don't know if it could be actually a great choice 

\[cpovirk edit: related to #5362\]

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -102,6 +102,34 @@
     <plugins>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-checkstyle-plugin</artifactId>
+        <version>3.1.1</version>
+        <dependencies>
+          <dependency>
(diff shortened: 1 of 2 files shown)


cpovirk:
Closing, as described in #5362. Thanks for running the test: It suggests that our Travis config is broken, so I'm putting together a fix like https://github.com/google/truth/commit/fdbab6a4b0c5889cb1ccdb42026cd0af14e9b885.

PR #5365, opened 2020-12-21
Temporary PR to retry #5363 to see if Travis config was fixed by #5364.

cpovirk:
Co-authored-by: Alberto Sinigaglia <sinigagliaalberto@gmail.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -102,6 +102,34 @@
     <plugins>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-checkstyle-plugin</artifactId>
+        <version>3.1.1</version>
+        <dependencies>
+          <dependency>
+            <groupId>com.puppycrawl.tools</groupId>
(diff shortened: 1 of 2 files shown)


cpovirk:
I didn't have permission to push changes to the original PR, so I couldn't pull in my Travis change and have it tested there.

Aside: I couldn't find a way to force Travis to rerun for a PR -- or even to view the Travis _settings_ for Guava. I'm pretty sure I've been able to force reruns before (for _some_ project -- not sure if Guava specifically), so maybe something changed. I do have write permission on the GitHub repo, which sounds like [it ought to be enough](https://stackoverflow.com/a/17624403/28465), but maybe it has something to do with how that's through the layer of indirection of our "team?" (I don't believe we've changed any protections yet for Copybara, but maybe I'm wrong about that?)

cpovirk:
OK, the fix worked.

Now the build gets far enough to fail with Checkstyle violations: https://travis-ci.org/github/google/guava/jobs/750924487

PR #5382, opened 2021-01-16
Removed some redundant 'else's using early return

JuHyun419:
removed some redundant 'else'


cgdecker:
Thanks for the PR, but I don't think removing these `else`s improves the clarity of the code.

PR #5383, opened 2021-01-16
Normative modification

514190950:
1.Add annotation @override to interesting methods 
2.Extract method parameters as static  to improve performance


cgdecker:
- Those `@Override`s can't be added, as explained: `@SuppressWarnings("MissingOverride") // Supermethod does not exist under GWT.`
- There's no reason to extract that `Pattern`, as the `newMatchResult()` method itself is only called once.

Issue #5387, opened 2021-01-20
Add bit operations to UnsignedInteger class


BlackRose01:
Could you please consider to implement bit operations to the UnsignedInteger class?

Currently I try to migrate a CRC32 algorithm from C# to Java. This algorithm works only with unsigned integer values and use bit operations like xor, and, left shift.

1c3t3a:
We are a group of students ([Tobias](https://github.com/kariustobias), [Elena](https://github.com/Kara3), [Dominik](https://github.com/DJCoding) and me), who would like to contribute to guava, could we pick this up?

suchit07-git:
I would like to contribute to this issue.

PR #5396, opened 2021-02-01
Issue5387 - Added logical operators to UnsignedInteger class

kariustobias:
Added `and, or` and `xor` operations for the `UnsignedInteger` class according to the issue #5387.

We thought about also implementing left shift and right shift.
Since the `UnsignedLong` has a similar behaviour, it would make sense to add them there as well.


kariustobias:
@googlebot I signed it!

DJCoding:
@googlebot i signed it!

1c3t3a:
@googlebot I consent

Kara3:
@googlebot I consent

Issue #5377, opened 2021-01-02
CharSource.asByteSource(...).openStream() does not implement InputStream.available()


Marcono1234:
The `InputStream` returned by `CharSource.asByteSource(...).openStream()` (i.e. internal class [`ReaderInputStream`](https://github.com/google/guava/blob/master/guava/src/com/google/common/io/ReaderInputStream.java)) does not implement [`InputStream.available()`](https://docs.oracle.com/javase/8/docs/api/java/io/InputStream.html#available--).

Example:
```java
InputStream in = CharSource.wrap("a").asByteSource(StandardCharsets.UTF_8).openStream();
System.out.println(in.available()); // Returns 0
```

Possible implementation:
- If `com.google.common.io.ReaderInputStream.byteBuffer` is non empty, its `remaining()` value can be returned
- Otherwise try encoding remaining chars in `com.google.common.io.ReaderInputStream.charBuffer`
  Though note that even if buffer is not empty it might require more chars, e.g. when it only contains high surrogate and is missing low surrogate
- Otherwise check if `com.google.common.io.ReaderInputStream.reader` is `ready()` and if so call `reader.read()` (other `read(...)` methods might block?); afterwards repeat steps above
- Otherwise return 0

Though the above might be pretty inefficient in case someone repeatedly calls `available()` because it would then only encode one (or at most two) char(s) at a time. However, [`java.io.Reader.ready()`](https://docs.oracle.com/javase/8/docs/api/java/io/Reader.html#ready--) sadly only guarantees that `read()` won't block. Therefore there might be implementations whose other `read(...)` methods might block so encoding more chars would not be possible.

1c3t3a:
We are a group of students ([Tobias](https://github.com/kariustobias), [Elena](https://github.com/Kara3), [Dominik](https://github.com/DJCoding) and me), who would like to contribute to guava, could we pick this up?

Marcono1234:
Additionally I think this check will always be `false`:
https://github.com/google/guava/blob/c9f4b9fcb4ae756af0be113a6c307f6702cde404/guava/src/com/google/common/io/ReaderInputStream.java#L153-L154

If `doneFlushing == true` it was set in this invocation because otherwise the draining check at the beginning of the method would have returned fast.
However, if `doneFlushing` was set in this invocation, then draining would have started afterwards:
[...]

PR #5397, opened 2021-02-02
Issue-5377: Implemented `available()` for ReaderInputStream

1c3t3a:
This PR adds an implementation of `available()` for the `ReaderInputStream` class. The implementation follows the suggestions of issue #5377.

### DIFF
--- a/guava/src/com/google/common/io/ReaderInputStream.java
+++ b/guava/src/com/google/common/io/ReaderInputStream.java
@@ -185,6 +185,26 @@ final class ReaderInputStream extends InputStream {
     }
   }
 
+  @Override
+  public int available() throws IOException {
+    if (byteBuffer.hasRemaining()) {
+      // if we have remaining chars in the buffer, their count
+      return byteBuffer.remaining();
+    } else if (charBuffer.hasRemaining()) {
+      // if not, try to make quick progress by encoding available
+      // chars that have been read from the users reader
+      encoder.encode(charBuffer, byteBuffer, endOfInput);
+      return available();
+    } else if (reader.ready() && !endOfInput) {
+      // if none of the above was true, check if
+      // the reader is ready, and try to make progress on it
+      endOfInput = reader.read(charBuffer) < 0;
+      return available();
+    }
+
+    return 0;
+  }
+
   /** Returns a new CharBuffer identical to buf, except twice the capacity. */
   private static CharBuffer grow(CharBuffer buf) {
     char[] copy = Arrays.copyOf(buf.array(), buf.capacity() * 2);


guava/src/com/google/common/io/ReaderInputStream.java
```diff
@@ -185,6 +185,26 @@ public int read(byte[] b, int off, int len) throws IOException {
...
+  public int available() throws IOException {
+    if (byteBuffer.hasRemaining()) {
+      // if we have remaining chars in the buffer, return their count
+      return byteBuffer.remaining();
+    } else if (charBuffer.hasRemaining()) {
+      // if not, try to make quick progress by encoding available
+      // chars that have been read from the users reader
+      encoder.encode(charBuffer, byteBuffer, endOfInput);
+      return available();
+    } else if (reader.ready() && !endOfInput) {
```
Marcono1234, line 198: There has to be another branch checking `endOfInput && !doneFlushing` calling `encoder.flush(byteBuffer)` then (and checking for `OVERFLOW`, increasing the buffer, or throwing an exception is the result indicates an error).

Otherwise it might be possible that previously in the encode loop `encoder.flush(byteBuffer)` returned `OVERFLOW` and therefore `doneFlushing==false` and there are still pending bytes.

For example:
```java
// Requires Java 11+

int byteBufferSize = 8192; // Buffer size used by CharSource.AsByteSource.openStream()

// Charset whose CharsetEncoder.flush(...) puts bytes
Charset charset = Charset.forName("Cp930");

// \u0391: Causes CharsetEncoder.flush(...) to put additional bytes
String s = "a".repeat(byteBufferSize - 3) + "\u0391";
InputStream in = CharSource.wrap(s).asByteSource(charset).openStream();
in.read(new byte[byteBufferSize]);

// Should return 1
in.available();
```

guava/src/com/google/common/io/ReaderInputStream.java
```diff
@@ -185,6 +185,26 @@ public int read(byte[] b, int off, int len) throws IOException {
...
 
+  @Override
+  public int available() throws IOException {
+    if (byteBuffer.hasRemaining()) {
+      // if we have remaining chars in the buffer, return their count
+      return byteBuffer.remaining();
+    } else if (charBuffer.hasRemaining()) {
+      // if not, try to make quick progress by encoding available
+      // chars that have been read from the users reader
+      encoder.encode(charBuffer, byteBuffer, endOfInput);
```
Marcono1234, line 196: Would be good to check the result here, throwing an exception if the result indicates an error, or increasing the `charBuffer` or `byteBuffer` sizes as necessary:
- `charBuffer`: if `charBuffer` is full, but result is `UNDERFLOW` and `byteBuffer` is still empty
- `byteBuffer`: if result is `OVERFLOW` but `byteBuffer` is still empty; as far as I know there are some charset implementations which only but multiple bytes at bulk, and therefore would not put any bytes if the buffer has only size 1, which `ReaderInputStream` permits

`CharSource.asByteSource(...).openStream()` chooses buffer sizes where increasing the buffer should not be necessary, but the `ReaderInputStream` class is written in a style which would detect too small buffers and increases them as necessary, so it would be good to follow that.

Marcono1234 (commented):
Implementing this method is rather tricky, I have left some review comments which are hopefully helpful for you.
Maybe it would be good to wait for feedback from some of the maintainers to know if implementing this method is really worth it.

It might be good to add some tests (possibly with the example code snippets I provide) to [`CharSourceTester.java`](https://github.com/google/guava/blob/master/guava-tests/test/com/google/common/io/CharSourceTester.java) (I hope that is the right test class).
And to add this change also to the [Android version of `ReaderInputStream`](https://github.com/google/guava/blob/master/android/guava/src/com/google/common/io/ReaderInputStream.java).

1c3t3a:
Allright, thank you very much for the detailed feedback! We implement a first version of what you described. We added a flushing case were the encoder gets flushed and we also now check for the `CoderResult` we receive when calling `encode()`. But running your test always produces an `OVERFLOW` when encoding to the buffer, even if we enlarged it with calling `startDraining()`. For clarity we committed our latest code and plan  to work on this in the next days.

kariustobias:
@googlebot I consent.

guava/src/com/google/common/io/ReaderInputStream.java
```diff
@@ -185,6 +185,76 @@ public int read(byte[] b, int off, int len) throws IOException {
     }
   }
 
+  @Override
+  public int available() throws IOException {
+    if (byteBuffer.hasRemaining()) {
+      // if we have remaining chars in the buffer, return their count
+      return byteBuffer.remaining();
+    } else if (charBuffer.hasRemaining()) {
```
Marcono1234, line 193: I think this also has to check `draining == true`, otherwise `CharSource.wrap("").asByteSource(charset).openStream().available()` would return the size of the empty byte buffer instead of 0.

Additionally you need to account for the situation where `draining == true`, but `byteBuffer.hasRemaining() == false` (which I think is the situation where you are experiencing the OVERFLOW).

The following should work (I hope):
```suggestion
    if (draining) {
      if (byteBuffer.hasRemaining()) {
        // if we have remaining bytes in the buffer, return their count
        return byteBuffer.remaining();
      } else {
        // if buffer has been drained completely, clear it then fall through to encoding
        draining = false;
        Java8Compatibility.clear(byteBuffer);
      }
    }
    
    if (charBuffer.hasRemaining()) {
```

----

Alternatively (and maybe better) you could also change `ReaderInputStream.read(...)` to always update `draining` before returning:
```diff
      // We stay in draining mode until there are no bytes left in the output buffer. Then we go
      // back to encoding/flushing.
      if (draining) {
        totalBytesRead += drain(b, off + totalBytesRead, len - totalBytesRead);
+       if (!byteBuffer.hasRemaining()) {
+         draining = false;
+         Java8Compatibility.clear(byteBuffer);
+       }
        if (totalBytesRead == len || doneFlushing) {
          return (totalBytesRead > 0) ? totalBytesRead : -1;
        }
-       draining = false;
-       Java8Compatibility.clear(byteBuffer);
      }
```

Then you only need to change it like this:
```suggestion
    if (draining) {
      // if we have remaining bytes in the buffer, return their count
      return byteBuffer.remaining();
    } else if (charBuffer.hasRemaining()) {
```

Marcono1234 (commented):
Thanks! I think this looks quite good now, though I am not a maintainer of this project and they might have other requirements.

Maybe it would be good to have a test in `CharSourceTester` like the following:
```java
public void testAsByteSource_openStream() throws IOException {
  Charset charset = StandardCharsets.UTF_8;
  byte[] expectedBytes = expected.getBytes(charset);
  ByteArrayOutputStream actualBytes = new ByteArrayOutputStream();

  InputStream inputStream = source.asByteSource(charset).openStream();

  while (actualBytes.size() < expectedBytes.length) {
    assertTrue(inputStream.available() > 0);
    int b = inputStream.read();
    if (b == -1) {
      break;
    }
    actualBytes.write(b);
  }

  assertEquals(0, in.available());
  assertArrayEquals(expectedBytes, actualBytes.toByteArray());
}
```

For the current test you added, it might be easier to have the more sophisticated tests which rely on certain buffer sizes in a separate test class (`ReaderInputStreamTest`) so that they can specify the buffer size (and `CodingErrorAction` if necessary) directly for `ReaderInputStream` instead of relying on implementation details of other classes.

guava/src/com/google/common/io/ReaderInputStream.java
```diff
@@ -185,6 +185,76 @@ public int read(byte[] b, int off, int len) throws IOException {
...
+      CoderResult result = encoder.encode(charBuffer, byteBuffer, endOfInput);
+
+      if (result.isOverflow()) {
+        // the byte buffer is too small, enlarge it
+        startDraining(true);
+      } else if (result.isUnderflow()) {
+        // the char buffer is too small, enlarge it
+        charBuffer = grow(charBuffer);
+      } else if (result.isError()) {
+        // Only reach here if a CharsetEncoder with non-REPLACE settings is used.
```
Marcono1234, line 205:
```suggestion
        // Only reach here if a CharsetEncoder with REPORT action is used.
```
[`IGNORE`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/charset/CodingErrorAction.html#IGNORE) would not cause this either.

guava/src/com/google/common/io/ReaderInputStream.java
```diff
@@ -185,6 +185,76 @@ public int read(byte[] b, int off, int len) throws IOException {
...
+
+      // the flip is necessary for appending at the right position
+      Java8Compatibility.flip(charBuffer);
+      // read until the read operation would block
+      while (reader.ready()) {
+        output = reader.read();
+
+        endOfInput = output < 0;
+        if (!endOfInput) {
+          charBuffer.append((char) output);
```
Marcono1234, line 223: `append(char)` works fine, but just delegates to `put(char)`. So maybe it would be better to directly use `put` here:
```suggestion
          charBuffer.put((char) output);
```

guava/src/com/google/common/io/ReaderInputStream.java
```diff
@@ -185,6 +185,76 @@ public int read(byte[] b, int off, int len) throws IOException {
...
+          Java8Compatibility.flip(byteBuffer);
+        } else if (result.isUnderflow()) {
+          // if an underflow occurs, the encoder is done with flushing
+          // and we proceed by draining the buffer
+          doneFlushing = true;
+          startDraining(false);
+          break;
+        } else if (result.isError()) {
+          // Only reach here if a CharsetEncoder with non-REPLACE settings is used.
+          result.throwException();
```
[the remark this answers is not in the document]
Marcono1234: Though on second thought leaving this check in would not hurt either and would protect against `CharsetEncoder` implementations whose `implFlush​` method violates the contract.
However, you would need to adjust the comment then.

guava/src/com/google/common/io/ReaderInputStream.java
```diff
@@ -185,6 +185,76 @@ public int read(byte[] b, int off, int len) throws IOException {
...
+
+    } else if (!endOfInput && reader.ready()) {
+      // if none of the above was true, check if
+      // the reader is ready, and try to make progress on it
+      int output;
+
+      // the flip is necessary for appending at the right position
+      Java8Compatibility.flip(charBuffer);
+      // read until the read operation would block
+      while (reader.ready()) {
```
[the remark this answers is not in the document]
Marcono1234: Though as additional optimization you could check whether the class of the reader is known to never block, i.e. [`StringReader`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/io/StringReader.html) (but to be safe check `getClass() ==` instead of `instanceof` because `StringReader` makes no guarantees whether it blocks and maybe someone wrote a blocking subclass) and Guava's [`CharSequenceReader`](https://github.com/google/guava/blob/master/guava/src/com/google/common/io/CharSequenceReader.java).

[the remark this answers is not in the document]
1c3t3a: > ath.min(charBuffer.remaining(), X)

Yeah that's a good question. Implementing the `Math.min(charBuffer.remaining(), X)` approach could lead to `available()` being blocked... on the other hand reading char by char is horribly inefficient.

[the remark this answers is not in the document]
Marcono1234: Ah, sorry my comment was not that clear; I meant that the `Math.min(charBuffer.remaining(), X)` limit should be used as additional condition for the `read()` loop, i.e. reading a single char at a time. This is still pretty inefficient, but at least not as inefficient as filling the complete buffer, especially if the caller might not even be interested in that many bytes.

cgdecker:
I haven't looked at the code yet, but I'm not really sure this is worth doing given that it sounds like it's not especially simple to implement. I have some questions about the reason for wanting this in the first place as well, which I asked on the other issue.

PR #5402, opened 2021-02-09
change in peek() method of PeekingImpl class.

NihalAgarwal:
Added validation, if the iterator.hasNext() return true, then only we change the value of hasPeeked to true, because before what we are doing is without checking iterator contains some value or not we mark hasPeeked to true in peek() method and, if the user call hasNext() method of PeekingImpl class immediately after calling peek() method, it will always return true.


NihalAgarwal:
@googlebot I fixed it.

NihalAgarwal:
@googlebot I fixed it.

PR #5403, opened 2021-02-09
change in peek() method of PeekingImpl class

NihalAgarwal:
Added validation, if the iterator.hasNext() return true, then only we change the value of hasPeeked to true, because before what we are doing is without checking iterator contains some value or not, we mark hasPeeked to true in peek() method and, if the user call hasNext() method of PeekingImpl class immediately after calling peek() method, it will always return true.

### DIFF
--- a/guava/src/com/google/common/collect/Iterators.java
+++ b/guava/src/com/google/common/collect/Iterators.java
@@ -1147,8 +1147,8 @@ public final class Iterators {
     @Override
     public E peek() {
       if (!hasPeeked) {
+        if(iterator.hasNext()) hasPeeked = true;
         peekedElement = iterator.next();
-        hasPeeked = true;
       }
       return peekedElement;
     }


netdpb:
Is there a particular bug you're fixing here?

guava/src/com/google/common/collect/Iterators.java
```diff
@@ -1147,8 +1147,8 @@ public void remove() {
     @Override
     public E peek() {
       if (!hasPeeked) {
+        if(iterator.hasNext()) hasPeeked = true;
```
netdpb, line 1150: I'm not convinced this change makes a difference.

`hasPeeked` starts out false inside this block.
If `iterator.hasNext()` returns false, then `iterator.next()` on the next line will throw, and even in the old code `hasPeeked` will stay false.
If `iterator.hasNext()` returns true, then `hasPeeked` will become true in the old code as well.

So I think in all cases the returned value and the state is the same, but your change introduces a new branch to the code where there wasn't one before. So I'm not sure your change simplifies the code or fixes a logic bug.

What do you think?

NihalAgarwal: In one of the case when iterator return null (no more elements left) and calling `peek()` method will return null and set `hasPeeked `value to true. In every case, either iterator has some value or not, `hasPeeked ` always set to true in old code. Immediately calling ` hasNext() `after` peek() `will always result to true, and `hasNext()` method is made to check whether iterator or peek contains some element or not.

netdpb: `iterator.next()` doesn't return null when there are no more elements left; [it throws `NoSuchElementException`](https://docs.oracle.com/en/java/javase/14/docs/api/java.base/java/util/Iterator.html#next()). That means that in the old code, `hasPeeked = true` won't be reached when there are no more elements. Right?

netdpb:
I'm going to close this for now, as I don't think it's adding value. Please reopen if you have more to discuss!

PR #5407, opened 2021-02-12
chore: remove deprecated method Class#newInstance()

surajdhamecha:


surajdhamecha:
@cpovirk ,PTAL

surajdhamecha:
@netdpb, PTAL

PR #5432, opened 2021-03-16
improve documentation

Eclixal:
Improved documentation of the Platform class

### DIFF
--- a/guava/src/com/google/common/base/Platform.java
+++ b/guava/src/com/google/common/base/Platform.java
@@ -59,20 +59,20 @@ final class Platform {
   }
 
   /**
-   * Returns the string if it is not null; an empty string if it is null
+   * Returns the string if it is not null; an empty string if it is null.
    *
    * @param string the string to test and possibly return
-   * @return {@code string} is is not null; {@code ""} if it is null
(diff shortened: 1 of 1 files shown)


guava/src/com/google/common/base/Platform.java
```diff
@@ -58,10 +58,22 @@ static boolean stringIsNullOrEmpty(@Nullable String string) {
     return string == null || string.isEmpty();
   }
 
+  /**
+   * Returns the string if it is not null; an empty string if it is null
```
eamonnmcmanus, line 62: Nitpick: can you add a period at the end of this sentence, and the sentence on the next method?

guava/src/com/google/common/base/Platform.java
```diff
@@ -58,10 +58,22 @@ static boolean stringIsNullOrEmpty(@Nullable String string) {
...
+   *
+   * @param string the string to test and possibly return
+   * @return {@code string} is is not null; {@code ""} if it is null
```
eamonnmcmanus, line 65: Typo: s/is is/if it is/

eamonnmcmanus (changes requested):
Thanks!

Issue #2549, opened 2016-08-16
Ints,Longs.tryParse can be made more flexible.


zolyfarkas:
can we make :

Ints.tryParse(String s, int radix)
Longs.tryParse(String s, int radix)

more flexible:

Ints.tryParse(CharSequence s, int radix)
Longs.tryParse(CharSequence s, int radix)

?

kluever:
Internally, Chris said:

"It's a binary-incompatible (but source-compatible) change. That said, it looks like Ints.tryParse and Longs.tryParse are both @Beta, so we're allowed to make binary-incompatible changes to them. We might still break people who ignored the @Beta policy, so we'll have to weigh that against the gains. Alternatively, we could add CharSequence overloads. That would make the API look a little bigger and scarier than it is, so I guess every approach has some small pros and cons."

zolyfarkas:
Overloads would work, the only concern is that the extra indirection (assume no copy paste) increases the likelihood of hitting the MaxInlineLevel limit which might result in perf degradation in certain use cases.

noamgrinch:
Hello,

Regarding #5465,
I do not have an example of actual code of mine which uses .tryParse on different implementations of CharSequence other than String. However, there might be some cases an overloading method would be better. For example, cases when manipulations and logic should be applied on a CharSequence is better done on StringBuilder or StringBuffer(Thread-safe).

PR #5465, opened 2021-04-05
Added Ints, Longs.tryParse(CharSequence) overloading methods regarding #2549.

noamgrinch:
The algorithms are basically the same as the tryParse(String) methods with very small changes.


nick-someone:
Hi @noamgrinch, thanks for your contribution.

Unfortunately, we've not gotten to deciding whether adding this overload is a good idea or not, so we can't accept this PR.

It would be really helpful for us to understand your use case though - do you have visible code where you have a non-`String` `CharSequence` that you'd like to `tryParse` from? Especially if turning that `CharSequence` into `String` via `toString()` or similar would be prohibitively expensive? Please comment on #2549 with any examples.

Thanks!

PR #5526, opened 2021-04-23
checkNotNull's argument should not be marked as NonNull

TomaszGaweda:
checkNotNull ensures that output it's not null, it's this method responsibility to check non-nullability. NotNull on input parameter should be used if method assumes input is not null/requires it.

Adding NotNull broke IntelliJ's code analysis. IntelliJ assumed that NonNull argument must be null on method input, however in fact method's contract is to ensure that on output the value is not null.

### DIFF
--- a/guava/src/com/google/common/base/Preconditions.java
+++ b/guava/src/com/google/common/base/Preconditions.java
@@ -873,7 +873,7 @@ public final class Preconditions {
    * @see Verify#verifyNotNull Verify.verifyNotNull()
    */
   @CanIgnoreReturnValue
-  public static <T extends @NonNull Object> T checkNotNull(T reference) {
+  public static <T> T checkNotNull(T reference) {
     if (reference == null) {
       throw new NullPointerException();
     }
@@ -891,7 +891,7 @@ public final class Preconditions {
    * @see Verify#verifyNotNull Verify.verifyNotNull()
    */
   @CanIgnoreReturnValue
-  public static <T extends @NonNull Object> T checkNotNull(
+  public static <T> T checkNotNull(
       T reference, @Nullable Object errorMessage) {
     if (reference == null) {
       throw new NullPointerException(String.valueOf(errorMessage));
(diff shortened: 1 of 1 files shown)


TomaszGaweda:
@googlebot I signed it!

TomaszGaweda:
@googlebot I signed it!

TomaszGaweda:
@googlebot I fixed it.

cpovirk:
Unfortunately, we are trying to accommodate many different tools' nullness analyses, and every change we make to accommodate one tends to make another one unhappy :(

It looks like this specific change was made to accommodate Error Prone's `ReturnMissingNullable`, which has since [been deleted](https://github.com/google/error-prone/commit/c8c034d99b48df0034aad64ea777b480f6c3cbf9). So we may be able to back it out.

And in fact, we're in the process of finally annotating Guava for nullness in a more modern way, as you can see in some of [these commits](https://github.com/google/guava/commits?after=ecbbcc5fc97eeb7384c18b4a2666e7b531e38304+0&author=cpovirk). So the change may even happen soon.

Still, there are two bits of bad news:

First, I still don't know if backing out the annotation will cause issues for _some_ tool. If it does, we might leave the annotation in place. And that's because...

Second, the purpose of `Preconditions.checkNotNull` is for code like the following:

```java
void myMethod(@NonNull Object o) {
  checkNotNull(o);
  ...
}
```

(We've written some more about [the meaning we intend for "preconditions."](https://github.com/google/guava/wiki/ConditionalFailuresExplained) But we intend to write more about `Preconditions.checkNotNull` specifically, since this is a common question.)

In code like that, it makes sense for `checkNotNull` to require a non-null input: That way, nullness analyses can catch you if you accidentally try to call `checkNotNull` on a parameter that _can_ be null.

It's true that there are still cases in which you need to perform null checks on values that can't be annotated as non-null. (This sometimes even happens with method parameters.) In that case, options include (depending on the specific tool you're using):
- `Objects.requireNonNull(o)`
- a manual `if (o == null) throw new NullPointerException();`
- using `checkNotNull` while suppressing the warning

Again, though, we should write this down in more detail somewhere.

...and, while we feel _OK_ about requiring a non-null value here, it's not clear if that's going to cause more trouble than it's worth. So we'll see how things go as we adopt nullness annotations more widely.

PR #5549, opened 2021-05-09
feature: add Optional ifPresent.

WallanceLee:
Hello. When I use Optional, I find that it needs to check isPresent before consuming value wrapped in Optional objects.
So I create a new method ifPresent with a parameter Consumer to wrap isPresent and do something.
This is my first commit in guava project and I have signed CLA. 
Thanks for your review for my PR.


WallanceLee:
I have signed CLA and install CLA in my repo. Please recheck.

PR #5563, opened 2021-05-18
FEATURE: Add BloomFilter support for conversion between BloomFilter.bits and Redis bitmap

Yanam:
This PR adds the method to convert guava bitmap and redis bitmap to each other.

```java
BloomFilter<Person> bloomFilter = BloomFilter.create(Funnels.stringFunnel(Charset.forName("utf-8")), 1000000, 0.01);

bloomFilter.put("foo");
bloomFilter.put("bar");

//Convert to  a byte array in Redis bitmap format
byte[] data = bf.serialToRedisBitmapBytes();

//Store in redis
redis.set(key.getBytes(), data)
```

Real life use-case:
In my project, I need to provide BloomFilter filtering service based on the distributed cache such as Redis. My implementation method is to base the basic logic on the existing code of BloomFilter of Guava, such as the hash method, but the bitmap is based on the bitmap of Redis. This works well so far, but there are situations where when building BloomFilter with a lot of data, such as building BloomFilter with a million data, the number of Redis operations becomes a performance bottleneck. My approach is to build BloomFilter in the local cache. Then convert Guava's bitmap to Redis's bitmap format and write it to Redis once.

### DIFF
--- a/guava-tests/test/com/google/common/hash/BloomFilterTest.java
+++ b/guava-tests/test/com/google/common/hash/BloomFilterTest.java
@@ -525,6 +525,19 @@ public class BloomFilterTest extends TestCase {
     assertEquals(bf, BloomFilter.readFrom(new ByteArrayInputStream(out.toByteArray()), funnel));
   }
 
+  public void testBitmapRedisSerialization(){
+    Funnel<byte[]> funnel = Funnels.byteArrayFunnel();
+    BloomFilter<byte[]> bf = BloomFilter.create(funnel, 100);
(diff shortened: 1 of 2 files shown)


cpovirk:
Thanks. My guess is that we won't merge this: We probably don't want to have code that we'd ideally have our CI test with Redis in order to ensure that it keeps working. But I'll keep it open to see if demand arises. If nothing else, maybe we'll see demand for exposing the underlying bits in a form that is more generally useful.

PR #5575, opened 2021-05-26
Avoid redundant calculation.

dugenkui03:


dugenkui03:
@googlebot I signed it!

amalloy:
This isn't such a trivial optimization, because `get` has side effects, and omitting a call to `get` could impact other axes of performance, perhaps negatively.  I think we would want to see evidence that this rewrite actually improves performance overall, rather than just for a single call to `getAll`, before we'd merge this.

PR #5588, opened 2021-06-02
Enabling Tables.newCustomTable with ConcurrentMap

bauna:
If I want to use [Tables.newCustomTable](https://guava.dev/releases/22.0/api/docs/com/google/common/collect/Tables.html#newCustomTable-java.util.Map-com.google.common.base.Supplier-) with ConcurrentMaps, the current implementation is not thread safe when the row is not present. Because it's delegating to the backing map by calling computeIfAbsent.

### DIFF
--- a/guava/src/com/google/common/collect/StandardTable.java
+++ b/guava/src/com/google/common/collect/StandardTable.java
@@ -130,12 +130,7 @@ class StandardTable<R, C, V> extends AbstractTable<R, C, V> implements Serializa
   }
 
   private Map<C, V> getOrCreate(R rowKey) {
-    Map<C, V> map = backingMap.get(rowKey);
-    if (map == null) {
-      map = factory.get();
-      backingMap.put(rowKey, map);
-    }
-    return map;
+    return backingMap.computeIfAbsent(rowKey, key -> factory.get());
   }
 
   @CanIgnoreReturnValue


amalloy:
Could you explain the problem in more detail? What is the use case where this method causes a problem? To me this looks like a correct implementation, given that the documentation for newCustomTable notes:

> Note that this implementation is not synchronized. If multiple threads access this table concurrently and one of the threads modifies the table, it must be synchronized externally.  
> The table assumes complete ownership over `backingMap` and the maps returned by factory. Those objects should not be manually updated [...]

moonfruit:
@amalloy Yes, you are right. `Tables.newCustomTable()` makes no guarantee about thread-safe.
But if we merge this little pull request, we can assure `Tables.newCustomTable(new ConcurrentHashMap<>(), ConcurrentHashMap::new)` thread-safe because of `ConcurrentHashMap.computeIfAbsent()` is thread safe.
Why not merge it?

bauna:
Hi @amalloy,
@moonfruit has explained it better than me. If we made that small change it enables us to Tables.newCustomTable(new ConcurrentHashMap<>(), ConcurrentHashMap::new) and be sure it is thread safe without change anything else. 
Also it's a more correct implementation because it follows the "Tell, don't ask" principle by delegating to the inner map.

cpovirk:
This change would not be enough to make `newCustomTable` thread-safe: While it makes calls to `getOrCreate` thread-safe (assuming that the backing map and supplier are thread-safe), it does not make other operations like `remove` thread-safe.

Still, the simplification to the code is nice. Unfortunately, we currently still maintain compatibility with older versions of Java for our Android and JRE7 users, so we can't use `computeIfAbsent` in code that works there. We _can_ use `computeIfAbsent` in environments in which it's available while avoiding it when it's not, but this typically makes our code more complex. So we almost always avoid that approach.

Once we start requiring all our users to target environments that contain `computeIfAbsent`, we could make changes like this. But I expect that not to happen for years.

PR #5591, opened 2021-06-04
Add methods similar to ConcurrentMap to Table

moonfruit:
I refer to the code [java.util.Map](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/Map.java) to make this change.


moonfruit:
@amalloy Is there any problem with this?

Issue #5612, opened 2021-06-17
Remove @Beta from com.google.common.util.concurrent.RateLimiter


zaragozamartin91:
RateLimiter has been around since version 13 and it is stilled marked as @Beta

jbduncan:
For people waiting for `RateLimiter` to be made stable, an alternative that looks good is [resilience4j](https://github.com/resilience4j/resilience4j), which also has a rate limiter, amongst other utilities for "fault tolerance".

I've never actually used it, so I can't personally vouch for it, but Hystrix, an older library in this space, recommends it in their [README](https://github.com/Netflix/Hystrix) at the time of me writing this.

cpovirk:
Thanks, @jbduncan! We've been wondering if we should recommend that, but we haven't found the time to look into it.

jbduncan:
As a side note, Netflix, the creators of Hystrix, say that they've using a technique called [adaptive concurrency limits](https://medium.com/@NetflixTechBlog/performance-under-load-3e6fa9a60581) instead of Hystrix since around 2018. They have a [library](https://github.com/Netflix/concurrency-limits) for this, but I know even less about it than resilience4j and Hystrix.

darshitpp:
While [resilience4j](https://github.com/resilience4j/resilience4j) is indeed a rate limiting alternative, I believe both the RateLimiters are different. Guava's RateLimiter is a blocking call, whereas Resilience4j RateLimiter throws an `RequestNotPermitted` exception. I found Guava's to be easier to use for my use case.

olehen:
Another alternative is [https://github.com/bucket4j/bucket4j](bucket4j) which have the option of using blocking calls if you want. But I agree that it would be nice to have this out of Beta.

beanthemoonman:
Is there any sort of thought on how the Rate Limiter API would change before getting marked as stable? The team I work with has an understandable aversion to including new libraries to our codebase, but Guava is currently a library we use and that's not changing. It'd be great to understand what the risk is to incorporating it into our codebase, because the functionality is very useful.

MikeEnRegalia:
After five+ years of beta status, wouldn't it be possible to just remove it and cancel the one remaining todo?

PR #5614, opened 2021-06-18
Update RateLimiter.java

ahmedElatrouz:
Removing the @beta from the RateLimiter


ahmedElatrouz:
> @googlebot I signed it!

cgdecker:
Thanks for the PR, but removing `@Beta` is more of an issue of deciding that we're ok with locking down the API as it is and committing to not making breaking changes. In general we're trying to be more open to that, but I'm not sure about `RateLimiter` because I believe the person on our team who's worked with it the most feels there are some shortcomings to the design. I'm going to close this PR (we can pretty easily remove the annotation when we decide it should happen) but I'll see if I can get a more informed opinion on #5612.

PR #5619, opened 2021-06-21
Added url to guava pom

### LINKED ISSUE #5618: wrong url in pom file
guava does not define a project URL in the pom file. In that case, maven inherits the value of the parent pom and appends the artifact id (see [maven documentation](https://maven.apache.org/ref/3.6.1/maven-model-builder/)).
By that maven resolves the URL to https://github.com/google/guava/guava, which does not exist.

## Suggested Solution

Add an URL tag to `guava/pom.xml`

jakobbraun:
Closes #5618


jakobbraun:
@googlebot I signed it!

PR #5638, opened 2021-07-07
Remove else keyword

JuHyun419:
If the if statement is satisfied, the method returns immediately.
so, remove else keyword.


cpovirk:
Some people prefer to include `else` so that the two `return` statements are parallel and so that it's _absolutely immediately_ clear that the two branches can't both execute. Others prefer to omit `else` because it's unnecessary extra code. I suspect that we're not consistent about this across all of Guava, but I think that that's OK, and I think this code is OK either way.

PR #5641, opened 2021-07-08
Remove Generic's type parameter

JuHyun419:
Remove Generic's type parameter(type inference) in return


kevinb9n:
JuHyun419, the pull requests you've been sending simply aren't helpful. Please don't send more like these.

kevinb9n:
I'm sorry for giving a too-brief reply that might have communicated hostility. I'll explain better.

Your last three PRs (up to this) have addressed only questions of pure preference -- some people prefer it one way, some the other.

And in fact, small cleanups to implementation code are not always worth merging, even when we would *agree* it's how we'd rather have done it the first time.

We explained _individually_ on the previous two issues, but I wanted to make sure to address the overall pattern.

HTH

PR #5644, opened 2021-07-12
Halve the memory used by BloomFilter.readFrom()

pwr2c:
The BloomFilters  readFrom method reads into a separately allocated array, then converts (clones) the array into the expected object.
The code has been modified to directly read into the final array, not requiring two instances of the array (for a short duration) and the time/cpu used for the copy.


crazyguero (approved):

cpovirk:
Closing in favor of #5770 (which has the CLA showing as accepted).

Issue #5648, opened 2021-07-16
Incorrect hash result from murmur3_32 with String input containing surrogate pairs


findepi:
Here is the test program comparing behavior of

- Guava's `Hashing.murmur3_32` with `hashBytes(str.getBytes(encoding))`
- Guava's `Hashing.murmur3_32` with `hashString(str, encoding)`
- Airlift's `Murmur3Hash32.hash(str.getBytes(encoding))` (https://github.com/airlift/slice/blob/d3a025291fd8d6a062e44f3823ee49196783ab9c/src/main/java/io/airlift/slice/Murmur3Hash32.java)

```
public class Murmur
{
    public static void main(String[] args)
    {
        for (String string : List.of("plan ascii", "BMP Piękna łąka w 東京都",  "surrogate pair 💰")) {
            for (Charset charset : List.of(StandardCharsets.UTF_16, StandardCharsets.UTF_8)) {
                int airliftValue = Murmur3Hash32.hash(Slices.wrappedBuffer(string.getBytes(charset)));
                int guavaValue1 = Hashing.murmur3_32().hashBytes(string.getBytes(charset)).asInt();
                int guavaValue2 = Hashing.murmur3_32().hashString(string, charset).asInt();
                System.out.println("airliftValue = " + airliftValue);
                System.out.println("guavaValue1  = " + guavaValue1);
                System.out.println("guavaValue2  = " + guavaValue2);
                System.out.println();
            }
        }
    }
}
```

Results when run with Guava `30.1-jre` and Airlift Slice `0.40`:

Note that results agree except for the last case. I expect them to agree in all cases, especially between the two Guava alternative API methods.

```
string: plan ascii, charset: UTF-16
airliftValue = -731716445
guavaValue1  = -731716445
guavaValue2  = -731716445

string: plan ascii, charset: UTF-8
airliftValue = -218266838
guavaValue1  = -218266838
guavaValue2  = -218266838

string: BMP Piękna łąka w 東京都, charset: UTF-16
airliftValue = -989030725
guavaValue1  = -989030725
guavaValue2  = -989030725

string: BMP Piękna łąka w 東京都, charset: UTF-8
airliftValue = 103331700
guavaValue1  = 103331700
guavaValue2  = 103331700

string: surrogate pair 💰, charset: UTF-16
airliftValue = 2147098392
guavaValue1  = 2147098392
guavaValue2  = 2147098392

string: surrogate pair 💰, charset: UTF-8
airliftValue = -1114908744
[...]

PR #5649, opened 2021-07-16
Fix murmur3_32 with UTF-8 encoding for input with non-BMP character

findepi:
Fixes https://github.com/google/guava/issues/5648

### DIFF
--- a/guava-tests/test/com/google/common/hash/Murmur3Hash32Test.java
+++ b/guava-tests/test/com/google/common/hash/Murmur3Hash32Test.java
@@ -20,6 +20,7 @@ import static com.google.common.hash.Hashing.murmur3_32;
 
 import com.google.common.base.Charsets;
 import com.google.common.hash.HashTestUtils.HashFn;
+import java.nio.charset.Charset;
 import java.util.Random;
 import junit.framework.TestCase;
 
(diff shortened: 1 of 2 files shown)


findepi:
Admittedly, this as is not the most optimal approach, i will try to figure out how to fix the code without bailing out.
However, even with current state i think it's an improvement ("correctness first").

findepi:
@lowasser i realized the proper fix is a one-liner.

@lowasser @cpovirk can you PTAL?

findepi:
@lowasser seems like you authored portions of the code being changed here.
would you like to drop your two cents on the PR?

eamonnmcmanus:
I think we have a concern that fixing this bug may produce problems for systems that were persisting the old wrong hashcode, for example using it as part of a key in persistent storage. We recognize that the current computed hashcode is wrong, and that it is inconsistent with the hashcode you get if you hash the same string in a different but supposedly equivalent way. We haven't yet figured the best way forward.

findepi:
@eamonnmcmanus thank you for your reply.

> I think we have a concern that fixing this bug may produce problems for systems that were persisting the old wrong hashcode, for example using it as part of a key in persistent storage.

i sympathize with concern. Actually this is this context where we were able to identify the bug.
Guava-based code was not producing same hash value as non-Guava-based code.

> We recognize that the current computed hashcode is wrong, and that it is inconsistent with the hashcode you get if you hash the same string in a different but supposedly equivalent way.

Thanks for making it clear.

Note that it's not only about Guava APIs that should return same results (but they do not).
It's also about interoperability between applications that use Guava for calculating murmur3 and those that do not.
For example, as in https://github.com/apache/iceberg/pull/2836#issuecomment-884771775, it would be hard to implement 'compatible' hashing for Python.

> We haven't yet figured out the best way forward.

I don't think not fixing a bug is a long term option for anyone, would you agree?
What are the options you're considering?

I think we can consider

- just fix the bug
- make the `hashString` throw for surrogate pairs, to force applications to realize there is a problem; and then, after few releases, fix the bug 
- keep bogus implementation next to `Hashing#murmur3_32()` (eg as `Hashing#murmur3_32_bogus`)

can you think of any other options?

guava-tests/test/com/google/common/hash/Murmur3Hash32Test.java
```diff
@@ -51,27 +52,37 @@ public void testKnownStringInputs() {
...
+    assertStringHash(0xa167dbf3, "hell", Charsets.UTF_8);
+    assertStringHash(0x248bfa47, "hello", Charsets.UTF_8);
+    assertStringHash(0x3d41b97c, "http://www.google.com/", Charsets.UTF_8);
+    assertStringHash(0x2e4ff723, "The quick brown fox jumps over the lazy dog", Charsets.UTF_8);
+    assertStringHash(0xb5a4be05, "ABCDefGHI\u0799", Charsets.UTF_8);
+    assertStringHash(0xfc5ba834, "毎月１日,毎週月曜日", Charsets.UTF_8);
+    assertStringHash(0x8a5c3699, "surrogate pair: \uD83D\uDCB0", Charsets.UTF_8);
+
+    assertStringHash(0, "", Charsets.UTF_16);
+    assertStringHash(0xae9d4799, "k", Charsets.UTF_16);
```
eamonnmcmanus, line 67: There's a subtle problem with this. (At least, it took me a while to figure it out.) A byte array from the `UTF_16` encoding starts with a two-byte Byte Order Mark (BOM) which is either `fe ff` or `ff fe` and indicates the endianness that the remaining bytes will use. It's not specified which endianness a given Java platform will use, so it's not correct to hash a string using the `UTF_16` encoding if you want the result to be portable. I discovered this because some non-public tests were using the opposite endianness and failing. I rewrote these test cases to use `UTF_16LE` and updated the hashcodes. (You don't need to do that here but if we go ahead with this change then the modified test will be in the version that we use.)

findepi: @eamonnmcmanus indeed, this isn't portable. Thanks for catching this.
UTF-16LE is as good, as the only point of this test coverage is to exercise the code path that was not optimized for UTF-8.

If there is any value in changing anything in this PR, i am happy to apply changes.

cpovirk: I saw a reference to this conversation, and I dug up [Android issue 37074504](https://issuetracker.google.com/issues/37074504) ([fix](https://android-review.googlesource.com/c/platform/libcore/+/437155)), last seen [as "https://code.google.com/p/android/issues/detail?id=196848" in `ByteStreamsTest`](https://github.com/google/guava/blob/d32a049827c0dd588b51c7bbb98d54eb7384a602/guava-tests/test/com/google/common/io/ByteStreamsTest.java#L383). I think what Éamonn is seeing is a bug in the very old version of Android that we test with.

UTF-16LE is still the right pragmatic fix -- and probably the one I should have employed in `ByteStreamsTest`!

findepi:
Per https://github.com/google/guava/pull/5649#discussion_r677571105 i changed UTF-16 to UTFLE.

I also rebased on current master to resolve conflicts and let the CI run.
https://github.com/google/guava/pull/5654 was merged which added a couple more test cases.

guava-tests/test/com/google/common/hash/Murmur3Hash32Test.java
findepi: Done, changed to UTF-16LE (and updated the expected hashes accordingly).

findepi:
@eamonnmcmanus @cpovirk is there anything i can do to help move this forward?

eamonnmcmanus:
> @eamonnmcmanus @cpovirk is there anything i can do to help move this forward?

Deciding what to do here is on our queue of things to do. The fix in this PR will form the basis of whatever we decide on, but we don't yet know what that will be. In answer to the specific question, I think you've given us everything we need. Thanks!

eamonnmcmanus:
I updated #5648 with our current plans.

findepi:
@eamonnmcmanus thanks for the update.

i still think it would be good to merge preparatory commits from this PR ("Test ..."). What do you think?

do you want me to change this PR to introduce`murmur3_32_fixed` as per https://github.com/google/guava/issues/5648#issuecomment-911944304?

eamonnmcmanus:
Thanks @findepi for finding this bug and preparing such a thorough fix! That fix will be the basis for the change we make, but it is more straightforward for us if we make the change in Google's internal repo and then push it out. We'll be sure to credit you appropriately. I expect the change to land within the next week.

PR #5654, opened 2021-07-20
Remove redundant bit masking

findepi:


guava/src/com/google/common/hash/Murmur3_32HashFunction.java
```diff
@@ -391,20 +391,20 @@ public HashCode hash() {
   }
 
   private static long codePointToFourUtf8Bytes(int codePoint) {
-    return (((0xFL << 4) | (codePoint >>> 18)) & 0xFF)
```
eamonnmcmanus, line 394: This is a bit subtle. Because a code point can't have more than 21 bits, it's true that there can be at most 3 nonzero bits left after shifting right by 18, so the `0xFF` mask is unnecessary. I think a comment to that effect would be helpful.

guava/src/com/google/common/hash/Murmur3_32HashFunction.java
```diff
@@ -391,20 +391,20 @@ public HashCode hash() {
...
 
   private static long charToThreeUtf8Bytes(char c) {
-    return (((0xF << 5) | (c >>> 12)) & 0xFF)
+    return ((0x7 << 5) | (c >>> 12))
         | ((0x80 | (0x3F & (c >>> 6))) << 8)
         | ((0x80 | (0x3F & c)) << 16);
   }
 
   private static long charToTwoUtf8Bytes(char c) {
-    return (((0xF << 6) | (c >>> 6)) & 0xFF) | ((0x80 | (0x3F & c)) << 8);
```
eamonnmcmanus, line 407: Likewise here it would be helpful to have a reminder that `c < 0x800`, so `c` has at most 11 nonzero bits, or 5 after shifting right by 6.

eamonnmcmanus (changes requested):
I agree with the idea behind this PR. The existing code has a rather weird way of implementing the UTF-8 principle where the number of high 1 bits in the first byte indicates how many bytes there are. Your rewrite makes this clearer, since the masks are `0x3`, `0x7`, `0xF` for 2, 3, or 4 bytes respectively.

guava/src/com/google/common/hash/Murmur3_32HashFunction.java
findepi: This subtlety applies to all the cases, not just first byte.

This cannot be `verify`-d because that would probably be too expensive.

I can add an `assert` if you accept that in the code base (like here https://github.com/google/guava/blob/769e1e9c5ea1c6bddf14cec66b71d6824a8d28a6/guava/src/com/google/common/hash/LittleEndianByteArray.java#L81)
or i can add a comment

findepi:
@eamonnmcmanus added `assert` (can replace with a comment if you prefer so).
Also added test cases which exercise edge case values.

guava/src/com/google/common/hash/Murmur3_32HashFunction.java
eamonnmcmanus: Although there are a few `assert` statements in this project, I think we would prefer it to be just a comment.

I'm not sure I understand
> This subtlety applies to all the cases, not just first byte.

It seems to me that it's only the first byte that is subtle. In the other cases, we are explicitly extracting out 6 bits to fit into a byte after the high `10`. That seems comparatively straightforward. On the other hand, a similar subtlety does apply to the first byte of each of the other methods, if that's what you meant.

guava/src/com/google/common/hash/Murmur3_32HashFunction.java
```diff
@@ -391,20 +391,23 @@ public HashCode hash() {
...
+    assert  ((1 << 16) <= codePoint) && (codePoint < (1 << 21)) : "Unexpected codePoint " + codePoint;
+    return ((0xFL << 4) | (codePoint >>> 18))
         | ((0x80L | (0x3F & (codePoint >>> 12))) << 8)
         | ((0x80L | (0x3F & (codePoint >>> 6))) << 16)
         | ((0x80L | (0x3F & codePoint)) << 24);
   }
 
   private static long charToThreeUtf8Bytes(char c) {
-    return (((0xF << 5) | (c >>> 12)) & 0xFF)
+    assert (c < Character.MIN_SURROGATE) || (c > Character.MAX_SURROGATE) : "Unexpected char " + c;
```
eamonnmcmanus, line 402: No particular need for a comment here. In the other cases it is useful for readers to understand that the value couldn't overflow into the mandatory 0 bit after the string of high 1 bits, but here a `char` has 16 bits so after shifting right by 12 that clearly does fit after the high `1110`.

findepi: From bit shifting perspective -- i agree.
as i were adding input verification, i considered it could be useful here too (for reasons other that bit shifting)

since you recommended comments (vs assert), i removed this

guava/src/com/google/common/hash/Murmur3_32HashFunction.java
findepi: > It seems to me that it's only the first byte that is subtle.

i agree. Nvrm.

findepi:
@eamonnmcmanus changed to comments
also, added a clarifying TODO comment in the test case, where i had to use an incorrect expected value.

eamonnmcmanus:
Thanks, this looks fine. Because of the way we sync between our internal repo and GitHub, we'll be closing this PR and sending out an equivalent one, duly attributed to you.

findepi:
Thanks for review

findepi:
@eamonnmcmanus thanks for merging

PR #5655, opened 2021-07-21
Add MoreFiles.buildTraverser() which has customizable error handling

PrachiPrakash:
This adds a MoreFiles.buildTraverser() which  can used in the following way to handle IOException in a lenient way
```java
 Traverser<Path> fileTraverser = MoreFiles.buildTraverser()
            .onFail((p, ex) -> System.err.println("the file "+p.toString()+" has some I/O Error"))
            .build();
Iterable<Path> files = fileTraverser.breadthFirst(directory);
```

### DIFF
--- a/guava-tests/test/com/google/common/io/MoreFilesFileTraverserTest.java
+++ b/guava-tests/test/com/google/common/io/MoreFilesFileTraverserTest.java
@@ -19,12 +19,15 @@ package com.google.common.io;
 import static com.google.common.truth.Truth.assertThat;
 
 import com.google.common.collect.Iterables;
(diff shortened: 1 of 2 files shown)


nick-someone:
Hi there!

Per https://github.com/google/guava/blob/master/CONTRIBUTING.md#api-changes, we generally don't accept changes to the public API without thorough review, accounting for the likelihood of wide use, available alternatives, etc.

If you're interested in getting this added, we'd love to understand your use case in more detail. However, we don't tend to provide APIs that swallow `IOException` in a way that masks underlying errors, so it might be a pretty hard sell.

PR #5671, opened 2021-08-06
fix: typo spelling grammar

slowy07:
fixing typo and replace to correct word with reference from [Merriam webster](https://www.merriam-webster.com/) and [wikitionary](https://www.wiktionary.org/)

### DIFF
--- a/android/guava/src/com/google/common/net/HttpHeaders.java
+++ b/android/guava/src/com/google/common/net/HttpHeaders.java
@@ -140,7 +140,7 @@ public final class HttpHeaders {
   public static final String PROXY_AUTHORIZATION = "Proxy-Authorization";
   /** The HTTP {@code Range} header field name. */
   public static final String RANGE = "Range";
-  /** The HTTP {@code Referer} header field name. */
+  /** The HTTP {@code Referrer} header field name. */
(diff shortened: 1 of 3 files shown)


android/guava/src/com/google/common/net/HttpHeaders.java
```diff
@@ -140,7 +140,7 @@ private HttpHeaders() {}
   public static final String PROXY_AUTHORIZATION = "Proxy-Authorization";
   /** The HTTP {@code Range} header field name. */
   public static final String RANGE = "Range";
-  /** The HTTP {@code Referer} header field name. */
+  /** The HTTP {@code Referrer} header field name. */
```
eamonnmcmanus, line 143: https://en.wikipedia.org/wiki/HTTP_referer

eamonnmcmanus (commented):
Thanks! We can't change `Referer`, but we can fix the other typo.

This PR will be superseded by an equivalent one generated by our internal source code repository and credited to you.

android/guava/src/com/google/common/net/HttpHeaders.java
slowy07: ok sir, i will fix it !

PR #5676, opened 2021-08-09
Use a single Supplier instance inside a configured MultiMapBuilder

### LINKED ISSUE #5673: MultimapBuilder creates unnecessary objects
Looking at high memory usage in one of our components, we noticed that for every MultiMap created by a MultimapBuilder, we have a Supplier too (in our case it is a HashSetSupplier, but it applies to other Suppliers too)

the reason for that is that we do:
```
  public static final MultimapBuilder.SetMultimapBuilder<Object, Object> RELATIONS_MAP_BUILDER = MultimapBuilder
      .hashKeys(2)
      .hashSetValues(1);
```

which in turn calls this Guava code
```
    public SetMultimapBuilder<K0, Object> hashSetValues(final int expectedValuesPerKey) {
      checkNonnegative(expectedValuesPerKey, "expectedValuesPerKey");
      return new SetMultimapBuilder<K0, Object>() {
        @Override
        public <K extends K0, V> SetMultimap<K, V> build() {
          return Multimaps.newSetMultimap(
              MultimapBuilderWithKeys.this.<K, V>createMap(),
              new HashSetSupplier<V>(expectedValuesPerKey));
        }
      };
    }
```

This creates a new `hashSetSupplier` every time build is invoked.
This looks incorrect to me, and the suppliers should be properties of the newly returned builder, shouldn't they?

CodingFabian:
fixes #5673

### DIFF
--- a/guava/src/com/google/common/collect/MultimapBuilder.java
+++ b/guava/src/com/google/common/collect/MultimapBuilder.java
@@ -285,7 +285,7 @@ public abstract class MultimapBuilder<K0 extends @Nullable Object, V0 extends @N
     abstract <K extends K0, V extends @Nullable Object> Map<K, Collection<V>> createMap();
 
     /** Uses an {@link ArrayList} to store value collections. */
-    public ListMultimapBuilder<K0, @Nullable Object> arrayListValues() {
+    public <V0 extends @Nullable Object> ListMultimapBuilder<K0, V0> arrayListValues() {
       return arrayListValues(DEFAULT_EXPECTED_VALUES_PER_KEY);
     }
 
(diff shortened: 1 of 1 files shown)


CodingFabian:
This change required a bit of generics magic, since previously the suppliers got a type assigned that was dependent on the V type of the assigned variable (from the return type of the build signature)

I followed the pattern already existing for EnumSet to derive V from V0, and that does compile correctly with the forced `V0 -> V` casts

generally speaking: the type here doesn't matter. the Suppliers will always create empty backing collections which can be of any type at compile time and runtime.

if there is a better way to do this, I appreciate the insight

CodingFabian:
If consistency is preferred, I think we can rework `LinkedListSupplier` supplier to also use an instance per builder, instead of the lazily initialized global one.

```
private enum LinkedListSupplier implements Supplier<List<?>> {
    INSTANCE;

    public static <V extends @Nullable Object> Supplier<List<V>> instance() {
      // Each call generates a fresh LinkedList, which can serve as a List<V> for any V.
      @SuppressWarnings({"rawtypes", "unchecked"})
      Supplier<List<V>> result = (Supplier) INSTANCE;
      return result;
    }
```

PR #5698, opened 2021-08-30
Simplify TopKSelectorTest.testExceedMaxIteration()

perceptron8:


perceptron8:
k = 7 is minimum for which `TopKSelector.trim` _can_ fall back to `Arrays.sort` as it is minimum k for which 2 * k >= ceil(log2(2 * k - 1)) * 3.

Test data was found by concatenating every possible permutation of [1..k] with k copies of [0]. Luckily, one of such concatenations triggers off-by-one error in `TopKSelector` (or at least was triggering before #5691).

nick-someone:
Thanks for the reduction case! This will be merged internally, then pushed back out, attributed to you.

nick-someone (approved):

PR #5712, opened 2021-09-10
Use constructors of HashSet/LinkedHashSet directly instead of factory methods from Sets class.

turbanoff:
As stated in `Sets` javadoc, usages of such methods is not needed for Java 7. As Guava compiles for Java 7+ now, code can now take advantage of diamond syntax.

### DIFF
--- a/guava/src/com/google/common/cache/LocalCache.java
+++ b/guava/src/com/google/common/cache/LocalCache.java
@@ -35,13 +35,11 @@ import com.google.common.cache.CacheBuilder.NullListener;
 import com.google.common.cache.CacheBuilder.OneWeigher;
 import com.google.common.cache.CacheLoader.InvalidCacheLoadException;
 import com.google.common.cache.CacheLoader.UnsupportedLoadingOperationException;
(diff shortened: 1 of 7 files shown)


kluever:
FYI, we still maintain a pre-Java8 backport, so like to minimize the diffs when possible.

turbanoff:
Which language level this pre-Java8 version uses? 5, 6, 7?

kluever:
From https://github.com/google/guava#guava-google-core-libraries-for-java

> If you need support for JDK 1.7 or Android, use the Android flavor. You can find the Android Guava source in the android directory.

turbanoff:
As I can see guava already uses diamond operator. It means java 7 language features can be used. I will update PR description, to make it clear, that java 7 is enough and there is nothing java8 specific in my changes.

eamonnmcmanus:
Thanks! This will probably be handled as part of a larger-scale change migrating uses of these methods and other similar ones. We don't think it is worth tackling individual cases piecemeal.

PR #5726, opened 2021-09-24
Remove ambiguous return statement

aymar99:
https://github.com/google/guava/blob/f9b25b423ac6c7201d77387efe079ff540954de2/guava/src/com/google/common/util/concurrent/AbstractCatchingFuture.java#L132-L142

The return statement in the catch seems to be less meaningful since return only happens after finally is executed. Modified the code for better clarity


aymar99:
@googlebot  I fixed it.

PR #5727, opened 2021-09-24
Remove ambiguous return statement

aymar99:
https://github.com/google/guava/blob/f9b25b423ac6c7201d77387efe079ff540954de2/guava/src/com/google/common/util/concurrent/AbstractCatchingFuture.java#L132-L142

The return statement in the catch seems to be less meaningful since return only happens after finally is executed. Modified the code for better clarity


cpovirk:
This is a behavior change in unusual situations: I'm not sure if we have a test that demonstrates it, but if `setResult` throws (maybe a `StackOverflowError`, this PR will cause it to be caught and most likely swallowed.

...and the test results just came in, confirming that we do have a test. That's nice, at least :)

cpovirk:
I forgot to also cite https://github.com/google/guava/commit/19d312977252b2f494f6997d5dd881f349af6292 just for fun, showing how once we made this exact mistake in a similar API :)

Issue #5763, opened 2021-10-27
Description mistake in SmoothRateLimiter's java doc, line 167


lijinliangyihao:
It says: `The state of the RateLimiter (storedPermits) is a vertical line in this figure.`, but the **horizontal** line is `storedPermits`.

amalloy:
I'm not sure I agree with your reading of the Javadoc. The *axis* is horizontal, but the drawing seems to suggest drawing a vertical line up from a point on that axis. @cpovirk, I'm not sure who the best person to check this with is, but it looks like you've made more edits to SmoothRateLimiter than anyone else in recent history; perhaps you can confirm my assessment or loop in someone who can?

chaoren:
I agree with Alan. `maxPermits` is a vertical line. `thresholdPermits` is a vertical line. `0` (`storedPermits`) is a vertical line. Those are various "state[s] of the RateLimiter," that the comment refers to. The horizontal axis is called `storedPermits`, but each value (state) along the axis is a vertical line.

lijinliangyihao:
@chaoren Thank you for your kindly explanation, now I understand.

PR #5765, opened 2021-10-28
fixed issue #5763

nbkju9b:
Fixed javadoc in SmoothRateLimiter.java for issue #5763.


chaoren:
We haven't decided on the validity of the issue yet. Please wait for a resolution in #5763 first. I don't think the proposed change is correct. We may end up clarifying the comment without referring to the orientation of the lines.

PR #5769, opened 2021-11-01
Override MapMakerInternalMap#compute

### LINKED ISSUE #5766: Efficient implementation for MapMakerInternalMap#compute
The map created by `MapMaker#build` currently use the inefficient `compute` implementation from `ConcurrentMap`. I assume that it could be optimized for atomic operations by reducing lookups and locking it. Contention will also cause the mapping function to be ran multiple time, preventing atomic update.

TheMode:
Ensure that `remappingFunction` is executed atomically (and only once). Prevent excessive lookups in the case of contention.
This is the behavior of `ConcurrentHashMap`

Fixes #5766


TheMode:
@googlebot I signed it!

Issue #5773, opened 2021-11-03
Make `Iterators.mergeSorted()` stable


msmerc:
I've noticed that `Iterators.mergeSorted()` is unstable.

So for example if I have arrays: `[A_1, B_1, C_1]` `[A_2, B_2, C_2]` (where the comparator looks only at the letter), I'd like the result to iterate `[A_1, A_2,  B_1, B_2,  C_1, C_2]`. The current implementation doesn't guarantee this.

Might it be possible to add a stable version?

I confess it's not obvious what to do if the arrays are something like `[A_1, A_2, B_1]`, `[A_3]` - to me it would make sense to empty the first iterator of `A_n` elements before moving on to the second, so `[A_1, A_2, A_3, B_1]`.

[I appreciate I'm being a little lazy with my terminology here - I can provide a fully fleshed example if it's not clear]

tuannh982:
Could you provide some examples please?

msmerc:
@tuannh982: given a really simple data class:

```java
    public static class Datum {
        private final String letter;
        private final int number;

        public Datum(String letter, int number) {
            this.letter = letter;
            this.number = number;
        }

        public String letter() { return letter; }
        public int number() { return number; }

        @Override
        public String toString() {
            return letter + " " + number;
        }
    }
```

And the following test:

```java
    @Test
    public void sortMe() {
        List<Datum> left = List.of(
                new Datum("B", 1),
                new Datum("C", 1)
        );

        List<Datum> right = List.of(
                new Datum("A", 2),
                new Datum("C", 2)
        );
        
        var comparator = Comparator.comparing(Datum::letter);
        var it = Iterators.mergeSorted(
                List.of(left.iterator(), right.iterator()
        ), comparator);
        it.forEachRemaining(next->System.out.println(next));
    }
```

I'd get:

```
A 2
B 1
C 2
C 1
```

Note how `C 2` comes before `C 1`. If the sort were stable I'd expect to see `C 1` before `C 2` because that's the order they appear in the input.

Full code here: https://gist.github.com/msmerc/279c32fe6620b0dcf2dc6f2c4dfa1469

tuannh982:
@msmerc: okay, please assign this issue to me.

HarunSMetin:
@chaoren @msmerc  Hello. I would like to work on this issue. But This is my first time contributing to an open-source project. Therefore, I need some guidance. Could you please give me more information about this issue?

msmerc:
[...]

PR #5779, opened 2021-11-10
Make Iterators.mergeSorted() return more stable iterator

tuannh982:
Enchantment #5773 : make Iterators.mergeSorted() return more stable iterator

### DIFF
--- a/guava/src/com/google/common/collect/Iterators.java
+++ b/guava/src/com/google/common/collect/Iterators.java
@@ -1314,14 +1314,25 @@ public final class Iterators {
    */
   private static class MergingIterator<T extends @Nullable Object> extends UnmodifiableIterator<T> {
     final Queue<PeekingIterator<T>> queue;
+    final Comparator<? super T> itemComparator;
+    @Nullable
+    PeekingIterator<T> lastUsedIter = null;
 
     public MergingIterator(
-        Iterable<? extends Iterator<? extends T>> iterators, Comparator<? super T> itemComparator) {
+            Iterable<? extends Iterator<? extends T>> iterators, Comparator<? super T> itemComparator) {
+      this.itemComparator = itemComparator;
       // A comparator that's used by the heap, allowing the heap
       // to be sorted based on the top of each iterator.
-      Comparator<PeekingIterator<T>> heapComparator =
-          (PeekingIterator<T> o1, PeekingIterator<T> o2) ->
-              itemComparator.compare(o1.peek(), o2.peek());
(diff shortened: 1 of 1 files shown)


perceptron8:
@tuannh982 
With the help of a wrapper like:

  ```java
private static class ItemWithIndex<T> {
    public final T item;
    public long index;
    
    public ItemWithIndex(T item, long index) {
      this.item = item;
      this.index = index;
    }
  }
  ```

you can achieve your goal with almost no changes to the original code:

```java
/**
   * An iterator that performs a lazy N-way merge, calculating the next value each time the iterator
   * is polled. This amortizes the sorting cost over the iteration and requires less memory than
   * sorting all elements at once.
   *
   * <p>Retrieving a single element takes approximately O(log(M)) time, where M is the number of
   * iterators. (Retrieving all elements takes approximately O(N*log(M)) time, where N is the total
   * number of elements.)
   */
  private static class MergingIterator<T extends @Nullable Object> extends UnmodifiableIterator<T> {
    final Queue<PeekingIterator<ItemWithIndex<T>>> queue;

    public MergingIterator(
        Iterable<? extends Iterator<? extends T>> iterators, Comparator<? super T> itemComparator) {
      // A comparator that's used by the heap, allowing the heap
      // to be sorted based on the top of each iterator.
      Comparator<PeekingIterator<ItemWithIndex<T>>> heapComparator = Comparator.comparing(PeekingIterator::peek, Comparator
        .comparing((ItemWithIndex<T> element) -> element.item, itemComparator)
        .thenComparingLong((ItemWithIndex<T> element) -> element.index)
      );

      queue = new PriorityQueue<>(2, heapComparator);

      long index = 0L;
      for (Iterator<? extends T> iterator : iterators) {
        long effectivelyFinalIndex = index;
        if (iterator.hasNext()) {
          queue.add(Iterators.peekingIterator(Iterators.transform(iterator, element -> new ItemWithIndex<>(element, effectivelyFinalIndex))));
        }
        index++;
      }
    }

    @Override
    public boolean hasNext() {
      return !queue.isEmpty();
    }

    @Override
    @ParametricNullness
    public T next() {
      PeekingIterator<ItemWithIndex<T>> nextIter = queue.remove();
      ItemWithIndex<T> next = nextIter.next();
      if (nextIter.hasNext()) {
        queue.add(nextIter);
      }
      return next.item;
    }
  }
  ```

Transformations are lazy, so there should be almost no overhead.

tuannh982:
@perceptron8 the wrapper class actually lots easier to implement, and just little overhead. So I will close this PR

PR #5793, opened 2021-11-26
Refine examples for mapWithIndex

TennyZhuang:
The examples for `mapWithIndex` are really confusing, due to the following reasons:

1. `i` is always used as an index, which is confused with `index`.
2. `a:b` always represents mapping from a to b, but the examples use it differently.
3. The element `0,1,2` is exactly the same as indices.

We cannot know which argument is index and which one is element from the examples.

### DIFF
--- a/guava/src/com/google/common/collect/Streams.java
+++ b/guava/src/com/google/common/collect/Streams.java
@@ -410,10 +410,10 @@ public final class Streams {
    * <pre>{@code
    * mapWithIndex(
    *     Stream.of("a", "b", "c"),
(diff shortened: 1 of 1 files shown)


jievince (approved):

cpovirk:
Thanks. I made a couple changes in preparation for sending this for internal review. If you have any thoughts on [what I did](https://github.com/google/guava/commit/c98868c646ff767848feb77e3bd5d62d938bdace), let me know. That changed version has been merged, but we can make more changes as needed.

PR #5800, opened 2021-12-01
Fix Javadoc search on JDK 11:

### LINKED ISSUE #5457: Broken link in Javadoc
1. Go to https://guava.dev/releases/30.1.1-jre/api/docs/
2. Search for "unsignedbytes"
3. Get redirected to https://guava.dev/releases/30.1.1-jre/api/docs/undefined/com/google/common/primitives/UnsignedBytes.html which returns HTTP 404.

dmitry-timofeev:
Fixed Javadoc search feature on JDK 11, which is currently used in our
scripts updating snapshot Javadocs on guava.dev and building
Guava releases, by adding `no-module-directories` flag.

This option is not present in javadoc tool from JDK 8 and 13+,
hence we use profiles to conditionally pass this flag on 9-12 only.

Note that on JDK 17 javadoc generation does not work,
as it seems to have changed the warning on LVTI usage in sources
to an error.

Fixes #5457

TESTED=Locally with `mvn javadoc:javadoc` for `guava` module and its android flavour on JDK 11

### DIFF
--- a/android/pom.xml
+++ b/android/pom.xml
@@ -18,6 +18,8 @@
     <checker-framework.version>3.12.0</checker-framework.version>
     <animal.sniffer.version>1.20</animal.sniffer.version>
     <maven-javadoc-plugin.version>3.1.0</maven-javadoc-plugin.version>
+    <!-- Empty for all JDKs but 9-12 -->
+    <maven-javadoc-plugin.additionalJOptions></maven-javadoc-plugin.additionalJOptions>
     <maven-source-plugin.version>3.2.1</maven-source-plugin.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
   </properties>
@@ -190,6 +192,7 @@
             </additionalOptions>
             <linksource>true</linksource>
             <source>8</source>
+            <additionalJOption>${maven-javadoc-plugin.additionalJOptions}</additionalJOption>
           </configuration>
           <executions>
             <execution>
@@ -382,6 +385,24 @@
             </plugin>
           </plugins>
       </build>
+    </profile>
+    <profile>
+      <!--
+          Passes JDK 9-12-specific `no-module-directories` flag to Javadoc tool,
+          which is required to make symbol search work correctly in the generated
+          pages.
+
+          This flag does not exist on 8 and 13+ (https://bugs.openjdk.java.net/browse/JDK-8215582).
+
+          Consider removing it once our release and test scripts are migrated to a recent JDK (17+).
+       -->
+      <id>javadocs-jdk9-12</id>
+      <activation>
+        <jdk>[9,13)</jdk>
+      </activation>
+      <properties>
+        <maven-javadoc-plugin.additionalJOptions>--no-module-directories</maven-javadoc-plugin.additionalJOptions>
+      </properties>
     </profile>
   </profiles>
 </project>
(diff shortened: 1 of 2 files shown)


cpovirk:
[Failure under JDK8](https://github.com/google/guava/runs/4385244999?check_suite_focus=true), which we run CI under. As expected:

> Error:  Exit code: 1 - javadoc: error - invalid flag: --no-module-directories

Maybe just add...

`-Dmaven.javadoc.skip=true`

...to...

https://github.com/google/guava/blob/e64ad2634216a3bb8e2beeb26d673f72d24e293a/.github/workflows/ci.yml#L43

...given that "actual" Javadoc generation (for snapshots) is handled by...

https://github.com/google/guava/blob/e64ad2634216a3bb8e2beeb26d673f72d24e293a/.github/workflows/ci.yml#L83

cpovirk:
Hmm, but then pre-commit CI won't check Javadoc generation, so we wouldn't hear about breakages until after the commit.

Maybe it makes more sense for the flag to be conditional on the Java version -- either in the CI config or in Maven, as I think you'd suggested.

@cgdecker in case he happens to have thought about this.

cpovirk:
...not sure offhand why Truth doesn't show this same problem....

cpovirk:
Sorry, I'm now going back through some of your earlier messages.

Arguably an even better way to approach this would be to always _build_ with some new version of javac but then _test_ with each version we're targeting. My guess, though, is that that's harder.

cpovirk:
...but I guess it's _nice_ for Guava users to be able to build Guava with different JDKs.

dmitry-timofeev:
Thanks for taking a look!

Oh, I was convinced we skip javadocs by default, but I see it makes sense to run them [after the integration tests](
https://github.com/google/guava/blob/master/pom.xml#L191) if we also publish them on each commit.


I agree that punishing 17+ -using users and requiring a 3y-old JDK is not good :)

What do you think would be the best thing to do:
1. Support correct Javadoc generation on all JDKs (8, 9–12, 13+): requires a JDK-version activated Maven profile, like in [this POM](https://github.com/exonum/exonum-java-binding/blob/master/exonum-java-binding/pom.xml#L671-L681)):
    - (+) Don’t depend on migrating all our release scripts to 17 — can fix this issue right now
    - (+) Will produce correct Javadocs on all JDKs
    - (-) / (?) A bit of extra complexity in the configs, but not too much
    - If we eventually accept that JDK 17+ is required for development, and javadoc generation is not needed on earlier JDKs, it is a temporary measure.
1. Skip Javadoc generation on 8 and 12+ (conditionally set maven.javadoc.skip through a JDK-version activated Maven profile):
    - (+) Don’t depend on migrating all our release scripts to 17 — can fix this issue right now
    - (-) Same complexity as ^, but less benefits
1. (If you anticipate that 17 issues will get addressed soon) Wait till migrate to 17:
    - (+) No changes needed , no extra complexity in the configs
    - (?) Search won’t work on Javadocs generated with 11, but, arguably, [it won’t matter](https://github.com/google/guava/issues/5801#issuecomment-984637851)

cpovirk:
Thanks. (3) has some chance of turning into a bunch of work, which I predict that I personally will not follow up on. So, between (1) and (2), I think you make a good case that (1) is nicer.

dmitry-timofeev:
Changed to (1) and updated the reasoning in the description.

However, on JDK 17 it doesn't work just yet: https://github.com/google/guava/issues/5801#issuecomment-985430060

PR #5804, opened 2021-12-02
Update COPYING

sumuduliyanage:
corrected a typo -- "use" was corrected as "using"
"License" shall mean the terms and conditions for using, reproduction,

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -7,7 +7,7 @@
 
    1. Definitions.
 
-      "License" shall mean the terms and conditions for use, reproduction,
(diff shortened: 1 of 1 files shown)


cpovirk:
I think either can be correct: "Use" [can be a noun](https://www.dictionary.com/browse/use). It could be interesting to see which way reads more clearly to more people. However, my understanding is that we should not change the text of the license for legal reasons. So we should keep "use" to match [the text from Apache](https://www.apache.org/licenses/LICENSE-2.0).

PR #5812, opened 2021-12-06
Fix confusing Javadoc of memoizeWithExpiration

MartinSpiessl:
The previous wording could be read such that the cache is cleared in
the background once the expiration time is reached. However this is
not the case, so we simply remove the following part of the Javadoc:
"and removes the cached value after the specified time has passed"
The rest of the Javadoc already accurately describes the behavior,
though the part with "After the expiration time" could again be
misinterpreted, so we rewrite this to clarify that things only
happen once the "get" method is called.

### DIFF
--- a/guava/src/com/google/common/base/Suppliers.java
+++ b/guava/src/com/google/common/base/Suppliers.java
@@ -203,10 +203,9 @@ public final class Suppliers {
   }
 
   /**
-   * Returns a supplier that caches the instance supplied by the delegate and removes the cached
-   * value after the specified time has passed. Subsequent calls to {@code get()} return the cached
-   * value if the expiration time has not passed. After the expiration time, a new value is
-   * retrieved, cached, and returned. See: <a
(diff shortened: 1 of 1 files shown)


MartinSpiessl:
Sure, submitted my Contributor License Agreement a minute ago.

MartinSpiessl:
It says I should request a rescan regarding CLA from a repository manager. I assume the person who does the releases is for sure one of the a repository managers, so that would be @cpovirk.

cpovirk:
This is indeed a common cause of confusion.

It would still be good for the first sentence to say _something_ about the expiration. Maybe use the original phrasing but with "removes" changed to "invalidates?" But even that may be misleading, given that [`Cache.invalidate`](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/cache/Cache.html#invalidate(java.lang.Object)) uses that word for _immediate_ removal (I think). Other ideas?

PR #5872, opened 2022-01-17
Allow constructing CacheLoader with JDK Function or Supplier

findepi:
This allows a method to accept a `java.util.function.Function` and pass
is it on to `CacheLoader`, without need to wrap it with a
`com.google.common.base.Function`.

The methods taking `com.google.common.base.Function` or
`com.google.common.base.Supplier` are retained for binary compatibility.

### DIFF
--- a/guava/src/com/google/common/cache/CacheLoader.java
+++ b/guava/src/com/google/common/cache/CacheLoader.java
@@ -18,8 +18,6 @@ import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.google.common.annotations.GwtCompatible;
 import com.google.common.annotations.GwtIncompatible;
-import com.google.common.base.Function;
-import com.google.common.base.Supplier;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.common.util.concurrent.ListenableFutureTask;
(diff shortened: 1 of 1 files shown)


jbduncan:
I've seen similar feature requests and PRs rejected because it's so easy to turn a Java `Function` into a Guava one, with something like `javaFunction::apply`. Just so you're aware. :)

cgdecker:
I don't personally know the details of what the JVM does with lambdas and method references in cases like this, but I have a very strong feeling is that this is not something that one should worry about at all. If nothing else, I'd want to see compelling benchmarks showing an improvement in performance for something like this.

cgdecker:
Oh, I should also point out that we're generally recommending people use [Caffeine](https://github.com/ben-manes/caffeine) over our cache library these days.

findepi:
> , I'd want to see compelling benchmarks showing an improvement in performance for something like this.

good point.

note that a micro benchmark won't show potential improvement, as `(com.google.common.baseFunction) func::apply` will compile to something inlined. I would assume there won't be a problem unless the call becomes megamorphic (not saying the opposite is certain).

As I don't plan on working on this in near future, let me close this. 

> Oh, I should also point out that we're generally recommending people use Caffeine over our cache library these days.

I am aware. Does it mean the Cache classes in Guava are in maintenance mode, and not supposed to be improved?

cgdecker:
> Does it mean the Cache classes in Guava are in maintenance mode, and not supposed to be improved?

I'd consider them essentially maintenance mode, but I didn't mean to imply that they shouldn't be improved. I think the threshold for what we'd consider doing with them would be a bit higher though.

Issue #5874, opened 2022-01-17
Inconsistency in behavior between unmodifiable and immutable maps


WojciechZankowski:
Hi, it might be not a real issue, but I am just rising it. 

```java
// JDK

final TreeMap<Integer, String> baseMap = new TreeMap<>() {{put(1, "one");}};

final NavigableMap<Object, Object> jdkUnmodifiableNavigableMap = Collections.unmodifiableNavigableMap(baseMap);
jdkUnmodifiableNavigableMap.compute(null, (k, v) -> v);

final SortedMap<Object, Object> jdkUnmodifiableSortedMap = Collections.unmodifiableSortedMap(baseMap);
jdkUnmodifiableSortedMap.compute(null, (k, v) -> v);

final Map<Object, Object> jdkUnmodifiableMap = Collections.unmodifiableMap(baseMap);
jdkUnmodifiableMap.compute(null, (k, v) -> v);

final Map<Integer, String> jdk11_ImmutableMap = Map.of(1, "one");
jdk11_ImmutableMap.compute(null, (k, v) -> v);

// Guava

final ImmutableMap<Integer, String> immutableMap = ImmutableMap.of(1, "one");
immutableMap.compute(null, (k, v) -> v);

final ImmutableSortedMap<Integer, String> guavaImmutableSortedMap = ImmutableSortedMap.of(1, "one");
guavaImmutableSortedMap.compute(null, (k, v) -> v);
```

and other immutable maps from guava will throw

```
java.lang.UnsupportedOperationException
```

At the same time there are 2 implementations in Guava library that have slightly different behavior:

```java
final TreeMap<Integer, String> baseMap = new TreeMap<>() {{put(1, "one");}};
final NavigableMap<Object, Object> guavaUnmodifiableNavigableMap = Maps.unmodifiableNavigableMap(baseMap);
guavaUnmodifiableNavigableMap.compute(null, (k, v) -> v);
```

will throw
```
java.lang.NullPointerException
```

And the other one will just silently go through. 

```java
final BiMap<Integer, String> baseBiMap = HashBiMap.create();
baseBiMap.put(1, "one");
final BiMap<Integer, String> unmodifiableBiMap = Maps.unmodifiableBiMap(baseBiMap);
unmodifiableBiMap.compute(null, (k, v) -> v);
```

I am not sure if this has been done on purpose by design or was just simply overseen. Happy to contribute if you would qualify it as a something to be changed.

cgdecker:
[...]

PR #5877, opened 2022-01-18
#5874: Unify unmodifiable map behavior

WojciechZankowski:
Pull request to the following issue: https://github.com/google/guava/issues/5874

I could see 2 solutions to the issue:

1) Explicitly override methods that are delegated to default implementation from ```Map``` interface by ```ForwardingMap``` with ```UOE``` 
2) Replace ```ForwardingNavigableMap``` and ```ForwardingMap``` abstractions with something new that would override methods with ```UOE``` in the abstraction.

Below solution goes with the first approach, but if you will have any better option then I would be happy to change it.


cgdecker:
I think 1 is fine for now. I'm importing this internally now.

Issue #5888, opened 2022-01-21
Which kind of ImmutableMultimap does ImmutableMultimap.builder().build() create?


h908714124:
Here's a quote from the class-level javadoc of `ImmutableMultimap`:

> **Warning:** avoid direct usage of ImmutableMultimap as a type (as with Multimap itself). Prefer subtypes such as ImmutableSetMultimap or ImmutableListMultimap, which have well-defined Multimap.equals(java.lang.Object) semantics, thus avoiding a common source of bugs and confusion. 

In this regard, it seems rather unfortunate that `ImmutableMultimap.builder().build()` returns `ImmutableMultimap`. Is this a SetMultimap or ListMultimap? At least the Javadoc should be clear about that, or better yet, `build()` should return one of the specific types.

See: https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/collect/ImmutableMultimap.html

mvanhorn:
Implemented in PR #8261 - documented on builder() and build() that they produce ImmutableListMultimap, with a recommendation to use the explicit subtype builders.

PR #5890, opened 2022-01-22
5888: Improve Javadoc in ImmutableMultimap

WojciechZankowski:
Pull request to the following issue: #5888 

```java ImmutableMultimap.Builder#build ```
build

```java ImmutableMultimap#copyOf ```
copyOf

```java ImmutableMultimap#of```
of

Happy to apply any suggestions to improve the javadoc.

### DIFF
--- a/guava/src/com/google/common/collect/ImmutableMultimap.java
+++ b/guava/src/com/google/common/collect/ImmutableMultimap.java
@@ -75,7 +75,7 @@ public abstract class ImmutableMultimap<K, V> extends BaseImmutableMultimap<K, V
     implements Serializable {
 
   /**
-   * Returns an empty multimap.
+   * Returns an empty multimap of type {@code ImmutableListMultimap}.
    *
    * <p><b>Performance note:</b> the instance returned is a singleton.
    */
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by eamonnmcmanus — CHANGES_REQUESTED** [review]
Sorry for the delay in getting back to you!

I'll note that `ImmutableMultimap` explicitly says this:
> Warning: avoid direct usage of [ImmutableMultimap](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/collect/ImmutableMultimap.html) as a type (as with [Multimap](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/collect/Multimap.html) itself). Prefer subtypes such as [ImmutableSetMultimap](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/collect/ImmutableSetMultimap.html) or [ImmutableListMultimap](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/collect/ImmutableListMultimap.html), which have well-defined [Multimap.equals(java.lang.Object)](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/collect/Multimap.html#equals(java.lang.Object)) semantics, thus avoiding a common source of bugs and confusion.

This might also extend to the static methods of `ImmutableMultimap`, and certainly does if people care about the actual type they return. It might be better for the javadoc on these methods to say something like this:

```
It is unspecified whether the returned value is a {@link ListMultimap} or a {@link SetMultimap}.
In cases where that matters, use the corresponding method from {@link ImmutableListMultimap} or {@link ImmutableSetMultimap} as appropriate.
```

What would you think of making that change?

eamonnmcmanus:
Closing due to lack of response.

PR #5892, opened 2022-01-22
Add a `constrain` method to Range.

### LINKED ISSUE #5882: Add a `constrain` method to Range
`com.google.common.primitives.Ints` already has the very useful `static int constrainToRange(int value, int min, int max)`. It would be nice to add this as a method to the `Range` class.

Two advantages:

1. This would be generically available to any `Comparable` type.
2. It just reads better- compare `Range.closed(3,9).constrain(4)` vs `Ints.constrainToRange(3,4,9)`. In the latter case it's not immediately clear which parameter is the input and which are the bounds, but in the former it's super obvious.

One problem is; what should the output of `Range.greaterThan(3).constrain(1)` be?
- On one hand, it seems like `myRange.contains(myRange.constrain(x))`should always be `true`, so maybe we should throw an `IllegalArgumentException` if the *exceeded* bound is open.
- On the other hand, `(3...5].constrain(1) = 3` doesn't seem entirely unreasonable either.

nakulj:
Fixes #5882.

### DIFF
--- a/guava-tests/test/com/google/common/collect/RangeTest.java
+++ b/guava-tests/test/com/google/common/collect/RangeTest.java
@@ -701,4 +701,11 @@ public class RangeTest extends TestCase {
         .addEqualityGroup(Range.closedOpen(1, 7), Range.range(1, CLOSED, 7, OPEN))
         .testEquals();
   }
+
+  public void testConstrain_closed() {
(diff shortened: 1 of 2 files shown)


netdpb:
We should focus on the requirements and the API before writing the implementation.

Issue #5438, opened 2021-03-21
Cache's computeIfPresent discards if error'd entry


ben-manes:
It appears that we still have a bug in the compute logic. The test case below should retain the previous value when the compute throws an error. Instead the failed compute is left in the cache and the prior value can be read. When the next compute occurs, it treats the exception as a null old value, ignores the mappingFunction, and removes the entry.

```java
public void testComputeIfPresent_error() {
  var cache = CacheBuilder.newBuilder().build();
  cache.put(key, "1");
  try {
    cache.asMap().computeIfPresent(key, (key, value) -> { throw new Error(); });
  } catch (Error e) {}
  assertEquals("1", cache.getIfPresent(key));
  assertEquals("2", cache.asMap().computeIfPresent(key, (k, v) -> "2")); // fails; removes mapping
}
```

Caffeine's test suite runs against a Guava shim in order to verify backwards compatibility. I was trying to update that to delegate to Guava's compute implementation after #5406. The other cases pass except, with minor stats differences that I ignore.

netdpb:
Thanks! Do you have a fix in mind? We would test and accept a PR.

ben-manes:
I think we can catch the exception and unwind back to the prior entry, as it holds the segment lock throughout. I don't like how it leaves a zombie, which causes these quirks.

asureshraja:
@netdpb @ben-manes 
I raised pull request for this issue.
Can you please verify 
https://github.com/google/guava/pull/5896

PR #5896, opened 2022-01-25
added try catch in computeifpresent to preserve old value incase of e…

asureshraja:
added try catch in computeifpresent to preserve old value incase of errors
Issue 5438 - https://github.com/google/guava/issues/5438


ben-manes:
Ideally fix the formatting to Google's style guide, but maybe if merged that person will do it.

You might want to run Caffeine's test suite against your new dependency. You can use jitpack.io or a file dependency, and update the non-atomic computes from `GuavaCacheFromContext`. It runs many tests to 1-2hrs if the full suite.

cpovirk:
We ended up going a different direction in https://github.com/google/guava/pull/7980.

PR #5912, opened 2022-02-09
Create aws-codeguru-reviewer.yml

gurudevtest:


nick-someone:
This change isn't CLA-approved, and provides no context as to why this would be needed.

PR #5925, opened 2022-02-16
Add preconditions overloads for double

findepi:
Add overloads for `checkArgument`, `checkState` and `verify` to avoid
boxing of sole `double` format parameter.


findepi:
Relates to https://github.com/airlift/units/pull/23#discussion_r807091300

Issue #5376, opened 2021-01-02
Add conversion from Reader to InputStream


Marcono1234:
It appears there are use cases where one wants to convert a `java.io.Reader` to a `java.io.InputStream` using a given charset, see [this Stack Overflow question](https://stackoverflow.com/questions/65541132/how-to-convert-reader-to-inputstream-in-java) or [this Baeldung article](https://www.baeldung.com/java-convert-reader-to-inputstream#guava).

Guava already has this functionality implemented as [`ReaderInputStream`](https://github.com/google/guava/blob/master/android/guava/src/com/google/common/io/ReaderInputStream.java), but only exposes it for `CharSource` (therefore not for `Reader`), see also https://github.com/google/guava/issues/642#issuecomment-380502353.

Would it make sense to expose it also for `Reader`, e.g. as `com.google.common.io.CharStreams.asInputStream(Reader, Charset)`?

Considerations:
- Users might want to specify buffer size?
- Users very likely want to specify behavior on unencodable chars, currently `ReaderInputStream` uses `CodingErrorAction.REPLACE` for them

Marcono1234:
Yes that looks reasonable, but it would probably be good to have an `asInputStream(Reader, Charset)` method for simple use cases as well.
Though I am not sure if this feature is actually something the maintainers want to add.

ManishOffi:
@Marcono1234 ,
Please can you assign me?

Marcono1234:
@ManishOffi, I am not a member of this project, so I cannot assign you sorry.

louiscb:
Hi @Marcono1234  & @pt657407064 is this issue still relevant? I'd like to contribute and try to fix it with a PR.

eamonnmcmanus:
It is _possible_ for anyone to do this with the existing API, like this:

```java
public static InputStream asInputStream(Reader reader, Charset charset) throws IOException {
  return new CharSource() {
    @Override public Reader openStream() {
      return reader;
    }
  }.asByteSource(charset).openStream();
}
```

Not very elegant, but the `com.google.common.io` API does put the emphasis on reusable sources and sinks rather than one-shot streams like `Reader` and `InputStream`.

louiscb:
Great thanks @eamonnmcmanus. Can I be assigned to the issue?

eamonnmcmanus:
[...]

PR #5939, opened 2022-03-01
#5376 Add conversion from  Reader to InputStream

louiscb:
As per discussions in this issue -> https://github.com/google/guava/issues/5376


eamonnmcmanus:
As mentioned on the issue, we don't think this API is worth adding.

PR #5989, opened 2022-03-27
style: Simplify some code

cedricai:
I found some code that can be simpified by using IDEA to inspect the code, so I submit a PR here.

### DIFF
--- a/guava/src/com/google/common/eventbus/Dispatcher.java
+++ b/guava/src/com/google/common/eventbus/Dispatcher.java
@@ -77,21 +77,10 @@ abstract class Dispatcher {
 
     /** Per-thread queue of events to dispatch. */
     private final ThreadLocal<Queue<Event>> queue =
-        new ThreadLocal<Queue<Event>>() {
-          @Override
-          protected Queue<Event> initialValue() {
-            return Queues.newArrayDeque();
-          }
-        };
+        ThreadLocal.withInitial(() -> Queues.newArrayDeque());
 
     /** Per-thread dispatch state, used to avoid reentrant event dispatching. */
-    private final ThreadLocal<Boolean> dispatching =
-        new ThreadLocal<Boolean>() {
-          @Override
-          protected Boolean initialValue() {
-            return false;
-          }
-        };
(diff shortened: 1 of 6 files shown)


cpovirk:
Thanks. We'll take at least part of this. A few notes:

- I'm not sure why we're inconsistent about fully qualifying the class name `sun.misc.Unsafe`. My guess is that we were worried about `sunapi` warnings from the compiler, which we couldn't suppress on an import? But since we're importing it now, we might as well use it unqualified everywhere.
- We can't used `ThreadLocal.withInitial` yet because it was added in Java 8 and thus isn't available under old versions of Android. (We _could_ use it in `guava-jre` but not `guava-android`, but we generally try to keep the two in sync unless it's very important not to. That said, I think there _are_ weird issues with `ThreadLocal` subclasses and classloaders, so maybe we should consider it here....)
- `FarmHashFingerprint64`: I wonder if the `*2L` actually fixes a bug, albeit one that would manifest only when hashing large arrays? I may investigate. If there's a real issue, we may make that change separately, maybe even introducing a new API for the fixed version.
- `LongAddables`: What the code is doing here (that is, trying to catch static-init failures) is probably already wrong. Changing to a method reference is probably technically not any worse, but we're safer leaving the code alone until we someday fix it to catch failures correctly. (Compare https://github.com/google/truth/commit/948f3edca4dd788dd20065a6582561d510471578, which changed similar to code to use reflection instead.)
- `LongAdder`, `Striped64`: We've mostly tried to keep these in sync with upstream versions, so I'm not sure if we want to make changes to them or not.

cpovirk:
(And you don't need to do anything: We'll make these changes in our internal repo, and they'll be mirrored back out to GitHub automatically, with you listed as the author.)

cpovirk:
Oh, I see: The `FarmHashFingerprint64` changes are only inside methods for hashing short arrays, so there's no concern about overflow there. Still, it's probably good style to avoid anything that might look like it overflows.

cpovirk:
Ok, I think we've now merged the pieces that we can take without taking on any risk. Thanks for the improvements. (We may also see about getting Error Prone to report the 2L issue that intellij pointed out.)

PR #6006, opened 2022-04-10
I find some code is redundant and improvable about Java level and grammar  in my reading code, ans I also find some grammar errors and typo, so I commit this PR to resolve these..

cedricai:


cpovirk:
There are some parts of this that we won't take, for various reasons, some of them covered in https://github.com/google/guava/pull/5989#issuecomment-1085940370.

Also:
- I don't think we want a comma before "so" in the cases you've added it: https://ell.stackexchange.com/a/141848/131135
- I think that that "stats collector" can mean "collector of stats" without an apostrophe.

I may have forgotten other things that we won't take. Still, some nice simplifications here.

jonahgreenthal:
The "its"-to-"it's" change is wrong. (Others might be wrong too; I didn't check that carefully.)

cpovirk:
Thanks, I had missed that one. Both of you are very welcome to check my work in https://github.com/google/guava/pull/6008.

(I will fix the author on that PR now.)

jonahgreenthal:
All the grammar changes in #6008 look correct to me (though a lot of them are unnecessary in the sense that the old version was perfectly correct already).

Issue #5794, opened 2021-11-28
Provide string constants for media types to use in annotations


darioseidl:
It would be nice to have string constants of common media types to use in annotations, for example, for Spring's `@RequestMapping` or Swagger's [`@ApiResponse`](https://github.com/swagger-api/swagger-core/wiki/Swagger-2.X---Annotations#apiresponse). 

Spring has a very limited set of constants in `org.springframework.http.MediaType`, where each media is available as `MediaType` object and as string constant (with suffix `_VALUE`), e.g. `MediaType.APPLICATION_JSON` is the `MediaType` object and `MediaType.APPLICATION_JSON_VALUE` is the string constant `application/json`. That's very convenient, but they only provide a handful of media types.

Guava's `com.google.common.net.MediaType` provides the most complete set of media types that I know of, but sadly cannot be used as compile-time string constants in annotations. There's a [StackOverflow Q/A](https://stackoverflow.com/questions/7854909/interface-enum-listing-standard-mime-type-constants) about this too, in which Guava is recommended, but this limitation is pointed out too.

Would you consider adding such constants? I can imaging writing and maintaining such a list by hand wouldn't be fun, but maybe the code could be generated from the existing `MediaType` constants.

---

A tiny side note: I saw this comment in the code, 
```
  // TODO(gak): make these public?
  private static final String APPLICATION_TYPE = "application";
  private static final String AUDIO_TYPE = "audio";
  private static final String IMAGE_TYPE = "image";
  private static final String TEXT_TYPE = "text";
  private static final String VIDEO_TYPE = "video";

  private static final String WILDCARD = "*";
```
and yea, I think, it would be useful (and I don't see any harm?) if those constants would be public.

PR #6042, opened 2022-05-07
Change private access modifiers for constants in MediaType

Scorpioncik:
Change access modifier to public for

```
private static final String APPLICATION_TYPE = "application";
private static final String AUDIO_TYPE = "audio";
private static final String IMAGE_TYPE = "image";
private static final String TEXT_TYPE = "text";
private static final String VIDEO_TYPE = "video";

private static final String WILDCARD = "*";
```

### DIFF
--- a/android/guava-tests/test/com/google/common/net/MediaTypeTest.java
+++ b/android/guava-tests/test/com/google/common/net/MediaTypeTest.java
@@ -133,7 +133,7 @@ public class MediaTypeTest extends TestCase {
 
   public void testCreate_invalidSubtype() {
     try {
-      MediaType.create("text", "pl@intext");
+      MediaType.create(MediaType.TEXT_TYPE, "pl@intext");
       fail();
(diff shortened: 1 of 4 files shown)


cgdecker:
Is there a particular reason you want these made public? I'm not necessarily opposed, but it would mean having a mix of public `MediaType` and `String` constants in the class, which could be awkward. We'd also need Javadoc on the public fields, etc.

Scorpioncik:
The change is related to these issues :
https://github.com/google/guava/issues/5794

jotka (approved):

isae:
`["IMG-20220925-WA0000.jpg",]("https://lh3.googleusercontent.com/kyhRC5CUw48M2dKmDMWKfWtPzjyCw01L0ji9ubyL9WQR0V0SOe32RiYTzcYaGZ8GmNkfCD3pz9WiWk9Bvkb6IaR5PGr0Y5NrmJTOGQ",) `

PR #6059, opened 2022-05-25
Avoid StackOverflowError in Graphs.hasCycle

perceptron8:
This implementation is based on https://stackoverflow.com/a/46506361/5021051.

I'm not native English speaker, so I left TODO in two places, where I would put javadoc otherwise - let me be excused. For the same reason, class and variable names may be far from perfect, so feel free to change them in any way you like.

### DIFF
--- a/android/guava-tests/test/com/google/common/graph/GraphPropertiesTest.java
+++ b/android/guava-tests/test/com/google/common/graph/GraphPropertiesTest.java
@@ -156,6 +156,18 @@ public class GraphPropertiesTest {
   }
 
   @Test
+  public void hasCycle_noStackOverflowError() {
+    int enoughForEventualStackOverflowError = 100_000;
+    for (MutableGraph<Integer> graph : graphsToTest) {
+      for (int i = 0; i < enoughForEventualStackOverflowError; i++) {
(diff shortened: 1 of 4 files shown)


jrtom:
Are you proposing this change because you encountered stack overflows in this code in actual operation, or just because you believe that it's a better design in the abstract?
* If you have encountered this in practice, some additional context would be helpful if you're free to share it.

perceptron8:
More in the abstract, it wasn't actual operation. I've encountered stack overflow in one of the (purposely mean) tests for DFS-based topological ordering I was implementing recently, where `!Graphs.hasCycle` is being used as one of preconditions.

I know that the iterative code is slightly less readable than it's recursive version, but I believe that `Graph.hasCycle` should be ready for every kind of graph possible, event it is some less `common.graph` ;-) It's up to you obviously and I don't insist for this change - you may as well wait for SOE in some real world scenario.

ahmed82 (approved):
lgtm

jrtom:
I think that, given the moderate additional code complexity of your approach, I'd prefer to leave it "as is" for now.  If we do encounter this issue in the wild, we'll reopen this.  Thanks for the suggestion.  :)

PR #6079, opened 2022-06-15
Switch CacheLoader to accepting Java Functions and Suppliers

ccrvincent:
Tried to use CacheLoader#from using a j.u.f.Function and realized this didn't get updated.


chaoren:
Unfortunately, we can't use `java.util.function.*` yet for compatibility reasons since that requires JDK 1.8 or above.

PR #6083, opened 2022-06-18
Optimize getLast function for Linked list to be O(1)

ArthurGazizov:


jbduncan:
@chaoren I might be missing something, but surely the presence of the loops indicates that the running time of `LinkedList.node()` is O(n/2), i.e. O(n)?

chaoren:
`index` would be `size - 1`, so the loop ends in one iteration.

jbduncan:
@chaoren Oh, of course, you're right! Thanks for clearing things up for me. :)

Issue #3422, opened 2019-03-25
Remove @Beta from Ints.tryParse and Longs.tryParse


cpovirk:
- We would have to wait a while to switch it to OptionalInt.... But we could add differently named "overloads" now to try to start shifting users that way....
- Accept CharSequence? https://github.com/google/guava/issues/2549
- Any regrets over the overload "Long.tryParse(foo, 10)," where "10" might look like a default value rather than a radix?

kluever:
This is now fixed.

PR #6103, opened 2022-07-10
Remove @Beta from Ints.tryParse and Longs.tryParse

ananthvyas:
First go at contributing here. Addressing https://github.com/google/guava/issues/3422


cpovirk:
Finally done for [32.0.0](https://github.com/google/guava/releases/tag/v32.0.0).

PR #6117, opened 2022-07-16
I optimization algorithm

luohongtu:
I optimization algorithm


luohongtu:
hashcode & (n - 1), We want to make sure that n goes to the power of 2

cpovirk:
I think we're happy to let classes like `HashMap` round up to the next power of 2 when necessary. That lets us keep this code simpler. Let me know if I'm overlooking something, though.

PR #6188, opened 2022-09-28
Fix Spliterator implementations for sorted collections

### LINKED ISSUE #6187: Spliterators for sorted collections cause inefficient stream operations when natural ordering is used
The Spliterators returned by sorted collections such as `ImmutableSortedSet` and `RegularImmutableAsList` result in unnecessary sorting occurring when a sort operation is performed on their associated Stream. The root cause is that the
Spliterators' `getComparator()` implementations always return a `Comparator` instance, even when natural ordering is used. 

According to the [Javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/Spliterator.html#getComparator--), `null` should be returned by `getComparator()` to indicate the source items are sorted in natural order.

You can see [here](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/stream/StreamOpFlag.java#L751-L755) that the `SORTED` flag gets unset when a spliterator has the `SORTED` characteristic but does not return a `null` comparator. And [here](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/stream/SortedOps.java#L136-L139) is where the missed optimization happens due to that.

The issue is trivially reproducible by doing the following, and stepping through the code in SortedOps where the expected no-op / optimization would ideally happen:

```java
ImmutableSortedSet<Integer> sortedSet = ImmutableSortedSet.of(1, 2, 3, 4);
// sorted should be a no-op, but isn't because ImmutableSortedSet's Spliterator does not return a null comparator.
sortedSet.stream().sorted().collect(Collectors.toList());
```

kilink:
Spliterators returned by CollectSpliterators.indexed and ImmutableSortedSet.spliterator violated the Spliterator API by not returning null in getComparator when the source items are naturally sorted. The effect was that sort operations on Streams backed by these Spliterators were not optimized away, resulting in additional unnecessary sorting.

Fixes #6187.


guava/src/com/google/common/collect/CollectSpliterators.java
```diff
@@ -96,7 +96,7 @@ public int characteristics() {
       @CheckForNull
       public Comparator<? super T> getComparator() {
         if (hasCharacteristics(Spliterator.SORTED)) {
-          return comparator;
+          return Ordering.natural().equals(comparator) ? null : comparator;
```
kevinb9n, line 99: We should special-case Comparator.naturalOrder() too.

kilink: Okay I've updated it to handle `Comparator.naturalOrder()` as well.

Issue #6176, opened 2022-09-19
`MediaType.parse` doesn't like `\r`


boris-petrov:
Guava 31.1-jre, JDK 18.

```java
MediaType.parse("application/pdf; \rname=\"foo\rbar.pdf\"")
```

Leads to an exception:
```
Caused by: java.lang.IllegalArgumentException: Could not parse 'application/pdf; 
name="foo
bar.pdf"'
	at com.google.common.net.MediaType.parse(MediaType.java:1078)
        ...
Caused by: java.lang.IllegalStateException
	at com.google.common.base.Preconditions.checkState(Preconditions.java:486)
	at com.google.common.net.MediaType$Tokenizer.consumeToken(MediaType.java:1101)
	at com.google.common.net.MediaType.parse(MediaType.java:1066)
```

I understand that these `\r` symbols don't make much sense there but I've seem them in the wild (coming from Java Mail) so it would be nice to support them.

kluever:
I don't think it should be the responsibility of the parser to sanitize inputs or trim white space.

boris-petrov:
But it handles correctly other kinds of whitespace, e.g. `MediaType.parse("application/pdf;    \n\n\nname=\"foo\n\nbar.pdf\"")` works totally fine. Why not fix it also for `\r`?

kluever:
From the class docs:

> Note that this specifically ... has no support for header-specific considerations such as line folding and comments".

Also it looks like the linear-white-space was chosen specifically:
```
  /*
   * This matches the same characters as linear-white-space from RFC 822, but we make no effort to
   * enforce any particular rules with regards to line folding as stated in the class docs.
   */
  private static final CharMatcher LINEAR_WHITE_SPACE = CharMatcher.anyOf(" \t\r\n");
```

/cc @cpovirk in case he has any opinions.

PR #6212, opened 2022-10-14
6176 | Fix for "MediaType.parse doesn't like \r"

pawel-janowski:
Improved method MediaType::parse to remove linear whitespaces from the attribute value during parsing.

This fixes issue https://github.com/google/guava/issues/6176


kluever:
As I mentioned on the bug, I don't think the parser should be responsible for sanitizing inputs.

Issue #6214, opened 2022-10-14
Strings.lenientFormat / Preconditions: Handling of attempts to use non-%s specifiers could perhaps be handled better


captainbkarthick:
https://github.com/google/guava/blob/989e3b261eb3c67b42c97f612b5d726f03e1e300/guava/src/com/google/common/base/Strings.java#L268

The method `lenientFormat(String template, Object... args)` has been called from many methods in the Preconditions.java class with different data type inputs as arguments.

This method internally converts all the arguments passed as varargs to String. Refer: https://github.com/google/guava/blob/989e3b261eb3c67b42c97f612b5d726f03e1e300/guava/src/com/google/common/base/Strings.java#L285

Though the method converts the arguments to string, the template String having other data type notations (eg., %d, %f etc.,) are missed to be converted.

This cause a wrong formatted text to be returned.

**Possible Solution:**
Create a method to replace the string %d/%f/... with %s 
or
Implement System.out.printf() logic to format the string.

PiyushKeshari24:
Please assign this to me

kevinb9n:
This is actually intentional behavior, for performance and code size reasons. String.format() is actually quite complex. Also, our core philosophy toward Preconditions has *always* been that whenever it's not quite right for you needs, it's 100% fine and good to return to the if/throw idiom. Preconditions is doing its job by making that need a lot less common.

captainbkarthick:
@kevinb9n 
The below code will not return the desired data
`Preconditions.checkArgument(false, "Port `%d` of %s must be a positive number", 8080, "Data");`

Ideally the above should return
**Port 8080 of Data must be a positive number**
Whereas the code will return
**Port %d of 8080 must be a positive number [Data]**

I feel that this is an issue which must be addressed.

kevinb9n:
Ah I see! What I said before still stands, but you do raise a valid point here. The fact that the 8080 appears in the wrong place can be misleading.

There could be something simple we could do, without bloating code size, that makes it significantly more _likely_ that the values come out in the right place. To give one *example*, it could:

* When the number of `%s` matches the number of args given, behave as today
* When it does not, check if the count of `%` matches the number of args; if so, assume that each represents an "attempt", and thereby make the correct matching between valid placeholders and arguments. The others would still appear at the end in brackets as they currently do.

perceptron8:
@captainbkarthick You may like to use [Error Prone](https://errorprone.info/) in your compilation. Two Checkers will catch this kind of bugs:
[...]

PR #6221, opened 2022-10-19
[ISSUE #6214] Strings.lenientFormat() - Enable handling of all datatype specifiers while formatting

captainbkarthick:


cgdecker:
As noted on the issue, this isn't something we're going to change.

PR #6266, opened 2022-12-08
Removed @Beta from HashingInputStream class

sebastienrospars:
Removed the Beta annotations from HashingInputStream. This class have been in beta since versions 16.0.


sebastienrospars:
Hi, do you have any update about this PR ? Thanks by advance

Issue #6296, opened 2023-01-16
Support ExecutorService.close in ForwardingExecutorService


findepi:
JDK's `ExecutorService` is `AutoCloseable` since Java 19.
`ForwardingExecutorService` does not forward the `close` method.

Forward close too, when running on Java 19.

findepi:
It's not hard to implement that -- I can file a PR -- but obviously not without reflection.

i searched through https://github.com/google/guava/wiki and https://github.com/google/guava/wiki/CollectionHelpersExplained to understand what's the project policy on such cases, but didn't find an answer (apologies, if i just overlooked it!).
I also searched open issues and PRs for "ExecutorService close" and "19", but there too didn't find a prior discussion.

@cpovirk is there a project policy on how to update `Forwarding*` classes when newer Java version has new methods?

cpovirk:
Our approach so far has been not to forward `default` methods (except for the case in which [a method already existed and merely _became_ a `default` method afterward](https://github.com/google/guava/commit/cee9942d5ac735d8d143dad9cee019670f3b8ada#diff-db179e5f5bacd8b115711f86d508e334cad27a9f2ad5c4847776fae0cef1df44R30)).

In the case of `ExecutorService.close`, the motivation for not forwarding is that someone might use `ForwardingExecutorService` to implement an `UnshutdownableExecutorService` by overriding `shutdown` and `shutdownNow`. If `ForwardingExecutorService` began to forward `close` calls to the underlying executor, then that would unexpectedly provide a way to shut down the executor.

This is definitely a wart, but it's probably our best option, given that `default` methods mess a little bit with our original assumptions about interfaces. (And it's another weird thing about the `Forwarding*` classes in general.)

From your linked PR (thanks!), I gather that you filed this issue at least partially because of a failing test (which maybe could be changed to ignore `default` methods... if not for the exception I mentioned above) but also partially because you're asking whether it's better for forward in this case or not. Let me know if there's other context that I missed from skimming over it.

In any case, we should document this behavior, as we did for `ForwardingCollection` and friends way back when.

findepi:
Thanks @cpovirk for your detailed explanation. It makes total sense.

> From your linked PR (thanks!), I gather that you filed this issue at least partially because of a failing test (which maybe could be changed to ignore default methods... if not for the exception I mentioned above) 

[...]

PR #6299, opened 2023-01-17
Document ForwardingExecutorService behavior for default methods

findepi:
Currently, Guava is built with Java 9 and `ExecutorService` has no default methods. This changes in Java 19, where `ExecutorService.close` is added. Ensure users have right expectations about future evolution of the `ForwardingExecutorService` and its compatibility with newer Java versions.

Fixes https://github.com/google/guava/issues/6296


findepi:
Thanks @cpovirk for adding the note to other classes, and for the merge!

Issue #6067, opened 2022-05-31
Add LineReader.lines() for parity with BufferedReader


JanecekPetr:
`LineReader` currently only has the `readLine()` method. However, already since Java 8 there's [`lines()`](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/io/BufferedReader.html#lines()), too.

Not to mention that it'd be nice to also maybe have `readAllLines()` and maybe even a `getLineNumber()` from `LineNumberReader`.

Yes, I know, there's `CharStreams.readLines()`, that's fine.

ritikBhandari:
so there are 3 methods to be implemented: `lines()`, `readAllLines()` and `getLineNumber()`. Right?

JanecekPetr:
Not really. I will say `lines()` definitely is needed to bring `LineReader` to where `BufferedReader` is. The other methods are up for a debate as they could be slightly controversial / unneeded:
- How often do you _really_ need a line count? It's probably not worth it for the component to keep track of it all the time if only a small percentage of use-cases will use it.
- It's unclear whether `readAllLines()` belongs here because `CharStreams.readLines()` exists. Doing one thing multiple ways is usually frowned upon in good API design. That said, `LineReader` kinda screams for the method. Not my call. This requires more API designing and discussions and, let's be frank, nobody is going to spend precious time on arguing whether the method belongs there or not, so it's not going to be there.

Let's only focus on `lines()` here.

ritikBhandari:
Makes sense. So its `lines()` for now. 
Should I resolve it?

maxmielchen:
I have added the lines method in this pull request https://github.com/google/guava/pull/6142
Can anyone approve it.

ayushsinghal90:
I have raised a cr for this.

PR #6308, opened 2023-01-22
added lines() in LineReader

ayushsinghal90:
- Added lines() in LinedReader.
- Modified LineBufferTest class to test lines() function.

Solves https://github.com/google/guava/issues/6067

### DIFF
--- a/guava-tests/test/com/google/common/io/LineBufferTest.java
+++ b/guava-tests/test/com/google/common/io/LineBufferTest.java
@@ -16,6 +16,8 @@
 
 package com.google.common.io;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
+
 import com.google.common.base.Function;
 import com.google.common.collect.Lists;
 import java.io.BufferedReader;
@@ -26,7 +28,6 @@ import java.io.StringReader;
 import java.nio.CharBuffer;
 import java.util.Arrays;
 import java.util.List;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 /**
@@ -76,10 +77,13 @@ public class LineBufferTest extends IoTestCase {
       assertEquals(expectRead, readUsingJava(input, chunk));
       assertEquals(expectRead, readUsingReader(input, chunk, true));
       assertEquals(expectRead, readUsingReader(input, chunk, false));
-      assertTrue(expectRead.containsAll(readUsingReaderGetLines(input, chunk, true)
-          .collect(Collectors.toList())));
-      assertTrue(expectRead.containsAll(readUsingReaderGetLines(input, chunk, false)
-          .collect(Collectors.toList())));
+      try (Stream<String> linesStream = readUsingReaderGetLineReader(input, chunk, true).lines()) {
+        assertTrue(expectRead.containsAll(linesStream.collect(toImmutableList())));
+      }
+
+      try (Stream<String> linesStream = readUsingReaderGetLineReader(input, chunk, false).lines()) {
+        assertTrue(expectRead.containsAll(linesStream.collect(toImmutableList())));
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


guava/src/com/google/common/io/LineReader.java
```diff
@@ -84,4 +94,43 @@ public String readLine() throws IOException {
...
+                        nextLine = Optional.ofNullable(readLine()).orElse("");
+                        return (!nextLine.equals(""));
+                    } catch (IOException e) {
+                        throw new UncheckedIOException(e);
+                    }
+                }
+            }
+            @Override
+            public String next() {
+                if (!nextLine.equals("") || hasNext()) {
```
SplotyCode, line 124: The `!nextLine.equals("")` condition should be already covered by the hasNext() right?

guava/src/com/google/common/io/LineReader.java
```diff
@@ -84,4 +94,43 @@ public String readLine() throws IOException {
...
+    /**
+     * Returns a {@code Stream}, the elements of which are lines read from
+     * this {@code LineReader}.
+     *
+     * @return a {@code Stream<String>} providing the lines of text
+     *         described by this {@code LineReader}
+     */
+    public Stream<String> lines()
+    {
+        Iterator<String> iterator = new Iterator<String>() {
```
SplotyCode, line 107: Perhaps using the [AbstractIterator](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/AbstractIterator.java) would make the logic quite a bit simpler

guava/src/com/google/common/io/LineReader.java
```diff
@@ -14,19 +14,29 @@
 
 package com.google.common.io;
 
-import static com.google.common.base.Preconditions.checkNotNull;
```
SplotyCode, line 17: Did you chance the import order accidentally? [Google Java Styleguide](https://google.github.io/styleguide/javaguide.html#s3.3-import-statements)

ayushsinghal90: yes, fixed.

guava-tests/test/com/google/common/io/LineBufferTest.java
```diff
@@ -121,6 +127,13 @@ private static List<String> readUsingReader(String input, int chunk, boolean asR
     return lines;
   }
 
+  private static Stream<String> readUsingReaderGetLines(String input, int chunk, boolean asReader) {
+      Readable readable =
+          asReader ? getChunkedReader(input, chunk) : getChunkedReadable(input, chunk);
```
SplotyCode, line 132: When looking through the code base, this is the way they split the ternary operator
```suggestion
      Readable readable = asReader 
          ? getChunkedReader(input, chunk) 
          : getChunkedReadable(input, chunk);
```

guava/src/com/google/common/io/LineReader.java
```diff
@@ -84,4 +91,31 @@ public String readLine() throws IOException {
     }
     return lines.poll();
   }
+
+  /**
+   * Returns a {@code Stream}, the elements of which are lines read from
+   * this {@code LineReader}.
+   *
+   * @return a {@code Stream<String>} providing the lines of text
+   *     described by this {@code LineReader}
```
SplotyCode, line 100: I believe the JavaDoc should:

- state that the caller is in charge of closing the Stream
- mention the fact that we throw UncheckedIOException
- what we define as a line (maybe just mention that a line has the same semantics as in LineReader#lines

SplotyCode (commented):
I am not a project maintainer. I just wrote the things that looked obvious to me. But since you added me as a reviewer i will try my best. But please take the comments as a 'open for discussion' rather then a todo

guava-tests/test/com/google/common/io/LineBufferTest.java
ayushsinghal90: I referred this function `readUsingReader`.

guava/src/com/google/common/io/LineReader.java
```diff
@@ -84,4 +94,50 @@ public String readLine() throws IOException {
...
+   * try (Stream<String> lines = source.lines()) {
+   *   lines.map(...)
+   *      .filter(...)
+   *      .forEach(...);
+   * }
+   * }</pre>
+   *
+   * @return a {@code Stream<String>} providing the lines of text
+   *     described by this {@code LineReader}
+   * @throws UncheckedIOException if an I/O error occurs.
```
SplotyCode, line 121: The method itself does not throw `UncheckedIOException` right?

PR #6311, opened 2023-01-25
Replace deprecated gradle instructions

ehenoma:
The gradle docs state that
> Dependencies should no longer be declared using the compile and runtime configurations
https://docs.gradle.org/current/userguide/upgrading_version_5.html#changes_6.0

### DIFF
--- a/index.md
+++ b/index.md
@@ -42,7 +42,7 @@ To add a dependency using Gradle:
 
 ```gradle
 dependencies {
-  compile 'com.google.guava:guava:{{ site.latest_release }}-jre'
+  implementation 'com.google.guava:guava:{{ site.latest_release }}-jre'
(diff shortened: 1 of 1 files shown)


cpovirk:
Thanks. We should use the same text here as we do [on our GitHub landing page](https://github.com/google/guava/blob/master/README.md#adding-guava-to-your-build). Let me see if I can change around our docs setup so that we use the same internal source file for both.

cpovirk:
> Let me see if I can change around our docs setup so that we use the same internal source file for both.

Not easily: I had been thinking that we had internal automation for our gh-pages branch, but what we actually have is automation for our wiki. We could set up such automation for our gh-pages branch, but for now, I just copied over the text.

PR #6326, opened 2023-02-14
Add Maven Wrapper to our repo.

cpovirk:
For now, I'm adding the wrapper for Maven 3.8.2, which is the version
most of us at Google have handy at the moment. Arguably we should switch
to 3.9.0, since much the point of the wrapper is to make us not have to
worry about which version we have installed :)


netdpb:
Curious: Why are we doing this as a standard PR rather than through Copybara?

cpovirk:
There's some explanation in CL 509334407. We'll see if Nick buys it :)

Marcono1234:
Is this obsolete now due to #6783?

cpovirk:
Oh, yes, thanks.

PR #6337, opened 2023-02-23
Avoid unnecessary interruptions

pgpego:
Modified all uninterruptible methods so that the thread interrupt flag is cleared (and remembered) at the start of the method. This way no unnecessary InterruptedException is thrown and no extra iteration is performed if the thread was already interrupted when the method is called.

### DIFF
--- a/android/guava/src/com/google/common/util/concurrent/Uninterruptibles.java
+++ b/android/guava/src/com/google/common/util/concurrent/Uninterruptibles.java
@@ -52,7 +52,7 @@ public final class Uninterruptibles {
   /** Invokes {@code latch.}{@link CountDownLatch#await() await()} uninterruptibly. */
   @GwtIncompatible // concurrency
   public static void awaitUninterruptibly(CountDownLatch latch) {
-    boolean interrupted = false;
+    boolean interrupted = Thread.interrupted();
     try {
       while (true) {
         try {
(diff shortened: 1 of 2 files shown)


cpovirk:
Thanks. I think we'd considered this at one point and worried that it would actually be slower in the common case in which the thread was not already interrupted. So the question is how often the thread is already interrupted--and then how much this PR saves in that case vs. how much it costs in the common case. Do you have any theories on that?

pgpego:
Oh, not really. I see your point and all that I can say is that maybe I was considering only my own use cases, where interruption is really a frequent thing, and I may have thought that this class was intended just for those situations. But thinking more carefully, it does not seem unlikely to use methods from this class, for example, just to avoid the verbosity of having to deal with InterruptedException in situations where the thread is actually never interrupted.

cpovirk:
Now that you mention it, we call `getUninterruptibly(...)` from `Futures.getDone(...)`, which we use pretty widely inside Google, so that would be one place where we'd see a lot more polling of interrupts. (That said, I guess we could give `Futures.getDone(...)` its own copy of `getUninterruptibly(...)` with slightly different performance characteristics if we wanted.)

Now, you could reasonably protest that `Futures.getDone(...)` _already_ performs an extra `volatile` read that callers don't "need": It checks `isDone()` before calling `getUninterruptibly` :) So maybe it's just not the method to call if you really really really really care about performance.

Maybe what we _really_ should do if we want to optimize that tightly is to give `Futures.getDone(...)` a special case for implementations of `AbstractFuture.Trusted`: It could pull the value out with no separate `isDone()` check and no interrupt checks _even if the thread is interrupted_. (That could be especially nice if an app calls `getDone(...)` on a bunch of futures in sequence when the current thread has already been interrupted.) For that matter, `AbstractFuture.get` could have a no-interrupt path that `getUninterruptibly(...)` could call...... And I am getting further and further off-topic here, plus I'm focusing heavily on the `Future` case.

I think we probably should leave the main `Uninterruptibles` code paths alone. There may be room for some other optimizations in this area, but judging from a couple searched I just did over our profiling data, most of those optimizations might be more "fun for me" than "useful" :)

Issue #3913, opened 2020-05-29
Maps.difference(..., BiPredicate) in addition to (..., Equivalence)


cpovirk:
Might not be worth the trouble for such a lightly used method. But a recent internal review (CL 313188882 + CL 313657501, if anyone internally is curious) drove home to me that we should probably err on the side of accepting `BiPredicate` over `Equivalence` (at least when we can assume Java 8+ APIs): It doesn't require defining a hash function, and it can be created with a lambda or method reference (unlike `Equivalence` [at present](https://github.com/google/guava/issues/2785#issuecomment-634056822)).

PR #6415, opened 2023-04-12
Fix issue #3913 Add Maps.difference(..., BiPredicate)

remeio:
Hello, Fix issue #3913 It can be created with a lambda or method reference, such as String::equals

### DIFF
--- a/android/guava/src/com/google/common/util/concurrent/FluentFuture.java
+++ b/android/guava/src/com/google/common/util/concurrent/FluentFuture.java
@@ -16,7 +16,6 @@ package com.google.common.util.concurrent;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
-import com.google.common.annotations.Beta;
 import com.google.common.annotations.GwtCompatible;
 import com.google.common.annotations.GwtIncompatible;
 import com.google.common.annotations.J2ktIncompatible;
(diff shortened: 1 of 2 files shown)


guava/src/com/google/common/collect/Maps.java
```diff
@@ -523,6 +524,38 @@ MapDifference<K, V> difference(
...
+   * a later time.
+   *
+   * <p>Since this method uses {@code HashMap} instances internally, the keys of the supplied maps
+   * must be well-behaved with respect to {@link Object#equals} and {@link Object#hashCode}.
+   *
+   * @param left the map to treat as the "left" map for purposes of comparison
+   * @param right the map to treat as the "right" map for purposes of comparison
+   * @param equivalence the equivalence relationship to use to compare values
+   * @return the difference between the two maps
+   * @since 10.0
```
Stephan202, line 539: Time travel ;)

guava/src/com/google/common/collect/Maps.java
```diff
@@ -523,6 +524,38 @@ MapDifference<K, V> difference(
...
+    return difference(left, right, new Equivalence<V>() {
+      @Override
+      protected boolean doEquivalent(V a, V b) {
+        return equivalence.test(a, b);
+      }
+      @Override
+      protected int doHash(V v) {
+        return v.hashCode();
+      }
+    });
```
Stephan202, line 556: Instead of creating this custom `Equivalence` just for `Maps#difference`, would it be an idea to instead introduce a new static factory method `Equivalence#equals(BiPredicate)`? That would allow the new feature to be combined with all other `Equivalence`-accepting APIs. (Downside may be that this alternative is less "discoverable" :thinking:.)

remeio: Good idea! I also think the static factory is better, because this is just a lightly used method. I will change the code and pull request again.

Stephan202:
This PR was replaced by #6416.

PR #6416, opened 2023-04-12
Fix issue #3913 Add Equivalence#predicate

remeio:
@Stephan202 Hello, I had changed the code using a new static factory method `Equivalence#predicate(BiPredicate)` to fix the issue #3913. 
It will make a distinction between `equas(BiPredicate)` and `equas(Object)`, so I use the `predicate` instead of `equals`.

### DIFF
--- a/android/guava-testlib/src/com/google/common/collect/testing/AbstractIteratorTester.java
+++ b/android/guava-testlib/src/com/google/common/collect/testing/AbstractIteratorTester.java
@@ -31,6 +31,7 @@ import java.util.NoSuchElementException;
 import java.util.Set;
 import java.util.Stack;
 import junit.framework.AssertionFailedError;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 /**
  * Most of the logic for {@link IteratorTester} and {@link ListIteratorTester}.
@@ -148,7 +149,7 @@ abstract class AbstractIteratorTester<E, I extends Iterator<E>> {
      * this to determine which stack to pop from on a call to {@code remove()} (or to pop from and
      * push to on a call to {@code set()}).
      */
-    Stack<E> stackWithLastReturnedElementAtTop = null;
+    @Nullable Stack<E> stackWithLastReturnedElementAtTop = null;
 
     MultiExceptionListIterator(List<E> expectedElements) {
       Helpers.addAll(nextElements, Helpers.reverse(expectedElements));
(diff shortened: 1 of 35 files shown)


guava/src/com/google/common/base/Equivalence.java
```diff
@@ -398,4 +409,29 @@ private Object readResolve() {
...
+    }
+
+    @Override
+    protected int doHash(V v) {
+      return v.hashCode();
+    }
+
+    public static <V> BiPredicateEquivalence<V> of(BiPredicate<? super V, ? super V> predicate) {
+      return new BiPredicateEquivalence<>(predicate);
+    }
```
Stephan202, line 433: This method doesn't need to be `public`, since the class isn't. (Maybe this method isn't necessary all all, since it merely delegates to the constructor?)

guava/src/com/google/common/base/Equivalence.java
```diff
@@ -345,6 +345,17 @@ public static Equivalence<Object> equals() {
     return Equals.INSTANCE;
   }
 
+  /**
+   * Returns an equivalence that delegates to {@link BiPredicate#test(Object, Object)} and {@link Object#hashCode}.
+   * {@link Equivalence#equivalent} returns {@code true} if the input values match the predicate, including in the case that a and b are both null. {@link Equivalence#hash} returns
+   * {@code 0} if passed a null value.
+   *
+   * @since 31.1
```
Stephan202, line 353:
```suggestion
   * @since NEXT
```
(This will be replaced before release.)

guava/src/com/google/common/base/Equivalence.java
```diff
@@ -345,6 +345,17 @@ public static Equivalence<Object> equals() {
     return Equals.INSTANCE;
   }
 
+  /**
+   * Returns an equivalence that delegates to {@link BiPredicate#test(Object, Object)} and {@link Object#hashCode}.
+   * {@link Equivalence#equivalent} returns {@code true} if the input values match the predicate, including in the case that a and b are both null. {@link Equivalence#hash} returns
+   * {@code 0} if passed a null value.
```
Stephan202, line 351: Looks like some reformatting / line wrapping is in order here :)

Stephan202 (commented):
I'm not a Guava maintainer, so will leave further review to @cpovirk (as the author of #3913 ).

cpovirk:
Thanks. We have talked about providing an easier way to create a `BiPredicate` from a lambda, as in https://github.com/google/guava/issues/2785#issuecomment-634056822. It would probably need a way to provide a custom implementation of hashing, too.

That said, we'd want to do some more research to see how often users need this. It's possible that something more like your original approach of #6415 would be a better alternative, or maybe we'd be able to justify both. At the moment, we probably wouldn't have a look unless we hear from potential users of this. (My issue #3913 was fairly speculative.) In the meantime, I'll keep this open so that we have a possible implementation handy.

PR #6422, opened 2023-04-14
Updated the README.md Lines 8 and 27. Possible mistakes corrected.

RabinChakraborty:
1. I noticed that "multimap" and "multiset" should be capitalized as "Multimap" and "Multiset" as they are the names of specific collection types in Guava Library. (line 8.).
2. I noticed one minor mistake: The correct version field for the Android flavor of Guava is actually specified as 31.1.1-android (with an extra ".1" after the version number). (line 27).

I only made the above changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,16 +5,16 @@
 [![OpenSSF Best Practices](https://bestpractices.coreinfrastructure.org/projects/7197/badge)](https://bestpractices.coreinfrastructure.org/projects/7197)
 
 Guava is a set of core Java libraries from Google that includes new collection types
-(such as multimap and multiset), immutable collections, a graph library, and
+(such as Multimap and Multiset), immutable collections, a graph library, and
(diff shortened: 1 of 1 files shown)


cpovirk:
Thanks.

For the collection types, we generally try to use a "plain word" when it makes sense. For example, the README also refers to "strings" instead of "`String`s" or "`String` instances." "Multimap" and "multiset," while not exactly _common_ words, are "enough of words" that we use them instead of the class names.

For the version number, I believe the latest is in fact only 31.1:

https://repo1.maven.org/maven2/com/google/guava/guava/

Are you looking at 30.1.1? Or maybe your organization has a custom patched version of 31.1?

(One bit of somewhat related good news is that we put some checks in place during our release process to force us to use 3-part version numbers, as Semantic Versioning specifies. So the next one will be "32.0.0" rather than "32.0.")

PR #6445, opened 2023-04-21
Refactor use System.arraycopy() to copy array.

remeio:


codeboyzhou:
I have a question about why should we use `System.arraycopy()` in this code snippet, for most small arrays, e.g. length less than 50, `for-loop` is generally faster than system array copy according to benchmark test.

kevinb9n:
Not sure the change was a bad idea. (Not sure it's worth changing either, but...)

remeio:
For case 1:
`'x' -> "length less than 50".toCharArray()`, the `for-loop` is better.
For case 2:
`'y' -> "length is too long, .................................................".toCharArray()`, the `System#arrayCopy` is better.
Such as.
```java
CharEscaper charEscaper =
        createSimpleCharEscaper(
            ImmutableMap.<Character, char[]>builder()
                .put('x', "length less than 50".toCharArray())
                .put('y', "length is too long, .................................................".toCharArray())
                .put('z', "<lo>".toCharArray())
                .buildOrThrow());
    UnicodeEscaper unicodeEscaper = Escapers.asUnicodeEscaper(charEscaper);
    unicodeEscaper.escape("xxxxxxx")
``` 
Usually, the case 1 is better way, do you think the case 2 is necessary?

abimarank:
`System.arraycopy()` should be faster as it's a native call.

cpovirk:
I dug up some links for fun:

- https://bugs.openjdk.org/browse/JDK-6912521
- https://bugs.openjdk.org/browse/JDK-8149758
- https://bugs.openjdk.org/browse/JDK-8150730
- https://stackoverflow.com/q/32834869/28465 (from probably before the improvement landed)

It looks like this code is rarely used (if ever) inside Google, so much so that it doesn't even appear in our global profiling data. I'm also not sure that it's actually even _reachable_ in the open-source release, so maybe I should just remove it....

PR #6454, opened 2023-04-25
Refactor use Integer#compare and Long#compare

remeio:


cpovirk:
I am surprised that [our Animal Sniffer check](https://github.com/google/guava/blob/f9e5c5d82a09a9d2c49a4e6c09aa08df04f0c217/android/pom.xml#L183) doesn't flag this. I wonder if that's a bug in the check....

That said, I believe that these methods actually _are_ safe to use, even under old versions of Android. See https://github.com/google/truth/blob/e48b2161560c7f1a1b83b7d81a2eb4b5eea4ee2a/core/src/test/java/com/google/common/truth/IterableSubjectTest.java#L1169-L1174

I would probably leave as they are and try to figure out why Animal Sniffer isn't catching this.

remeio:
> I am surprised that [our Animal Sniffer check](https://github.com/google/guava/blob/f9e5c5d82a09a9d2c49a4e6c09aa08df04f0c217/android/pom.xml#L183) doesn't flag this. I wonder if that's a bug in the check....
> 
> That said, I believe that these methods actually _are_ safe to use, even under old versions of Android. See https://github.com/google/truth/blob/e48b2161560c7f1a1b83b7d81a2eb4b5eea4ee2a/core/src/test/java/com/google/common/truth/IterableSubjectTest.java#L1169-L1174
> 
> I would probably leave as they are and try to figure out why Animal Sniffer isn't catching this.

ok

chaoren:
Yeah, I think this change defeats the purpose of the two methods:

```
   * <p><b>Note for Java 7 and later:</b> this method should be treated as deprecated; use the
   * equivalent {@link Integer#compare} method instead.
```

```
   * <p><b>Note for Java 7 and later:</b> this method should be treated as deprecated; use the
   * equivalent {@link Long#compare} method instead.
```

If `Integer#compare` or `Long#compare` were available, the caller should just be using those instead.

PR #6457, opened 2023-04-26
mergeStableSorted was added to create a stable version of mergeSorted…

### LINKED ISSUE #5773: Make `Iterators.mergeSorted()` stable

igmtz:
… in Iterators

I'm adding an option in Iterators for stable sort in mergeSorted with documentation. Resolves #5773

### DIFF
--- a/guava/src/com/google/common/collect/Iterators.java
+++ b/guava/src/com/google/common/collect/Iterators.java
@@ -34,19 +34,8 @@ import com.google.common.base.Preconditions;
 import com.google.common.base.Predicate;
 import com.google.common.primitives.Ints;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
-import java.util.ArrayDeque;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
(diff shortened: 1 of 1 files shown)


chaoren:
We haven't decided that we would accept the proposed changes yet. Please avoid sending pull requests before the decision is made to proceed with a proposal.

Issue #6510, opened 2023-05-24
Add some annotations from "On Leveraging Tests to Infer Nullable Annotations" data


cpovirk:
After a conversation with @jensdietrich in [an Error Prone issue](https://github.com/google/error-prone/issues/3792), Jens provided us with some data on Guava fields, parameters, and method returns that see null values during the course of testing (aside from tests that are checking that NPE occurs :)). That comes from the paper in the title:

> Jens Dietrich, David Pearce and Mahin Chandramohan: On Leveraging Tests to Infer Nullable Annotations. ECOOP’23.

I've looked at the data for a few classes already and submitted some changes:

- [`ClassToInstanceMap` and `TypeToInstanceMap` had previously had non-nullable types because we were limiting which annotations we used.](https://github.com/google/guava/commit/2b98d3c1e96b750dc997c29f283084aeb72fb3cf)
- [`MapMakerInternalMap` was missing many annotations (and is very likely still be missing a few) because we hadn't tried to annotate it.](https://github.com/google/guava/commit/3da34f983be8adfb2df893f6d3cd8faf95e2b3cb)
- [`MoreObjects.firstNonNull` was missing a nullness annotation, and the data did _not_ show that because it turned out that we'd never tested the case or even open-sourced the tests that we _did_ have.](https://github.com/google/guava/commit/4aaebca4e1f05891ec8e5ff0183cdde9894815f3)

I may try to incorporate some more annotations in the future, but that's the story so far.

dokson:
Investigated the current state — looks like this tracker may have been effectively superseded.

Since the original three commits and #6683, master shows ~60+ nullness-related commits, with the systematic driver having shifted from the paper's runtime-instrumentation data to Error Prone's [`RedundantNullCheck`](https://errorprone.info/bugpattern/RedundantNullCheck) static check (e.g. 125fd02, 680f8e1, 62ed153, plus google/error-prone#5121). In parallel:

- `@CheckForNull` → JSpecify `@Nullable` migration completed (d997ad9)
- `Void` → `@Nullable Void` sweep landed (77cf121), addressing the original thread on google/error-prone#3792
- Newly enabled nullness checking is being rolled out module-by-module (`base`, `escape`, `html`, `math`, `primitives`), which is itself surfacing and fixing missing annotations

[...]

PR #6525, opened 2023-06-01
Create NullableAnnot.java

Shahil093:
#6510

### DIFF
--- a/guava/src/com/google/common/annotations/NullableAnnot.java
+++ b/guava/src/com/google/common/annotations/NullableAnnot.java
@@ -1,0 +1,7 @@
+import javax.annotation.Nullable;
+
+public class ExampleClass {
+    public void exampleMethod(@Nullable String nullableParameter) {
+        // Method implementation
+    }
+}


jbduncan:
@Shahil093 Welcome to Guava, again!

I don't know if you've seen the [contribution guidelines](https://github.com/google/guava/blob/master/CONTRIBUTING.md) yet, but on a cursory look, it looks like this PR doesn't really add anything to Guava and be very unlikely to be accepted, for two reasons:

1. The story you linked to is about annotating _existing code_ with results from the mentioned paper, whereas your code just adds _new, example code_, so unless you're the author of the paper, this story probably wouldn't be a good one to pick up.
2. Guava very rarely accepts external contributions nowadays for reasons noted in the contribution guidelines, with small bug fixes and documentation fixes being two notable exceptions.

If you're looking you get into open source, then you may find this [this advice](https://github.com/junit-team/junit5/issues/528#issuecomment-250795455) and [this advice](https://github.com/junit-team/junit5/issues/528#issuecomment-250914304)  I wrote many years ago helpful.

I hope this helps!

chaoren:
Thanks for looking into this and responding, @jbduncan. Indeed, this PR seems to be coming from a misunderstanding of #6510.

Issue #6478, opened 2023-05-08
Confusion on contents of saved `Graph.adjacentNodes(n)` if `n` is eventually removed


jbduncan:
I recently found myself writing this code.

```java
MutableGraph<Integer> graph = GraphBuilder.undirected().build();
graph.addNode(1);

Set<Integer> adjacentNodesView = graph.adjacentNodes(1);
System.out.println(graph.edges()); // []
System.out.println(adjacentNodesView); // []
System.out.println(graph.adjacentNodes(1)); // []

graph.putEdge(1, 2);
System.out.println(graph.edges()); // [[2, 1]]
System.out.println(adjacentNodesView); // [2]
System.out.println(graph.adjacentNodes(1)); // [2]

graph.removeNode(1);
System.out.println(graph.edges()); // []
System.out.println(adjacentNodesView); // [2] <-- Unexpected! Should be []?
System.out.println(graph.adjacentNodes(1)); // throws IAE
```

I was surprised by the contents of `adjacentNodesView` at the end, because if `graph.edges()` is cleared, then I intuitively expected the adjacent nodes to be cleared too.

But then I thought about it some more and realised this is a weird grey area, because if a user calls `graph.adjacentNodes(1)` again, then an IAE is rightfully thrown to prevent bugs.

I'm torn between `adjacentNodesView` being `[]` or `[2]` in this case.

@jrtom, what do you think?

netdpb:
It seems to me that the set view of adjacent nodes of a node that was since removed from the graph is invalid, and its behavior could be undefined. It might be nice to invalidate that set-view, such that all calls throw `IllegalStateException`, but I don't know how urgent that might be.

jrtom:
That is an interesting one, for sure.  

Logically, the behavior is undefined (that is, we've never formally defined that behavior and I think that there are a couple of reasonable choices we could make, including both throwing ISE and returning the empty set), but we've never formally declared it to be undefined.  We might consider that solution, although it's not very satisfying.

As a practical matter, I am surprised that `adjacentNodesView` isn't empty; as a live view I would expect it to have been cleared.  This seems like it might indicate the presence of a bug that might have other manifestations.

Since `adjacentNodes()` is defined to return the set union of `predecessors()` and `successors()` (and is implemented via `Sets.union()` at present), it would take a bit of digging to figure out exactly where this is breaking down, as there are a few layers involved.

I don't have the bandwidth to look into this right now, so if you'd like to take a look we'd appreciate your insights.  :)

[...]

PR #6553, opened 2023-06-15
Make `Graph.adjacentNodes` more transparent when a node gets deleted.

ineuwirth:
When a node in the graph is deleted, it can sometimes cause confusion with the view provided by the `Graph.adjacentNodes` function. This happens because the view might have been created before the node was removed.

The removeNode function in `StandardMutableGraph` (which calls `StandardMutableValueGraph.removeNode`) meant to eliminate the connections related to the node. However, the `{Directed,Undirected}GraphConnections.adjacentNodeValues` associated with the to-be-removed node might still be referred to by the view constructed by `Graph.adjacentNodes` (which calls `StandardValueGraph.adjacentNodes`).

So, this can result in indirect references to the values of adjacent nodes (from the perspective of the node that was removed), even though they've been deleted. This makes the view less transparent, because it's showing information that isn't accurate anymore after the node has been removed.

The solution is to cleanup the associated `GraphConnections` before finally removing the node from the graph. Also it is now better aligned to the implementation of `StandardMutableNetwork.removeNode` reusing the same mechanism of copying the view to an immutable collection to avoid modifying the underlying view while iterating over it.


guava/src/com/google/common/graph/StandardMutableValueGraph.java
```diff
@@ -136,17 +137,23 @@ public boolean removeNode(N node) {
       }
     }
 
-    for (N successor : connections.successors()) {
+    // Since views are returned, we need to copy the successors that will be removed.
+    // Thus we avoid modifying the underlying view while iterating over it.
+    for (N successor : ImmutableSet.copyOf(connections.successors())) {
```
netdpb, line 142: Why `ImmutableSet` here and not `ImmutableList`? Seems unnecessary to deduplicate at this point.

ineuwirth: Well, `connection.successors()` returns with `Set<N>`, so I would not expect further deduplication at this point. I would be though concerned about the possible reordering what could happen through a hash-based set. (It might not cause issue during this cleanup, but it is not that trivial to prove.)

Anyway I borrowed this approach from `StandardMutableNetwork.removeNode` where elements are indeed copied to `ImmutableList`, probably I just unintentionally changed the type.

Thank you, I have fixed it.

netdpb:
@kevinb9n [suggested](https://github.com/google/guava/issues/6554#issuecomment-1599317581) that the right thing to do is in fact to throw, rather than to make the collection view empty in this case. Do you see a path towards doing that here instead?

ineuwirth:
It seems to me that the direction with throwing views probably requires a bigger rework in the `Graph` implementation in general.
So in that case we can just close this PR.

guava/src/com/google/common/graph/StandardMutableValueGraph.java
```diff
@@ -136,17 +137,23 @@ public boolean removeNode(N node) {
...
       // requireNonNull is safe because the node is a successor.
       requireNonNull(nodeConnections.getWithoutCaching(successor)).removePredecessor(node);
+      requireNonNull(connections.removeSuccessor(successor));
```
jrtom, line 145: It's been a while since I've looked at this code, but I'm a bit suspicious of the three different ways in which we're now accessing the connections of a node in this method:

* direct reference (or alias) for the internal data structure: `nodeConnections.get(node)` (via `connections`, line 127; `nodeConnections` is declared in the superclass as `MapIteratorCache<N, GraphConnections<N, V>> nodeConnections`)
* direct reference without caching: `nodeConnections.getWithoutCaching()` (line 144)
* (immutable) copy of the output of that structure (line 142)

We should (a) validate that we want those three different mechanisms, or use only the ones we actually want, and (b) document which one(s) we do want, and why. 

There's a lot of clever stuff going on in this code, which makes it easy to make mistakes in updating it.

(To be clear, I think you've explained why you want the copy of the set in the places that you've made it, but the reason why we're using each of the above two mechanisms in different places is less clear.)

jrtom:
Thank you very much for looking into this.  :)

This change looks fine (as far as it goes) if we assume that the correct behavior for a graph view that's been invalidated due to the removal of an argument to `Graph.adjacentNodes()` is to act as though the view is empty.  (And this is certainly better than the existing behavior of returning the original contents of the view, as reported in #6478.)

Since there's an ongoing discussion in #6554 of what the correct behavior _should_ be (which I'm about to weigh into), I think we should hold onto this change for the moment; if we decide that "act as though the view is empty" is correct, then we should merge it.

jrtom:
FYI, we are implementing your change internally; I goofed when attempting to do the merge (because the test code had changed) and it's easier to just do a clean version of it internally and then push that out than to fix your branch.

The internal change is tagged with you (@ineuwirth) as the original author so you will continue to get credit for it.  :)

Note that we're still considering (per the discussion in #6554) whether we ultimately want to throw in this case rather than returning an empty set, but in the meantime, returning an empty set is definitely better than returning a set whose membership is clearly incorrect.

Issue #2130, opened 2015-08-11
guava-tests build is failing on windows 


AnkushPuri:
Guava tests build is failing on windows. PFB the failure snippet from surefire-reports. 

guava-tests/target/surefire-reports/com.google.common.base.ThrowablesTest.txt:Tests run: 30, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.026 sec <<< FAILURE!
Binary file guava-tests/target/surefire-reports/com.google.common.io.ResourcesTest.txt matches
guava-tests/target/surefire-reports/com.google.common.reflect.ClassPathTest.txt:Tests run: 32, Failures: 5, Errors: 0, Skipped: 0, Time elapsed: 0.463 sec <<< FAILURE!

And here is the text for those two files

Test set: com.google.common.base.ThrowablesTest
-------------------------------------------------------------------------------
Tests run: 30, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.026 sec <<< FAILURE!
testGetStackTraceAsString(com.google.common.base.ThrowablesTest)  Time elapsed: 0.002 sec  <<< FAILURE!
junit.framework.AssertionFailedError: null
        at junit.framework.Assert.fail(Assert.java:47)
        at junit.framework.Assert.assertTrue(Assert.java:20)
        at junit.framework.Assert.assertTrue(Assert.java:27)
        at com.google.common.base.ThrowablesTest.testGetStackTraceAsString(ThrowablesTest.java:498)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:606)
        at junit.framework.TestCase.runTest(TestCase.java:168)
        at junit.framework.TestCase.runBare(TestCase.java:134)
        at junit.framework.TestResult$1.protect(TestResult.java:110)
        at junit.framework.TestResult.runProtected(TestResult.java:128)
        at junit.framework.TestResult.run(TestResult.java:113)
        at junit.framework.TestCase.run(TestCase.java:124)
        at junit.framework.TestSuite.runTest(TestSuite.java:243)
        at junit.framework.TestSuite.run(TestSuite.java:238)
[...]

PR #6560, opened 2023-06-17
#2130 - Stop skipping certain tests under Windows

ineuwirth:
Some tests depended on Unix line ending, changing those to use `System.lineSeparator()` instead. Now they can run on Windows too.

### DIFF
--- a/android/guava-tests/test/com/google/common/base/ThrowablesTest.java
+++ b/android/guava-tests/test/com/google/common/base/ThrowablesTest.java
@@ -658,9 +658,6 @@ public class ThrowablesTest extends TestCase {
   @J2ktIncompatible
   @GwtIncompatible // getStackTraceAsString(Throwable)
   public void testGetStackTraceAsString() {
-    if (isWindows()) {
-      return; // TODO: b/136041958 - We probably just need to accept \r\n line delimiters.
-    }
     class StackTraceException extends Exception {
       StackTraceException(String message) {
         super(message);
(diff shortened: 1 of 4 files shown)


ineuwirth:
I have noticed in the CI script that android tests are not executed under Windows but it still looked better to keep the sources as close as possible. I guess you would have done it internally anyway.

cpovirk:
Thanks!

> I have noticed in the CI script that android tests are not executed under Windows but it still looked better to keep the sources as close as possible. I guess you would have done it internally anyway.

Perhaps surprisingly, there are several factors here :) The short answer is that we _do_ want to keep them in sync but that it's simplest for you not to bother, as our automation all carry over the changes automatically (at least when they merge cleanly, as they do here).

(When we update both copies, then anyone who runs the Android tests under Windows (even though we don't do it under _our CI_) gets to run the tests. Plus, as you noted, keeping the sources in sync is itself an advantage.)

Anyway, there's no need to change anything here; I've already pulled this in. And like your other PR, this one will be re-created in another PR with the commit attributed to you.

(FYI, I'm backing out the new usage of `String.join` to maintain compatibility with ancient versions of Android. I could keep it in the mainline, backing it out only in the backport, but that would cut against the goal of keeping the two in sync.)

ineuwirth:
Thank you for sharing the behind the scenes details about how you handle the android part!

cpovirk:
> And like your other PR, this one will be re-created in another PR with the commit attributed to you.

I'm sorry, I got it right for the other PR but messed it up for this one. (I had some trouble with our automation—related, I theorize but haven't checked, to Windows line endings in the patch!—and ended up putting together the commits myself, only to forget to modify the author line in the one.)

ineuwirth:
No problem at all.

PR #6563, opened 2023-06-17
#2130 - Stop skipping certain tests under Windows

ineuwirth:
_(Edited: what I first thought was completely demolished by the CI results, see the other comments below)_

Certain tests were deactivated within AbstractFutureTest on Windows with the explanation that a subset of these test cases run slowly. This is certainly the case, but these tests also exhibit slow performance on Unix. However, they are marginally faster on Unix compared to Windows for some unknown reason. Overall, the difference in execution times doesn't appear to be significant enough to warrant excluding tests on any particular platform.

### DIFF
--- a/guava-tests/test/com/google/common/util/concurrent/AbstractFutureTest.java
+++ b/guava-tests/test/com/google/common/util/concurrent/AbstractFutureTest.java
@@ -218,9 +218,6 @@ public class AbstractFutureTest extends TestCase {
   }
 
   public void testToString_oom() throws Exception {
-    if (isWindows()) {
-      return; // TODO: b/136041958 - Some tests in this file are slow, but I'm not sure which.
-    }
     SettableFuture<Object> future = SettableFuture.create();
     future.set(
         new Object() {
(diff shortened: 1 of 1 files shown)


ineuwirth:
On the same machine (Intel® Core™ i5-10300H CPU @ 2.50GHz, 8 cores and 32GB RAM) I measured the followings:

image

Basically there are two tests (testFutureBash and testSetFutureCancelBash) that can be really slow, the third testToString_delayedTimeout has some intrinsic "sleeps" and the rest usually take way less than one second.

Still the Windows job (https://github.com/google/guava/actions/runs/5300189412/jobs/9593609626?pr=6563) had been running for ~1h 9m so clearly something is off. I am trying to exclude the two aforementioned known-to-be slow test cases to check whether the rest behaves well.

ineuwirth:
So keeping exclusion only for `testFutureBash` and `testSetFutureCancelBash` seems sufficient - `AbstractFutureTest` cases took 5.039 sec in the Windows job (https://github.com/google/guava/actions/runs/5300498860/jobs/9594079978?pr=6563) with these two excluded tests while it took 52.043 sec on Unix (https://github.com/google/guava/actions/runs/5300498860/jobs/9594079917?pr=6563) with all tests enabled.

Do we know something about the hardware running under the GitLab CI Windows jobs?

cpovirk:
Thank you for all the research! It's reassuring to hear that there's not a performance problem on a developer's Windows machine. I wouldn't be surprised if Windows GitHub CI uses some kind of emulation that normally works fine but has some performance edge cases when dealing with intentionally pathological code like what we have here.

It's also great to know that we can run all but the two tests under Windows: While I don't expect any of _those_ tests to start failing under Windows but not Linux, I'm happy that more of our tests will become runnable for contributors on Windows machines.

We'll get this merged. For technically reasons, it gets regenerated as a separate PR, but the commit will still be attributed to you.

cpovirk:
(_If_ for some reason we decided that we wanted _all_ the tests to run under Windows _normally_ but to be skipped only during CI, I'm sure we could wire that up by making the Windows CI set a system property and the tests look for it. But I prefer the simpler approach you've taken here. People can let us know if it causes them issues.)

PR #6583, opened 2023-06-24
extract repeated code segment into a new method in DoubleUtils

commandini:
Following code segment repeats 12 times in the code base:

`(cmpXToRoundArbitrarily <= 0) ? roundArbitrarily : Math.nextUp(roundArbitrarily);`

We can extract it into a new method in `DoubleUtils`.

If you think this refactoring is fine, please also suggest a better name for the new method.
 (current signature: ` double nextUp(int value, double nextUpValue) `)


chaoren:
> repeats 12 times in the code base

1. I wouldn't count repeated code under guava/ and android/guava/ separately, so it's really 6 times.
2. The  `roundToDouble` in `LongMath` looks like it's just a `long`-specific version of the generic `roundToDouble` in `ToDoubleRounder`. It even has this comment: `// Logic adapted from ToDoubleRounder.`, so it's really just used 3 times in `ToDoubleRounder`, but copied over.
3. If we were to extract `(cmpXToRoundArbitrarily <= 0) ? roundArbitrarily : Math.nextUp(roundArbitrarily)` out to a function, then we should probably do the same with `(cmpXToRoundArbitrarily >= 0) ? roundArbitrarily : DoubleUtils.nextDown(roundArbitrarily)`. That one happens equally many number of times. It's just split over multiple lines.
4. Given that it's really only used by `ToDoubleRounder`, I think it might be more appropriate as a package-private static method in `ToDoubleRounder`.

> (current signature: `double nextUp(int value, double nextUpValue)`)

5. it's only conditionally calling `Math.nextUp`, so calling it just `nextUp` is definitely not a good idea.
6. `value` and `nextUpValue` as parameter names also just loses all relevant context.

Honestly, I don't think this is worth refactoring. Maybe we can statically import `Math.nextUp` and `DoubleUtils.nextDown` and rename the variables to make the lines simpler, but refactoring three usages of a ternary just seems overkill.

Issue #2970, opened 2017-10-18
Add module-info.java


hrchu:
So that projects depend on this can be published to a public artifact repository.
Note that this is not breaking backward compatibility. All codes except this file can be still compiled in Java 6.

jbduncan:
Guava has an `Automatic-Module-Name` in its `MANIFEST.MF` now, so I believe this is not quite as important as it may seem. But if I'm mistaken, then I'd be more than happy to be proven so.

(BTW, I think I might have misunderstood something, because `module-info.java` is Java 9 specific, right? Would a Java 8 compiler (which I believe is what is used to compile guava-jre) or an Android compiler (for guava-android) happily process it or otherwise ignore it?)

cpovirk:
Our current thinking is that we'll look into this next quarter. We have seen some problems from `module-info` files in third-party jars, since they're Java 9 `.class` files and not everyone uses Java 9 yet. So, if we can get away with `Automatic-Module-Name`, we'll continue to do so, possibly even beyond next quarter. But if there are cases in which `Automatic-Module-Name` isn't good enough, that would be good to know.

hrchu:
It is possible to only compile ```modile-info.java```, so the jar file is still compatible to users who uses earlier Java version. 

There is a maven example here:
https://github.com/twonote/radosgw-admin4j/blob/java9/pom.xml#L127

cpovirk:
Right, thanks. We've seen problems even when the main `.class` files are compiled for an older version but the `module-info.class` is present. As I understand it, various tools try to process all `.class` files, and they need to be updated to ignore `module-info.class`.

hrchu:
@cpovirk could you tell me more about this problem?

cpovirk:
I wasn't personally involved in fixing the problems, but the basic idea seems to be that people scan the whole classpath (using something like [`ClassPath.getTopLevelClasses`](http://google.github.io/guava/releases/snapshot-jre/api/docs/com/google/common/reflect/ClassPath.html#getTopLevelClasses--) -- which might be an example of something that needs updated to ignore `module-info.class` :)) and then try to examine/load the classes with a tool that understands only, say, Java 8.

orionll:
It's worth mentioning that if we add `module-info.java`, all packages will not be open anymore.

jbduncan:
@orionll Am I right to think/remember that in the JPMS, open packages are packages whose internal classes can be inspected with reflection?

orionll:
[...]

PR #6606, opened 2023-06-29
Remove annotation-only Jars from runtime classpath (in Gradle)

jjohannes:
Proposal to remove the following Jars from the **runtime classpath** (in Gradle projects):
- `checker-qual-*.jar`
- `error_prone_annotations-*.jar`
- `jsr305-*.jar`

They are still transitively visible on the **compile classpath**, which can not be expressed in POMs, but in Gradle Metadata which is published starting with the next Guava release (33). In a Gradle build setup, this would be "compileOnlyApi" scope. See also https://github.com/google/guava/issues/2824#issuecomment-358069931 and the linked Maven feature request for the discussion about having such a "compileOnlyApi scope" in Maven.

Context:
- Original issue on this topic: #2824
- It was considered to do this as part of #3683 but not done for more [as this might be a breaking change for some users](https://github.com/google/guava/pull/3683/commits/c5e364f4775b2ac50d8c6565a203a261387124a4).
- I think for most (almost all?) users these Jars are not needed at runtime **because of Guava**.  If they are needed for other reasons, they should be declared directly by the users in any case. For rare cases where only the use of Guava requires these to be added (if someone has examples for such cases, please share) the user can still add them them as `runtimeOnly`. You could consider them as an "optional feature" that you have to add yourself if needed.


OrangeDog:
> which can not be expressed in POMs

`<scope>provided</scope>`?

cpovirk:
Sadly, `<scope>provided</scope>` [also removes the annotations at compile time](https://issues.apache.org/jira/browse/MNG-2205?focusedCommentId=16267311&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16267311), and that can lead to problems.

ben-manes:
In the past compiler bugs like [JDK-8152174](https://bugs.openjdk.org/browse/JDK-8152174), https://github.com/scala/bug/issues/7751, and [earlier](https://www.scala-lang.org/old/node/10575) meant putting annotations on the classpath was the safest option. Would this change allow for encountering those types of bugs again? (No preference, just backstory)

jjohannes:
That's what this is about: not running into problems like https://github.com/google/guava/pull/6606#issuecomment-1614267029 again but still solving #2824.

- You want the annotation libraries at **compile time** to avoid compilation issues as the ones @ben-manes listed above.
- You do not want them at runtime (i.e. package them with your application) - that's what #2824 is about.

For Gradle, we can solve this now: have dependencies at compile time, but not at runtime. That's what this PR does for the annotation libraries.

cpovirk:
https://github.com/google/guava/wiki/UseGuavaInYourBuild#what-about-guavas-own-dependencies mentions at least 2 runtime problems, one of which is still probably a concern for one of our deps:

- [JDK-8152174](https://bugs.openjdk.org/browse/JDK-8152174), fixed in JDK9, but I'd assume that we have enough JDK8 users for someone to be affected.
- [JDK-8247797](https://bugs.openjdk.org/browse/JDK-8247797), still present but maybe not actually a problem in all the situations it claims, including in Guava?

JDK-8152174 is potential trouble, albeit only for `checker-qual`. (That said, `checker-qual` is the largest dependency, so it's probably the one that users would most like to see us cut. The plan is still to move to `jspecify` once it's ready. That would help with the size.) Perhaps our next step should be to hide deps _other_ than `checker-qual`.

JDK-8247797 is probably fine, which is interesting: Despite what its writeup suggests, I can successfully reflect on `ParametricNullness`, which is annotated with `javax.annotation.Nonnull`, including using an enum element with value `javax.annotation.meta.When.UNKNOWN`, even when jsr305 is not on the classpath:

```
$ cat GuavaReflectionNoDeps.java
import com.google.common.base.Function;
import java.lang.annotation.Annotation;
import java.lang.reflect.Method;

class GuavaReflectionNoDeps {
    for (Method method : Function.class.getMethods()) {
      for (Annotation onFunction : method.getAnnotations()) {
        System.out.println(onFunction);
        for (Annotation onAnnotation : onFunction.annotationType().getAnnotations()) {
          System.out.println("  " + onAnnotation);
        }
      }
    }
  }
}

$ ( CP=$HOME/.m2/repository/com/google/guava/guava/32.1.0-jre/guava-32.1.0-jre.jar:$HOME/.m2/repository/com/google/guava/failureaccess/1.0.1/failureaccess-1.0.1.jar; ~/openjdk-8u342/bin/javac -cp $CP GuavaReflectionNoDeps.java && ~/openjdk-8u342/bin/java -cp .:$CP GuavaReflectionNoDeps )
@com.google.common.base.ParametricNullness()
  @java.lang.annotation.Retention(value=RUNTIME)
  @java.lang.annotation.Target(value=[FIELD, METHOD, PARAMETER])

```

It looks like the problem _actually_ surfaces only when `Nonnull` _is_ on the classpath and `When` is _not_. That of course won't happen here, since both are part of the same jsr305 artifact: Either both will be present, or neither will:

```
$ cp ~/.m2/repository/com/google/code/findbugs/jsr305/3.0.2/jsr305-3.0.2.jar /tmp

$ zip -d /tmp/jsr305-3.0.2.jar javax/annotation/meta/When.class
deleting: javax/annotation/meta/When.class

$ ( CP=$HOME/.m2/repository/com/google/guava/guava/32.1.0-jre/guava-32.1.0-jre.jar:$HOME/.m2/repository/com/google/guava/failureaccess/1.0.1/failureaccess-1.0.1.jar; ~/openjdk-8u342/bin/javac -cp $CP GuavaReflectionNoDeps.java && ~/openjdk-8u342/bin/java -cp .:$CP:/tmp/jsr305-3.0.2.jar GuavaReflectionNoDeps )
@com.google.common.base.ParametricNullness()
Exception in thread "main" java.lang.NoClassDefFoundError: javax/annotation/meta/When
        at java.lang.Class.getDeclaredMethods0(Native Method)
        at java.lang.Class.privateGetDeclaredMethods(Class.java:2701)
        at java.lang.Class.getDeclaredMethods(Class.java:1975)
        at sun.reflect.annotation.AnnotationType$1.run(AnnotationType.java:112)
        at sun.reflect.annotation.AnnotationType$1.run(AnnotationType.java:109)
        at java.security.AccessController.doPrivileged(Native Method)
        at sun.reflect.annotation.AnnotationType.<init>(AnnotationType.java:109)
        at sun.reflect.annotation.AnnotationType.getInstance(AnnotationType.java:85)
        at sun.reflect.annotation.AnnotationParser.parseAnnotation2(AnnotationParser.java:266)
        at sun.reflect.annotation.AnnotationParser.parseAnnotations2(AnnotationParser.java:120)
        at sun.reflect.annotation.AnnotationParser.parseAnnotations(AnnotationParser.java:72)
        at java.lang.Class.createAnnotationData(Class.java:3521)
        at java.lang.Class.annotationData(Class.java:3510)
        at java.lang.Class.getAnnotations(Class.java:3446)
        at GuavaReflectionNoDeps.main(GuavaReflectionNoDeps.java:10)
Caused by: java.lang.ClassNotFoundException: javax.annotation.meta.When
        at java.net.URLClassLoader.findClass(URLClassLoader.java:387)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:418)
        at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:352)
        at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
        ... 15 more
```

jjohannes:
Thanks for @cpovirk for clarifying that some issues are also about runtime. My understanding though is that you have to be in a very specific situation to get into such an issue. You need to use reflection on a class that uses these annotations. Which you could do an a Guava class. (But is that an intended use of Guava?)
And IIUC, it is only a bug in JDK8 that this may cause an problematic error, while in other cases there is no "real" problem.

I guess my take on this is, that if users do some of these things, it is most likely that they should directly declare dependencies on the annotation libraries they want to deal with and not rely on Guava brining them in transitively. I don't know "real" projects doing something like this, though. So my assumption here might be wrong.

I don't know how big the part of the user base is that still uses JDK8 **and** does something so specific that they run into the bug. It feels like it could be ok to document that you need to add a certain `<runtime>` / `runtimeOnly` dependency yourself (or move to a newer Java) if you run into it. (But maybe I am not conservative enough :))

We _could_ also drive the variant setup further and have a separate dependency setup for JDK8 (that keeps one or all dependencies) and do the clean setup for everyone using JDK9+.

cpovirk:
There seem to be frameworks out there that try to automatically (e.g.) serialize arbitrary objects by looking at their fields. That is of course not a great idea, and it's one of the reasons that the Java module system exists (and perhaps a reason for [us to release a jar with a `module-info` someday](https://github.com/google/guava/issues/2970)). Maybe it's not a bad idea for us to "gently" break such users now in preparation for breaking them more firmly in the future.... And the results could be educational for us in general.

I would probably still start with omitting the other deps first, but I'm warming a little to trying this out for checker-qual, too.

We'll see if others have something to add. Given the flood of releases lately, we'll probably wait a little before publishing another one, so we have time to figure this out.

vlsi:
>  You need to use reflection on a class that uses these annotations. Which you could do an a Guava class. (But is that an intended use of Guava?)

Well, type annotations add information to the type.
Do you think every reflection should be disabled unless explicitly permitted by the library authors?
If the reflection for public methods should be available, it should probably provide type information like "acceptable parameter type" and "result type", and it is important to know if the type is nullable or not.

For instance, a JSON parser might user object constructor when parsing JSON, and it might inspect if the parameters are `@Nullable`. If the parameter is non-nullable, and JSON misses the value, then the parser could produce a better error message than a mere "parameter X is null".

OrangeDog:
@vlsi in such cases the tool depends on those annotation types itself. Or in the case of "nullable" just looks at the name and none of the details of the annotation.

vlsi:
> Or in the case of "nullable" just looks at the name and none of the details of the annotation

If the annotation class is not provided in the runtime, the reflection API does not return the annotation, so the tool can't even detect annotation was present in the first place.
That is like the "parameter type varies depending on the runtime classpath" issue. It is not the best issue to debug.

Of course, there's a backdoor like `getResourceAsStream("ClassName.class")` and then parsing the bytecode, however, it would sound strange. Of course, Java specification says Java must ignore annotations that are not present on the classpath, yet bugs like https://github.com/google/guava/pull/6606#issuecomment-1614267029 exist :-(

In other words, I believe it is fair to expect that "types used in the public signatures" (e.g. return types of the public APIs) should be present in "direct runtime dependencies".

At the same time, the annotations form error-prone do not sound like "type annotations", and they are much more like "compile-time-only" annotations, so having them as `compileOnlyApi` (==transitive `compile` dependencies) rather than `compile` (==transitive `compile` + `runtime`) probably makes sense.

cpovirk:
(breadcrumb: If we do this, we'll want to watch out for GWT issues like https://github.com/google/guava/issues/7134. That's not a big concern, and I remain more nervous about omitting dependencies for other reasons than for GWT reasons.)

jjohannes:
Wit the switch to jspecify, this is fairly outdated I believe. Closing.

chadlwilson:
The PR itself is outdated so makes sense to close, especially without an apparent path forward, but the core irritation still exists for those looking to have clean/simple sets of runtime dependencies.

For the record, the list to exclude on Guava `33.4.8` via Gradle for users not exposed to some of the quirks above is something like

```groovy
  implementation(group: 'com.google.guava', name: 'guava', version: '33.4.8-jre') {
    // workaround for Guava metadata declaring dependencies that are not needed at runtime
    // see https://github.com/google/guava/pull/6606
    exclude(module: 'listenablefuture')
    exclude(module: 'jspecify')
    exclude(module: 'error_prone_annotations')
    exclude(module: 'j2objc-annotations')
  }
```

or in a different style where multiple dependencies and/or transitives use Guava:

```groovy
  components {
    // workaround for Guava metadata declaring dependencies that are not needed at runtime
    // see https://github.com/google/guava/pull/6606
    withModule('com.google.guava:guava', { details ->
      details.allVariants {
        withDependencies {
          removeAll {
            it.name in [ "listenablefuture", "jspecify", "error_prone_annotations", "j2objc-annotations" ]
          }
        }
      }
    })
  }
```

PR #6662, opened 2023-08-01
Supplement toByteArray docs with respect to byte order

findepi:


perceptron8:
1. Shouldn't you use `ByteOrder.BIG_ENDIAN` instead of `ByteOrder.LITTLE_ENDIAN`?
2. It's `ByteOrder.BIG_ENDIAN` by default. Adding explicit order makes code less readable, at least for me.

eamonnmcmanus:
The Guava team has discussed this and we agree with @perceptron8. If we were going to make a change like this, it would have to say `BIG_ENDIAN` rather than `LITTLE_ENDIAN`. But anyway the text already says "big-endian" immediately before. If someone doesn't know that `ByteBuffer.allocate` returns a big-endian buffer, they probably don't know enough about the `ByteBuffer` API for the cited equivalence to be useful to them, whether or not we add a redundant call to `.order(BIG_ENDIAN)`.

Issue #6634, opened 2023-07-12
From Guava 32.0.1-jre, Files.createTempDir() can fail or create un-deletable directories when used from a Windows service


fmarot:
Hello I rencently upgraded from 31.1-jre to 32.0.1-jre.
FYI 32.0.0-jre was KO because of the then-corrected-in-32.0.1-jre error: java.lang.UnsupportedOperationException: 'posix:permissions' not supported as initial attribute. So let's not talk about this version.

In 32.0.1-jre the method Files.createTempDir() has been modified and a program RUNNING ON WINDOWS is now unable to delete its own folders created by this call.

Here is a sample creating a folder with java's own "createTempDirectory" and deleting it (this works fine) and then the same with Guava and it cannot delete it's own folder created with guava.
```

        @Test
	public void testApp() throws IOException {
		Path tempDir = java.nio.file.Files.createTempDirectory("temp");
		displayDirectoryPermissions(tempDir);
		boolean isDeleted = tempDir.toFile().delete();
		assertTrue(isDeleted);

		System.out.println(" - - - - - - - - - ");

		Path tempDir2 = Files.createTempDir().toPath();
		displayDirectoryPermissions(tempDir2);
		boolean isDeleted2 = tempDir2.toFile().delete();
		assertTrue(isDeleted2);
	}

	private static void displayDirectoryPermissions(Path directory) throws IOException {
        System.out.println("Informations sur les droits d'accès du répertoire : " + directory.toAbsolutePath());

		AclFileAttributeView view = java.nio.file.Files.getFileAttributeView(directory, AclFileAttributeView.class);
		System.out.println(view.toString());

		view.getAcl().forEach(it -> System.out.println(it.toString()));
    }
```
Here are the (desidentified) trace and there is a huge difference beween both calls:

```
Informations sur les droits d'accès du répertoire : e:\xxxxxxx\target\temp18063228888024228555
sun.nio.fs.WindowsAclFileAttributeView@7494f96a
BUILTIN\Administrators:READ_DATA/WRITE_DATA/APPEND_DATA/READ_NAMED_ATTRS/WRITE_NAMED_ATTRS/EXECUTE/DELETE_CHILD/READ_ATTRIBUTES/WRITE_ATTRIBUTES/DELETE/READ_ACL/WRITE_ACL/WRITE_OWNER/SYNCHRONIZE:ALLOW
[...]

Issue #2686, opened 2016-12-15
Consider running CI under Windows


jbduncan:
As is currently done in [google-java-format](https://github.com/google/google-java-format), I think it's worth testing Guava against Windows using AppVeyor, or at least I think it's worth for the Guava team to consider looking into it.

Currently, Guava makes no guarantees that it will run successfully on Windows, as explained in the README:

> **IMPORTANT WARNINGS**
> ...
> 5, We unit-test and benchmark the libraries using only OpenJDK 1.7 on Linux. Some features, especially in `com.google.common.io`, may not work correctly in other environments.

I do understand that there would need to be a number of changes or improvements before Guava can build successfully on Windows (https://github.com/google/guava/issues/2130), so I suspect this would be a long-term thing to tackle.

jbduncan:
@cpovirk Oh no! Yeah, the potential incompatibility between the new implementations and Windows occurred to me earlier today. 😆

cpovirk:
This probably did get a lot easier with the switch of our CI to GitHub Actions. It might be roughly just this...

https://github.com/google/error-prone/blob/3602ab040d4dc0f7348fe6103ccfd20252ec1af9/.github/workflows/ci.yml#L27-L29

..._and_ then some number of `if (isWindows()) { return; }` or `assumeThat(isWindows()).isFalse()` checks in our tests... :) But #2130 suggests that even that might not be too bad....

I've been trying to remember if the tests would have caught any production problems under Windows (other than, you know, the significant recent one...). I think we knew about `Files.deleteRecursively` (#365) before it was reported (and I think a similar problem exists even for Linux?), and I'm not sure if #3402 was Windows-specific, either—though perhaps the default Windows temporary directory would have triggered the bug, and that would have been useful.

cpovirk:
Now playing with https://github.com/google/guava/pull/6542...

Our "print surefire reports" step seems likely to go poorly, given that it runs a shell script, but maybe I'll be surprised?

cpovirk:
Well, we might need to skip another test, too:

Linux:
```
Running com.google.common.util.concurrent.AbstractFutureFallbackAtomicHelperTest
Tests run: 43, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 102.883 sec
```

Windows:
```
Running com.google.common.util.concurrent.AbstractFutureFallbackAtomicHelperTest
[...]

PR #6731, opened 2023-09-26
clone of https://github.com/google/guava/pull/6730 but with Windows testing on JDK8

cpovirk:
Fix `Files.createTempDir` and `FileBackedOutputStream` under Windows _services_, a rare use case.

Fixes https://github.com/google/guava/issues/6634

Relevant to https://github.com/google/guava/issues/2686 in that it shows that we would ideally run our Windows testing under both Java 8 *and* a newer version....

RELNOTES=`io`: Fixed `Files.createTempDir` and `FileBackedOutputStream` under [Windows _services_, a rare use case](https://github.com/google/guava/issues/6634).
PiperOrigin-RevId: 568604081

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,11 +20,11 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest ]
-        java: [ 8, 11, 17 ]
-        root-pom: [ 'pom.xml', 'android/pom.xml' ]
+        java: [ 8 ]
+        root-pom: [ 'pom.xml' ]
         include:
           - os: windows-latest
-            java: 17
+            java: 8
             root-pom: pom.xml
     runs-on: ${{ matrix.os }}
     env:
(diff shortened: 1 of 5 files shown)


cpovirk:
The failure is an unrelated problem, so I think this is behaving as hoped on Windows under Java 8.

That failure:

```
  testServiceStartupTimes(com.google.common.util.concurrent.ServiceManagerTest): expected to be at least: 150
```

During the course of this testing, I've also seen a failure in a near-identical test:

```
  testServiceStartupDurations(com.google.common.util.concurrent.ServiceManagerTest): expected to be at least: PT0.15S
```

Here's the full version of the first failure:

```
testServiceStartupTimes(com.google.common.util.concurrent.ServiceManagerTest)  Time elapsed: 0.362 sec  <<< FAILURE!
expected to be at least: 150
but was                : 146
	at com.google.common.util.concurrent.ServiceManagerTest.testServiceStartupTimes(ServiceManagerTest.java:131)
```

I suspect that I've seen it only during these Windows Java 8 tests, so I may set up the test to no longer run in that environment (not that we run _CI_ for Windows Java 8, anyway). (As with [the other Windows slowness we've seen](https://github.com/google/guava/issues/2130), I wouldn't be surprised if the problem were less "Windows" and more "emulation.")

PR #6807, opened 2023-10-25
Add Maven wrapper distribution checksum

### LINKED ISSUE #6805: Add checksums for Maven wrapper
It is possible to specify the expected checksums for the Maven wrapper and the Maven distribution using `wrapperSha256Sum` and `distributionSha256Sum` in the `maven-wrapper.properties` file, see https://maven.apache.org/wrapper/index.html#checksum-verification-of-downloaded-binaries.

Do you think it would make sense to specify these checksums for Guava? This could cause issues though if the user does not have the needed packages installed locally, see for example:
https://github.com/google/guava/blob/050f89f39f3a0e3276cd5696a82e0bb050915b52/mvnw#L271-L272

(As side note, if the files had already been downloaded before, their checksums are apparently not checked anymore, see [MWRAPPER-93](https://issues.apache.org/jira/browse/MWRAPPER-93))

Marcono1234:
Resolves #6805

Currently the checksums are not published ([MWRAPPER-117](https://issues.apache.org/jira/browse/MWRAPPER-117)), so if you want to verify that they are correct you have to calculate them yourself, or locally edit them to something incorrect and see that `mvnw` execution fails (assuming you have deleted previous downloads in `.m2/wrapper/dists`, see [MWRAPPER-93](https://issues.apache.org/jira/browse/MWRAPPER-93)).

### DIFF
--- a/.mvn/wrapper/maven-wrapper.properties
+++ b/.mvn/wrapper/maven-wrapper.properties
@@ -15,4 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.9.5/apache-maven-3.9.5-bin.zip
+distributionSha256Sum=7822eb593d29558d8edf87845a2c47e36e2a89d17a84cd2390824633214ed423
 wrapperUrl=https://repo.maven.apache.org/maven2/org/apache/maven/wrapper/maven-wrapper/3.2.0/maven-wrapper-3.2.0.jar
+# TODO: Does not work properly on Windows, see https://issues.apache.org/jira/browse/MWRAPPER-103
+# wrapperSha256Sum=e63a53cfb9c4d291ebe3c2b0edacb7622bbc480326beaa5a0456e412f52f066a


.mvn/wrapper/maven-wrapper.properties
```diff
@@ -15,4 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 distributionUrl=https://repo.maven.apache.org/maven2/org/apache/maven/apache-maven/3.9.5/apache-maven-3.9.5-bin.zip
+distributionSha256Sum=7822eb593d29558d8edf87845a2c47e36e2a89d17a84cd2390824633214ed423
```
Marcono1234, line 18: This is apparently not affected by MWRAPPER-103 because it validated in Java code (probably [here](https://github.com/apache/maven-wrapper/blob/maven-wrapper-3.2.0/maven-wrapper/src/main/java/org/apache/maven/wrapper/Installer.java#L94-L98))

cpovirk: Oh, neat, so maybe there's no need for `sha256sum` to be present on the machine, either?

cpovirk: Oh, oops, we would need it if we were verifying the wrapper. (I'd gotten myself mixed up again between `script` and `only-script`.) So it's true that we don't need it now, but we'd like to verify the wrapper someday, at which point we will need it.

PR #6839, opened 2023-11-16
Improve GitHub issue templates

Marcono1234:
Tries to improve the GitHub issue templates, as originally suggested in the comments on #6785.

Please let me know if you want anything changed. Also, no worries if you don't want to implement these changes, I just wasn't sure whether anyone had seen my suggestions on #6785.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -39,6 +39,13 @@ body:
     validations:
       required: true
 
+  - type: input
+    attributes:
+      label: Guava Version
+      description: Guava version for which you encountered the bug, for example `32.1.3-jre`.
+    validations:
+      required: true
+
   - type: dropdown
     attributes:
       label: Packages
@@ -72,6 +79,7 @@ body:
         - Java 8
         - Java 11
         - Java 17
+        - Java 21
 
   - type: checkboxes
     attributes:
--- a/.github/ISSUE_TEMPLATE/feature_addition_request.yaml
+++ b/.github/ISSUE_TEMPLATE/feature_addition_request.yaml
@@ -9,7 +9,7 @@ body:
 
 
         Be aware, though: most feature requests are not accepted, even if they're suggested by
-        a full-time Guava team member. [Feedback](http://stackoverflow.com/a/4543114/869736) from
+        a full-time Guava team member. [Feedback](https://stackoverflow.com/a/4543114) from
         our users indicates that they really appreciate Guava's high power-to-weight ratio. It's
         important to us to keep Guava as easy to use and understand as we can. That means boiling
         features down to compact but powerful abstractions, and controlling feature bloat carefully.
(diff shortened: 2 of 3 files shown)


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -39,6 +39,13 @@ body:
     validations:
       required: true
 
+  - type: input
+    attributes:
+      label: Guava Version
```
Marcono1234, line 44: I put this at this position to be close to the "Packages" selection. But maybe it would be better to move it further up (maybe even as first field)?

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -39,6 +39,13 @@ body:
...
+      description: Guava version for which you encountered the bug, for example `32.1.3-jre`.
```
Marcono1234, line 45: The `32.1.3-jre` is currently displayed a bit small by the GitHub UI:
GitHub issue forms screenshot

Is that ok nonetheless, or should it for example be written as '32.1.3-jre'?

.github/ISSUE_TEMPLATE/feature_addition_request.yaml
```diff
@@ -9,7 +9,7 @@ body:
 
 
         Be aware, though: most feature requests are not accepted, even if they're suggested by
-        a full-time Guava team member. [Feedback](http://stackoverflow.com/a/4543114/869736) from
+        a full-time Guava team member. [Feedback](https://stackoverflow.com/a/4543114) from
```
chaoren, line 12: I copied all of these directly from the [wiki](https://github.com/google/guava/wiki/HowToContribute), but yeah all of these should probably be https if possible. I'll update all of the links in the wiki as well.

.github/ISSUE_TEMPLATE/bug_report.yaml
chaoren: Maybe we can also add a (required?) checkbox to say something like

> - [ ] This is currently [the latest version of Guava available](https://github.com/google/guava/releases/latest).

chaoren (commented):
Thanks for the suggestions. I think we do want all of these changes.

.github/ISSUE_TEMPLATE/bug_report.yaml
Marcono1234: Should this be part of the "Checklist" below then? Or / and should the description of the version field also mention this?

.github/ISSUE_TEMPLATE/bug_report.yaml
chaoren: We can move that into the `placeholder` of the `input`.

.github/ISSUE_TEMPLATE/bug_report.yaml
chaoren: The checkbox will show up in the issue once it's filed, so it might be a little weird to put it at the top. I think we can add some instructions at the top asking them to first make sure the bug is still present in the latest version. The checkbox can be at the bottom.

chaoren (commented):
I'm gonna also update the wording in the instructions a bit. We can just import PRs directly anyway, so feel free to just leave this open. I'll resolve it with an internal change.

PR #6860, opened 2023-12-05
Update AbstractNetworkTest.java

ctarnold:
Fix typo in comment to improve readability.


jrtom:
Thanks for spotting that.  As it happens, I'm editing AbstractNetworkTest.java for another fix, so I've incorporated your fix into that change.

PR #6923, opened 2024-01-16
Make BloomFilter.bitSize() public

Siddharth1410:

### DIFF
--- a/android/guava/src/com/google/common/hash/BloomFilter.java
+++ b/android/guava/src/com/google/common/hash/BloomFilter.java
@@ -214,7 +214,7 @@ public final class BloomFilter<T extends @Nullable Object> implements Predicate<
 
   /** Returns the number of bits in the underlying bit array. */
   @VisibleForTesting
-  long bitSize() {
+  public long bitSize() {
     return bits.bitSize();
   }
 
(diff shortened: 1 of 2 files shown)


android/guava/src/com/google/common/hash/BloomFilter.java
```diff
@@ -214,7 +214,7 @@ public long approximateElementCount() {
 
   /** Returns the number of bits in the underlying bit array. */
   @VisibleForTesting
```
rohit-satya, line 216: now that we've made it public should we remove the annotation? 

The documentation states the following

> Do not use this interface for public or protected declarations: it is a fig leaf for bad design, and it does not prevent anyone from using the declaration---and experience has shown that they will. If the method breaks the encapsulation of its class, then its internal representation will be hard to change. Instead, use RestrictedApiChecker , which enforces fine-grained visibility policies.

guava/src/com/google/common/hash/BloomFilter.java
```diff
@@ -215,7 +215,7 @@ public long approximateElementCount() {
 
   /** Returns the number of bits in the underlying bit array. */
   @VisibleForTesting
```
rohit-satya, line 217: same here

SAIVARDHAN15:
hey, anyone working on #6923? , 
if not, I am glad to work on it

PR #6982, opened 2024-02-11
added class PrimitiveMath, that focus on performing Addition Operatio…

harishwar-babu:
Created a new class called PrimitiveMath , which has two static functionalities
* Addition(v1,v2) --> (for integer, long,short) --> returns integer value.
* isEven(v1) ---> (for integer,long,short) ---> returns boolean value.

### DIFF
--- a/guava/src/com/google/common/math/PrimitiveMath.java
+++ b/guava/src/com/google/common/math/PrimitiveMath.java
@@ -28,7 +28,7 @@ public final class PrimitiveMath{
      * The input takes 2 integer values.
      */
     public static int add(int value1,int value2){
-        return (value1^v2)+2*(value1&value2);
+        return (value1^value2)+2*(value1&value2);
     }
 
     /*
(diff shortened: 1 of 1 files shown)


eamonnmcmanus:
Please see the [How to Contribute](https://github.com/google/guava/blob/master/CONTRIBUTING.md) document. We generally don't accept pull requests that add a new API without prior discussion. I note too that the code in this PR doesn't build.

harishwar-babu:
ok, shall I discuss the new API that I have designed with you?

Thanks
Harish

harishwar-babu:
ok, shall I discuss on the new API that I have designed with you?

eamonnmcmanus:
The document I linked to explains the process. A proposed new API starts with a new issue describing what the API looks like and what use cases it addresses.

I will say, to save everyone time, that I think it is unlikely that an API like the proposed `add` will be accepted. The claim that a method adding integers by using bitwise operations is faster than the native `+` operator is frankly astonishing. I think you would need to produce some very convincing evidence in the form of benchmarks.

There is maybe _slightly_ more justification for `isEven`, but I think most people know or can find out that they can write `(i & 1) == 0` or `i % 2 == 0`.

harishwar-babu:
Thanks for your Tip sure, will check out The bench mark. And I try to re-
commit the code to avoid build failure issue.
Thanks
Harish

harishwar-babu:
So, shall I attach the file about the bench marks and justification in this thread. for your reference.

Issue #6901, opened 2024-01-03
There is no charset parameter on application/json


reschke:
### Description

Ref: https://datatracker.ietf.org/doc/html/rfc8259#section-11 (last paragraph)

But see:

https://github.com/google/guava/blob/f347fb7a2ddd975e7ab94e0f3f71ab3f0d21f580/guava/src/com/google/common/net/MediaType.java#L492

This type should be deprecated (and users encouraged to use a variant without charset).

### Example

```java
MediaType.JSON_UTF_8 is in non-sense media type; no charset param exists for JSON.
```

### Expected Behavior

Use should lead to a deprecation warning.

### Actual Behavior

It does not.

### Packages

com.google.common.net

### Checklist

- [X] I agree to follow the [code of conduct](https://github.com/google/.github/blob/master/CODE_OF_CONDUCT.md).

cpovirk:
Interesting, thanks. I don't know that this has come up before.

We actually had `application/json` _without_ a charset before replacing it with the current constant back in 2012 (before `MediaType` was added to Guava). That presumably was the right move then (since it predates the 2017 RFC you've shared).

It is interesting that the RFC also says "Adding \[a charset\] really has no effect on compliant recipients," which suggests that including one should be harmless for compliant recipients.

But wait, https://www.rfc-editor.org/errata/eid5853 says that that sentence should be replaced :\ I'd have to read more to understand whether including a `charset` parameter should in fact technically be harmless.

There's additionally the question of whether the `charset` parameter makes things better or worse for _non-compliant_ recipients. (And then there's the question of whether helping non-compliant recipients is a good thing or a bad thing... :))

Our internal security guidance says that it is "critical" to include the `charset` parameter. That said, the guidance dates from at least 7 years ago, and I don't know how recently it's been reevaluated. Some chain of other links led me to https://portswigger.net/research/json-hijacking-for-the-modern-web, which was from 2016 (with some kind of update in 2022), which likewise suggests that the `charset` is important (or at least was back then). However, I haven't read it nearly closely enough to have much confidence in anything.

Someone seems to be [reporting](https://stackoverflow.com/a/56846529) that Dart needed the parameter back in 2019. Ditto [some "HttpClient" in 2020](https://stackoverflow.com/a/60662152).

And I've seen another report or two that some receivers _reject_ anything that includes `charset` ([example](https://github.com/hspec/hspec-wai/issues/46))....

[...]

PR #7027, opened 2024-02-21
RFC 9512: YAML Media Type

simon04:
https://www.rfc-editor.org/rfc/rfc9512.html

### DIFF
--- a/guava/src/com/google/common/net/MediaType.java
+++ b/guava/src/com/google/common/net/MediaType.java
@@ -492,6 +492,11 @@ public final class MediaType {
   public static final MediaType JSON_UTF_8 = createConstantUtf8(APPLICATION_TYPE, "json");
 
   /**
+   * For <a href="https://tools.ietf.org/html/rfc9512.">YAML Media Type</a>.
+   */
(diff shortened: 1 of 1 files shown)


kalebers (approved):

cpovirk:
Thanks. Do you know where I might learn more about the `charset` parameter? In https://www.iana.org/assignments/media-types/application/yaml, I see "Optional parameters: N/A; unrecognized parameters should be ignored." That's not as strong as https://www.iana.org/assignments/media-types/application/json, which has a Note that "No 'charset' parameter is defined for this registration." (But see https://github.com/google/guava/issues/6901#issuecomment-1875546386 for more on _that_ :)) But it's still a contrast to https://www.iana.org/assignments/media-types/application/xml, which lists `charset` explicitly and links to the relevant RFC, which also mentions it. (RFC 9512 mentions support for non-UTF-8 charsets as a difference between JSON (no) and YAML (yes); https://yaml.org/spec/1.2.2/ likewise mentions other encodings. Neither appears to speak directly to `charset`.)

Issue #1927, opened 2014-12-22
Make Functions.constant covariant in argument type.


vincentk:
Presently, a constant function is provided in the `Functions` class with a type signature as follows

``` java
public static <E> Function<Object, E> constant(@Nullable E value) 
```

Personally, I think that a type signature as follows would be superior

``` java
public static <Q, E> Function<Q, E> constant(@Nullable E value) 
```

the reason being that while functions in the "mathematical" sense are co-variant with regard to their argument types, generics in java are not (for well-known reasons).

It's probably not trivial to introduce such a change without triggering a compilation error here or there. Still I think it might be worth considering.

PR #7037, opened 2024-02-25
generalize Functions#constant to accept any key type

singhbaljit:
This generalizes the interface for `Functions#constant` so that it is more lenient. Currently, this method can only be used in places that expect `Function<Object, K>`. Due to the covariance/contravariance rules of Java, this function cannot be passed to an arbitrary method, i.e. a method that may accept `Function<Integer, E>`. The alternative is to do `Functions.forSupplier(Supplies.ofInstance(...))`, which is clunky. 

This is sort of a breaking change for those using the method with explicit type definition using `<>`. The interface goes from `Functions.<E>constant` to `Functions.<F, T>constant`. However, most people are probably not doing...

### DIFF
--- a/guava-tests/test/com/google/common/base/FunctionsTest.java
+++ b/guava-tests/test/com/google/common/base/FunctionsTest.java
@@ -350,11 +350,11 @@ public class FunctionsTest extends TestCase {
   }
 
   public void testConstant() {
-    Function<@Nullable Object, Object> f = Functions.<Object>constant("correct");
+    Function<@Nullable Object, Object> f = Functions.constant("correct");
     assertEquals("correct", f.apply(new Object()));
     assertEquals("correct", f.apply(null));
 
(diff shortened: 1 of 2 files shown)


cpovirk:
You have come across a surprisingly controversial API... :)

We decided years ago not to provide this sort of control. There is an explanation in https://github.com/google/guava/issues/1927#issuecomment-68305122, and you can see some other discussion in that thread.

You also raise a good point about the `Functions.<E>constant` change. While such a change maintains binary compatibility (which is the thing we care much more about), it would surely break some existing users at build time. (I see a good number in our monorepo who would have to change.)

Additionally, as `Function.constant` mentions in its Javadoc, we recommend `x -> value` over `constant(value)` nowadays, so the method is unlikely to see many new users.

Given that, we'll keep the signature as it is now, whether that would have been the right choice if we were designing it from scratch today or not.

PR #7038, opened 2024-02-26
Add local cache unit test for long async refresh case

paggynie:
This is a follow up on #6851 about adding unit test to avoid similar breaking change happen again. This is a generic test to test async refresh on another thread takes long enough for entry to be expired and we still expect to view the reload result without any exception.

I verified with change introduced from #6851 , the test failed with 
```
Tests run: 9571, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 1.677 sec <<< FAILURE!
testLongAsyncRefresh(com.google.common.cache.LocalCacheTest)  Time elapsed: 0.102 sec  <<< ERROR!
java.lang.IllegalStateException: Recursive load of: test
```

after the change from #6851 roll back, the test passes
```
Running com.google.common.cache.LocalCacheTest
Tests run: 9571, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 4.962 sec
```

### DIFF
--- a/guava-tests/test/com/google/common/cache/LocalCacheTest.java
+++ b/guava-tests/test/com/google/common/cache/LocalCacheTest.java
@@ -58,6 +58,8 @@ import com.google.common.testing.FakeTicker;
 import com.google.common.testing.NullPointerTester;
 import com.google.common.testing.SerializableTester;
 import com.google.common.testing.TestLogHandler;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListenableFutureTask;
 import com.google.common.util.concurrent.UncheckedExecutionException;
 import java.io.Serializable;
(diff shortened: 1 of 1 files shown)


cpovirk:
@paggynie, my internal reviewer ended up challenging me to rewrite this test in a way that avoided `sleep` and otherwise guaranteed that it would catch the bug, no matter what happened with scheduling. I think I managed that in https://github.com/google/guava/commit/d5fbccac90aba8501c633e896ea67e2b0bfb426d, but if anything worries you about my changes, let me know. I still left the PR credited to you, since you reported the bug and provided the PR that I started from. But if anyone tells you there was something wrong with the commit that we merged with your name on it, you can point to my explanation here :) Sorry for taking so long on this, and sorry again for the bug.

PR #7066, opened 2024-03-03
Resolve iFrame-Injection and Open-Redirection.

RedYetiDev:
> [!NOTE]
> This is a security issue, but due to the nature of it, it is not confidential.

This pull request resolves the possibility of Open-Redirection and iFrame-Injection in older versions of the Guava documentation. This pull request implements the `validURL` function present in later versions of JavaDoc.


RedYetiDev:
Hi! Any update on this?

cpovirk:
Sorry about that.

I see that the `validURL` function in 16.0 differs slightly from the function you've added for 1.0 - 15.0...

```patch
     if (targetPage.indexOf(":") != -1 || (targetPage != "" && !validURL(targetPage)))
         targetPage = "undefined";
     function validURL(url) {
-        try {
-            url = decodeURIComponent(url);
-        }
-        catch (error) {
-            return false;
-        }
         var pos = url.indexOf(".html");
         if (pos == -1 || pos != url.length - 5)
             return false;
@@ -29,8 +23,7 @@
             if ('a' <= ch && ch <= 'z' ||
                     'A' <= ch && ch <= 'Z' ||
                     ch == '$' ||
-                    ch == '_' ||
-                    ch.charCodeAt(0) > 127) {
+                    ch == '_') {
                 allowNumber = true;
                 allowSep = true;
             } else if ('0' <= ch && ch <= '9'
```

...but I see that the one you're adding matches, e.g., 26.0-jre. So if there's anything to do there, it would likely be to update 16.0 (and others?) to match. I assume that that change is less important than validating the URL at all, but let me know if I'm wrong.

Thanks for the fix. I don't have a full sense of the impact, but I know we've encountered the general principle that you don't want a random mix of sources in your frameset. (At minimum, we used to linked to Javadoc that didn't allow embedding in our frameset.)

RedYetiDev:
It does differ, but it shouldn't cause an issue; this version works, and I figured that rather to match it to the Guava versions, I should find a version that I know will work and just use that.

PR #7067, opened 2024-03-04
Add back `Boolean` constructor calls to `BooleansTest`

Marcono1234:
The test explicitly checks that other `Boolean` instances, created with `new Boolean(...)`, are canonicalized to `Boolean.TRUE` and `Boolean.FALSE`.

See 25ebb6447803ef07906ed4d5432c0cce27b0ff2d which added this test.

Partially reverts #7026, CC @kluever
I think the other changes there are fine though.

However, if there are internal reasons why `new Boolean(...)` cannot be used then maybe it would be necessary to rewrite or completely remove these tests.

### DIFF
--- a/android/guava-tests/test/com/google/common/primitives/BooleansTest.java
+++ b/android/guava-tests/test/com/google/common/primitives/BooleansTest.java
@@ -576,16 +576,16 @@ public class BooleansTest extends TestCase {
 
   public void testAsListCanonicalValues() {
     List<Boolean> list = Booleans.asList(true, false);
-    assertThat(list.get(0)).isSameInstanceAs(true);
-    assertThat(list.get(1)).isSameInstanceAs(false);
+    assertThat(list.get(0)).isSameInstanceAs(Boolean.TRUE);
+    assertThat(list.get(1)).isSameInstanceAs(Boolean.FALSE);
     @SuppressWarnings("deprecation")
-    Boolean anotherTrue = true;
+    Boolean anotherTrue = new Boolean(true); // intentionally calls `new Boolean(...)`
     @SuppressWarnings("deprecation")
-    Boolean anotherFalse = false;
+    Boolean anotherFalse = new Boolean(false); // intentionally calls `new Boolean(...)`
     list.set(0, anotherTrue);
-    assertThat(list.get(0)).isSameInstanceAs(true);
+    assertThat(list.get(0)).isSameInstanceAs(Boolean.TRUE);
     list.set(1, anotherFalse);
-    assertThat(list.get(1)).isSameInstanceAs(false);
+    assertThat(list.get(1)).isSameInstanceAs(Boolean.FALSE);
   }
 
   public void testCountTrue() {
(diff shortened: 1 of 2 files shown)


Marcono1234:
Interestingly the JLS section 25ebb6447803ef07906ed4d5432c0cce27b0ff2d refers to was made stricter in recent versions apparently (by [JDK-7190924](https://bugs.openjdk.org/browse/JDK-7190924)?):
> If the value `p` being boxed is the result of evaluating a _constant expression_ ...

([JLS 17 §5.1.7](https://docs.oracle.com/javase/specs/jls/se17/html/jls-5.html#jls-5.1.7))

If one assumes that the compiler uses `Boolean.valueOf` then [its documentation](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Boolean.html#valueOf(boolean)) guarantees that the values are canonicalized.
However, the JLS does not seem to require that `Boolean.valueOf` is used (see [related Stack Overflow question](https://stackoverflow.com/q/31445024)).

So it seems technically it is allowed for boxing non-constant `boolean` values to return distinct `Boolean` instances.
Maybe it would be necessary to mention this to the OpenJDK maintainers; possibly it was not intentional that JDK-7190924 made it stricter and left the behavior for non-constant values unspecified (this 'breaking change' is also mentioned in [this related Stack Overflow answer](https://stackoverflow.com/a/59637974)). It could then be clarified whether the JLS can or should guarantee that the `valueOf` method of the wrapper classes is used for boxing.

@eamonnmcmanus, would it be necessary to adjust the documentation of Guava's `Booleans.asList` again to remove that guarantee?

cpovirk:
Thanks for digging up the history and the surprising spec change. I'll be interested to hear people's thoughts here, including whether @eamonnmcmanus wants to pursue the spec discussion upstream.

- I don't think the guarantee we make here is of much practical significance. My recollection of my interpretation of @eamonnmcmanus 's earlier change was that it was strengthening/correcting/clarifying the copy-pasted Javadoc from `Ints`, etc., where the same-instance guarantee does not hold. But that's only my recollection of my interpretation.
- Hmm, the guarantee might hold for `Bytes.asList`, too, come to think of it :)
- @kluever 's change was made to prepare for the removal of the primitive-wrapper classes somewhere in the future as part of Valhalla. However, we've since heard that that removal might not need to happen, so we might be able to revert the change on those grounds alone.

kluever:
> @kluever 's change was made to prepare for the removal of the primitive-wrapper classes somewhere in the future as part of Valhalla. However, we've since heard that that removal might not need to happen, so we might be able to revert the change on those grounds alone.

+1 happy to rollback my change given they aren't being removed.

eamonnmcmanus:
This was a good catch. (Hello @Marcono1234!)

I had not noticed the spec change, which I think was to allow very memory-limited devices some wiggle room in how they manage boxing. I think it would be absurd to extend that to boxed non-constant booleans, though, given that there are only two of them. So I'm comfortable with assuming that every boxed boolean is one of two canonical instances (which are also `Boolean.TRUE` and `Boolean.FALSE`). If we really wanted to avoid that assumption, we'd need to make the test change here _and_ change [the code](https://github.com/google/guava/blob/b775e685131023d1dbafd98f10b8ff299a6fe22a/guava/src/com/google/common/primitives/Booleans.java#L406) from this:
```java
      return array[start + index];
```
to this:
```java
      return array[start + index] ? Boolean.TRUE : Boolean.FALSE;
```

Obviously I don't think we should do that.

It would be good if the JLS were cleaned up, so it would at least say that every boxed boolean is one of two canonical instances. But I admit I am not motivated to drive that change.

PR #7079, opened 2024-03-05
Add action version comments in GitHub workflow files

### LINKED ISSUE #7068: Add action version comments in GitHub workflow files
The GitHub workflow files of Guava currently refer to action versions by commit hash, for example:
https://github.com/google/guava/blob/a6a34dc42627f2dc5b8937f428322739b74ce8f3/.github/workflows/ci.yml#L69-L70

While this is good from a security perspective, it makes it difficult at a glance to see which version number (such as "v1.2.3") of that action is used.

What do you think about adding version comments, such as `# v1.2.3`, behind the commit hashes? It seems Dependabot will [automatically update those comments](https://github.blog/changelog/2022-10-31-dependabot-now-updates-comments-in-github-actions-workflows-referencing-action-versions/).

Marcono1234:
Resolves #7068

The `scorecard.yml` workflow file already had version comments, so no changes were needed there.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,13 +32,13 @@ jobs:
     steps:
       # Cancel any previous runs for the same branch that are still running.
       - name: 'Cancel previous runs'
-        uses: styfle/cancel-workflow-action@85880fa0301c86cca9da44039ee3bb12d3bedbfa
+        uses: styfle/cancel-workflow-action@85880fa0301c86cca9da44039ee3bb12d3bedbfa # 0.12.1
         with:
           access_token: ${{ github.token }}
       - name: 'Check out repository'
-        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -32,13 +32,13 @@ jobs:
     steps:
       # Cancel any previous runs for the same branch that are still running.
       - name: 'Cancel previous runs'
-        uses: styfle/cancel-workflow-action@85880fa0301c86cca9da44039ee3bb12d3bedbfa
+        uses: styfle/cancel-workflow-action@85880fa0301c86cca9da44039ee3bb12d3bedbfa # 0.12.1
```
Marcono1234, line 35: The version number here has no leading `v` because the Git tags in that repository don't have it either. Dependabot [should be able to update this as well](https://github.com/dependabot/dependabot-core/blob/795dafedf31666de6e91b078b799fd7f2517478b/github_actions/spec/dependabot/github_actions/file_updater_spec.rb#L384).

PR #7089, opened 2024-03-08
chore(ci): improvements to gha workflows

sgammon:
## Summary

Applies updates to GHA CI, and refactors into [Reusable Workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows). Hardens supply chain security for Guava both in CI and release. Fixes and closes #7088.

### Related PRs

- **Optional**: Included in #7094 

### Security hardening

- CI flows now have an enumerated set of allowed internet endpoints. Other endpoint access is blocked.
- CI flows can now publish [SLSA provenance metadata](https://slsa.dev)
- Maven will now reject dependencies which experience hash mismatches
- Maven dependencies for Guava are now reported to the [GitHub Dependency Graph](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-the-dependency-graph)

### Enclosed features

- Ran [StepSecurity](https://stepsecurity.io) on current jobs
- Added support for [SLSA Provenance](https://slsa.dev) during the release flow
- Added support for [Sigstore](https://sigstore.dev) signing during the release flow
- Added support for the [Github Dependency Graph](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-the-dependency-graph) and [Dependency Review](https://github.com/actions/dependency-review-action)
- Added support for scans via [CodeQL](https://codeql.github.com/)
- All CI builds and tests now appear on one screen

### Changelog

<details>
<summary>Expand for full changelog</summary>
<pre>
- chore(ci): apply hardening to ci jobs
- chore: apply 'Harden Runner' auditing to all ci tasks
- chore: apply `persist-credentials: false` to checkout tasks
- chore: publish dependency graph and add dependency review check
- chore: add codeql scan job (temp)
- chore: bump `actions/checkout` → `4.1.1`
- chore: bump `actions/dependency-review-action` → `4.1.3`
- chore(ci): refactor into reusable workflows
- chore: move ci jobs to `workflow_call` trigger
- chore: add entrypoint jobs for PR and Push events
- chore: cleanup permissions and dispatch checks/tests
- chore(ci): switch to enforced hardening mode
- chore: gather and apply network endpoints for each job
- chore: move to `block` mode for `egress-policy` in `step-security/harden-runner`
- feat(ci): slsa provenance support
- feat: add slsa support to build workflow
- chore: split `test` into `build` and `test` workflows
- chore: use new workflows (build/test) from push/pr triggers
- chore(build): parameterize deploy repositories
[...]

PR #7092, opened 2024-03-10
fix: drop jdk srzip javadoc infra

sgammon:
## Summary

Drops the JDK source steps involved with Javadoc builds, as suggested in #7089. There will be some build warnings related to Javadoc until subsequent PRs are merged.

Split out from work in #7094 

## Changelog

- chore: drop jdk srczip dependencies
- chore: cleanup comments and config related to srczip

### DIFF
--- a/android/guava/pom.xml
+++ b/android/guava/pom.xml
@@ -106,23 +106,8 @@
       <plugin>
         <artifactId>maven-source-plugin</artifactId>
       </plugin>
-      <!-- TODO(cpovirk): include JDK sources when building testlib doc, too -->
       <plugin>
         <artifactId>maven-dependency-plugin</artifactId>
-        <executions>
-          <execution>
-            <id>unpack-jdk-sources</id>
-            <phase>generate-sources</phase>
-            <goals><goal>unpack-dependencies</goal></goals>
-            <configuration>
-              <includeArtifactIds>srczip</includeArtifactIds>
-              <outputDirectory>${project.build.directory}/jdk-sources</outputDirectory>
-              <silent>false</silent>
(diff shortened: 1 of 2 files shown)


android/guava/pom.xml
```diff
@@ -106,23 +106,8 @@
       <plugin>
         <artifactId>maven-source-plugin</artifactId>
       </plugin>
-      <!-- TODO(cpovirk): include JDK sources when building testlib doc, too -->
       <plugin>
         <artifactId>maven-dependency-plugin</artifactId>
```
cpovirk, line 110: Do you happen to have thoughts on whether it might make sense to remove this entirely? I'm guessing so, but I'm not sure.

cpovirk (commented):
Thanks. One question in case you already gave it some thought. In the meantime, I'll try it out and see what happens.

cpovirk:
As for:

> There will be some build warnings related to Javadoc until subsequent PRs are merged.

I got curious, and I found `ValueGraph` + `primitives` + `ExecutionSequencer` changes in https://github.com/google/guava/pull/7094/files#diff-2d2576b049285a93c10e132a6ad16ea883efc4bca40d4c9d68d42c93170ea424, plus `AbstractFuture` and `ExecutionSequencer` changes for what looks to me like a questionable behavior by `javadoc`, in which it appears to be ignoring supermethods' `@throws` clauses for unchecked exceptions? Thanks for taking care of those.

cpovirk:
(And it looks like those probably exist at head, too, at least when using Java 21. That's fine; I'm just noting that this PR doesn't make things worse (even temporarily) AFAICT.)

PR #7087, opened 2024-03-08
Fix: Java 18+ Builds

sgammon:
## Summary

Amends the build to be compatible with JDKs newer than 18. Adds JDK21 to the CI matrix.

### Related issues

- Fixes and closes #7065 
- Fixes testsuite up to JVM21 w.r.t. #6245 

## Changelog

- fix: build against embedded jdk sources needs `--enable-preview` when built against jvm21
- fix: add `-Djava.security.manager=allow` to fix security manager tests under modern java
- fix: upgrade asm → latest
- chore: add jvm21 to ci

PR #7094, opened 2024-03-10
Guava + JPMS

### LINKED ISSUE #2970: Add module-info.java

sgammon:
## Summary

This PR adds modular Java support to Guava, by adding a `module-info.java` definition to the `guava` and `guava-testlib` modules. Upstream work in the [Error Prone Compiler](https://github.com/google/error-prone/pull/4311), [J2ObjC Annotations](https://github.com/google/j2objc/pull/2302), and [Checker Framework](https://github.com/typetools/checker-framework/pull/6326) has also completed, rendering Guava able to build entirely on the `modulepath`.

Fixes and closes #2970 

### Current Status

- ✅ Guava builds as a module
- ✅ Guava builds completely on `modulepath`
- ✅ All tests pass
- ✅ All warnings fixed (within reason, some cannot be fixed or suppressed)
- ✅ Javadoc builds at JVM21
- ✅ Integration and smoke tests
- ✅ Final PR cleanup + autorebase

<details>
<summary>Old PR tree</summary>

### PR Tree

On the way to a finished PR, there were a number of other changes and improvements that surfaced. These have been broken out into their own PRs for easier review; they are listed below as _Downstream_ PRs.

PRs that are _Included_ in this PR can be rebased away, either when they are merged or closed without merging, at the Guava team's choosing.

The following PRs relate to this one:

- **Downstream:** (Pending: benchmark split)
- **Downstream:** #7093
- **Downstream:** #7092
- **Included:** #7089
- **Included:** #7087 
- **Upstream:** google/error-prone#4314
- **Upstream:** google/error-prone#4311
- **Upstream:** google/j2objc#2302
- **Upstream:** typetools/checker-framework#6326

</details>

### Using Modular Guava

From a Java 9+ application, in **`module-info.java`**:
```java
module my.module {
  requires com.google.common;
}
```

**Maven:**
```xml
<dependency>
  <groupId>com.google.guava</groupId>
  <artifactId>guava</artifactId>
  <version>33.4.0-jre-jpms</version>
</dependency>
```

**Gradle (Groovy):**
```groovy
dependencies {
  api('com.google.guava:guava:33.4.0-jre-jpms')
}
```

**Gradle (Kotlin):**
```kotlin
dependencies {
  api("com.google.guava:guava:33.4.0-jre-jpms")
}
```

**Gradle (Version Catalog + Kotlin):**
```toml
[versions]
guava = "33.4.0-jre-jpms"

[libraries]
guava = { module = "com.google.guava:guava", version.ref = "guava" }
```
```kotlin
dependencies {
  api(libs.guava)
}
```

### Forcing Resolution

Sometimes, your build will resolve to a different version of Guava. This usually happens because of a transitive dependency. Here are some instructions for Gradle:

**`build.gradle.kts`:**
```kotlin
configurations.all {
  resolutionStrategy.force("com.google.guava:guava:33.4.0-jre-jpms")
}
```

## Changelog

<details>
<summary>Expand for full changelog</summary>
<pre>
- feat: add `module-info.java` to `guava` module
- feat(jpms): add `module-info.java` to `failureaccess`
- feat(jpms): add `module-info.java` to `testlib`
- fix: necessary fixes to get testsuite running on modular java
- chore: update `guava` to build mrjar
- chore: adjust dev version → `1.0-HEAD-[jre|android]-SNAPSHOT`
- chore: upgrade maven compiler plugin → `3.14.0`
</pre>
</details>

### DIFF
--- a/android/pom.xml
+++ b/android/pom.xml
@@ -6,7 +6,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>com.google.guava</groupId>
   <artifactId>guava-parent</artifactId>
-  <version>HEAD-android-SNAPSHOT</version>
+  <version>1.0-HEAD-android-SNAPSHOT</version>
   <packaging>pom</packaging>
   <name>Guava Maven Parent</name>
   <description>Parent for guava artifacts</description>
@@ -42,7 +42,7 @@
     <module.status>integration</module.status>
     <variant.jvmEnvironment>android</variant.jvmEnvironment>
     <variant.jvmEnvironmentVariantName>android</variant.jvmEnvironmentVariantName>
-    <otherVariant.version>HEAD-jre-SNAPSHOT</otherVariant.version>
+    <otherVariant.version>1.0-HEAD-jre-SNAPSHOT</otherVariant.version>
     <otherVariant.jvmEnvironment>standard-jvm</otherVariant.jvmEnvironment>
     <otherVariant.jvmEnvironmentVariantName>jre</otherVariant.jvmEnvironmentVariantName>
   </properties>
@@ -86,6 +86,7 @@
     <module>guava-bom</module>
     <module>guava-testlib</module>
     <module>guava-tests</module>
+    <module>../futures/failureaccess</module>
   </modules>
   <build>
     <!-- Handle where Guava deviates from Maven defaults -->
@@ -211,7 +212,7 @@
         </plugin>
         <plugin>
           <artifactId>maven-jar-plugin</artifactId>
-          <version>3.2.0</version>
+          <version>3.4.0</version>
         </plugin>
         <plugin>
           <artifactId>maven-javadoc-plugin</artifactId>
--- a/guava-testlib/src/module-info.java
+++ b/guava-testlib/src/module-info.java
@@ -1,0 +1,30 @@
+/*
+ * Copyright (C) 2024 The Guava Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+/**
+ * Guava Testlib
+ */
+open module com.google.common.testlib {
+  requires java.logging;
+  requires com.google.common;
+  requires com.google.common.util.concurrent.internal;
+
+  exports com.google.common.collect.testing;
+  exports com.google.common.collect.testing.features;
+  exports com.google.common.collect.testing.google;
+  exports com.google.common.collect.testing.testers;
+  exports com.google.common.escape.testing;
+  exports com.google.common.testing;
+  exports com.google.common.util.concurrent.testing;
+}
--- a/guava/src/module-info.java
+++ b/guava/src/module-info.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright (C) 2008 The Guava Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Google Guava
+ */
+module com.google.common {
+  requires static jdk.unsupported;
+  requires java.logging;
+  requires com.google.common.util.concurrent.internal;
+
+  exports com.google.common.annotations;
+  exports com.google.common.base;
+  exports com.google.common.cache;
+  exports com.google.common.collect;
+  exports com.google.common.escape;
+  exports com.google.common.eventbus;
+  exports com.google.common.graph;
+  exports com.google.common.hash;
+  exports com.google.common.html;
+  exports com.google.common.io;
+  exports com.google.common.math;
+  exports com.google.common.net;
+  exports com.google.common.primitives;
+  exports com.google.common.reflect;
+  exports com.google.common.util.concurrent;
+  exports com.google.common.xml;
+}
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>com.google.guava</groupId>
   <artifactId>guava-parent</artifactId>
-  <version>HEAD-jre-SNAPSHOT</version>
+  <version>1.0-HEAD-jre-SNAPSHOT</version>
   <packaging>pom</packaging>
   <name>Guava Maven Parent</name>
   <description>Parent for guava artifacts</description>
@@ -42,7 +42,7 @@
     <module.status>integration</module.status>
     <variant.jvmEnvironment>standard-jvm</variant.jvmEnvironment>
     <variant.jvmEnvironmentVariantName>jre</variant.jvmEnvironmentVariantName>
-    <otherVariant.version>HEAD-android-SNAPSHOT</otherVariant.version>
+    <otherVariant.version>1.0-HEAD-android-SNAPSHOT</otherVariant.version>
     <otherVariant.jvmEnvironment>android</otherVariant.jvmEnvironment>
     <otherVariant.jvmEnvironmentVariantName>android</otherVariant.jvmEnvironmentVariantName>
   </properties>
@@ -87,6 +87,7 @@
     <module>guava-gwt</module>
     <module>guava-testlib</module>
     <module>guava-tests</module>
+    <module>futures/failureaccess</module>
   </modules>
   <build>
     <!-- Handle where Guava deviates from Maven defaults -->
@@ -146,15 +147,7 @@
             <target>1.8</target>
             <encoding>UTF-8</encoding>
             <parameters>true</parameters>
-            <compilerArgs>
-              <!--
-                   Make includes/excludes fully work:
-                   https://issues.apache.org/jira/browse/MCOMPILER-174
-
-                   (Compare what guava-gwt has to do for maven-javadoc-plugin.)
-              -->
-              <arg>-sourcepath</arg>
-              <arg>doesnotexist</arg>
+            <compilerArgs combine.children="override">
               <!-- https://errorprone.info/docs/installation#maven -->
               <arg>-XDcompilePolicy=simple</arg>
               <arg>--should-stop=ifError=FLOW</arg>
@@ -182,6 +175,7 @@
               <arg>-J--add-exports=jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED</arg>
               <arg>-J--add-opens=jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED</arg>
               <arg>-J--add-opens=jdk.compiler/com.sun.tools.javac.comp=ALL-UNNAMED</arg>
+              <arg>-Xlint:-removal,-options</arg>
             </compilerArgs>
             <annotationProcessorPaths>
               <path>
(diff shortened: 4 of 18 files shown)


guava/src/module-info.java
```diff
@@ -0,0 +1,45 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Google Guava
+ */
+module com.google.common {
+  requires java.base;
```
norrisjeremy, line 23: I don't think you need to explicitly declare a requirement for `java.base`.
See [JLS 7.7.1](https://docs.oracle.com/javase/specs/jls/se11/html/jls-7.html#jls-7.7.1):
```
If the declaration of a module does not express a dependence on the java.base module, and the module is not itself java.base, then the module has an implicitly declared dependence on the java.base module.
```

sgammon: I actually learned this the other day; thanks for catching this. I'll drop`java.base` from the module definitions when I return to clean up.

guava/src/module-info.java
```diff
@@ -0,0 +1,45 @@
...
+  exports com.google.common.hash;
+  exports com.google.common.html;
+  exports com.google.common.io;
+  exports com.google.common.math;
+  exports com.google.common.net;
+  exports com.google.common.primitives;
+  exports com.google.common.reflect;
+  exports com.google.common.util.concurrent;
+  exports com.google.common.xml;
+  exports com.google.thirdparty.publicsuffix;
```
norrisjeremy, line 44: Should `com.google.thirdparty.publicsuffix` actually be exported?
All the public classes in this package state the following, which makes it seems like they are meant for internal use only:
```
Do not use this class directly. For access to public-suffix information, use com.google.common.net.InternetDomainName.
```

cpovirk: Thanks, yes, we should have put those in `common.net` itself so that we could have made them package-private. (We may still move them someday, but it requires untangling some things in our internal builds.) It would be better to not export them.

guava/src/module-info.java
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * Google Guava
+ */
+module com.google.common {
+  requires java.base;
+  requires java.logging;
+  requires jdk.unsupported;
```
norrisjeremy, line 25: Curious as to which parts of Guava require `jdk.unsupported`?
Not sure if it's correct to explicitly add a module requirement on it, since this is not a module dictated in the JLS, so it could be changed/removed/altered in some incompatible way in future Java releases?

guava/src/module-info.java
sgammon: Easy fix, thanks for spotting

guava/src/module-info.java
sgammon: @norrisjeremy I believe this is because of `sun.misc.Unsafe`; I'm not sure there is any way around `requires jdk.unsupported` at least for the time being. Once Unsafe is refactored away this can be removed. We're 10 years in to the launch of JPMS and `jdk.unsupported` is still around, and it is supported on all current Java versions up to latest LTS, so it should be okay.

norrisjeremy: Ok, I suppose.
But not sure if you have seen [JDK-8323072](https://bugs.openjdk.org/browse/JDK-8323072) & it's associated [draft JEP](https://openjdk.org/jeps/8323072), which I believe may land in Java 23 later this year.
I think the doomsday clock for `sun.misc.Unsafe` is advancing ever closer to midnight.

sgammon: @norrisjeremy Oh, agreed, for sure; but this PR just focused on JPMS for now. JPMS and Unsafe are both sensitive changes that should be done soon, no question.

Java 22 comes out this month, in about 10 days. So you're right to be concerned for Java 23, but Guava has some time to address that before it becomes a problem. JPMS, meanwhile, has been out since Java 9 and is _currently_ unsupported by Guava, so this PR advances on that front.

Even if Unsafe is deprecated in Java 23 (I don't foresee it getting removed completely yet), this changeset will let Guava work at Java 9 - Java 22 with fully modular Java. Then, working to refactor/fix the `Unsafe` usages becomes easier, because Guava is shipping MRJARs, so `Unsafe` can safely be used for earlier JVMs only, with later JVMs providing classes that override with other intrinsics.

norrisjeremy: Yes, that makes sense.
Btw, finally adding JPMS support is greatly appreciated btw. :)

sgammon: @norrisjeremy Thank you likewise for the reviews :)

Sineaggi: Does guava require unsafe, or can it safely fallback?

sgammon: @Sineaggi If I understand correctly, the implications of that are basically the modular equivalent of a compile-time dependency, right? So users could effectively opt-in by adding `requires jdk.unsupported`?

In any case, great catch. You're right: the testsuite passes in full with only `requires static`.

@cpovirk Curious about your thoughts on behavior defaults here. By the way: _technically_ it **is legal** to have a slightly different `module-info.java` for different versions of a single JAR. The public API surface of the module graph can't change, but including things from `java` or `jdk` was specifically imagined to be possible, so that newer class suites can leverage the latest JVM features.

That should mean that it is possible, in theory, to build with `requires` or `requires static` for `jdk.unsupported` based on the active JVM. It would add some build complexity, but it's possible. So I think that means the options are:

- **`requires jdk.unsupported`:** All users are opting in to `Unsafe` whether they like it or not; this is a drawback but the benefit is that the library performs as expected out of the box (i.e. it implies no performance hit from falling back from `Unsafe`)

- **`requires static jdk.unsupported`:** All users can opt-in to use of `Unsafe`; when they upgrade Guava, depending on the shape of their workload, they may see a performance hit, but _only_ if they are using fully modular Java, which would be a new use case at that library version anyway. Otherwise, on the classpath, there is no isolation and so there is no difference.

- **Varied by JVM level:** Perhaps, for users of Java 9 or even Java 11, we use `requires jdk.unsupported`, so that the library performs as expected for the vast "legacy" use cases, which happen to begin building with Guava on the `modulepath` simply because it ships a `module-info.java`. For Java 11+, the `module-info.java` specifies `requires static jdk.unsupported`, and we trust more modern Java users (who are building in a fully modular manner, remember, classpath is still fine) to add `requires jdk.unsupported` if they need it.

Thoughts? Maybe this is overthinking it, I'm not sure.

sgammon: (Also, it would be helpful to see the actual performance hit that should realistically be expected; I'm midway through breaking the benchmark suite out and modernizing with JMH, so perhaps we'll look to some numbers from there or Calibre to help inform this. I think a big question mark is exactly _what_ performance degradation loss would be expected with the loss of `Unsafe` as a default.)

norrisjeremy: Yes, perhaps adding `requires static jdk.unsupported` to avoid a hard requirement on it at runtime would work?
I believe users that are performance sensitive and also desiring the ability to then also use Guava on the module-path instead of class-path could still make it work by adding `--add-exports jdk.unsupported/sun.misc=com.google.common --add-opens jdk.unsupported/sun.misc=com.google.common` to their application.

sgammon:
New complication to this: `failureaccess` will need to see a modular release. I've added a definition and set the current version to `1.0.3-jpms`. It will be added shortly to the attic repo.

Sineaggi:
What's left to get this merged?

sgammon:
Guava with JPMS is released and under testing via the [javamodules.dev](https://javamodules.dev) repository. I am using it in several projects with no problems.

sgammon:
With Netty and other libraries expressing interest as well, I plan to come back to this shortly.

sgammon:
@cpovirk This is cleaned up now and ready for review. Two commits, 18 files changed, and less than 400 net lines adjusted.

pom.xml
```diff
@@ -87,6 +87,7 @@
     <module>guava-gwt</module>
     <module>guava-testlib</module>
     <module>guava-tests</module>
+    <module>futures/failureaccess</module>
```
cpovirk, line 90: I suspect that we omitted this here before because we don't want to implicitly build and release `failureaccess` each time that we release Guava. (That's mainly because we're trying to release only ~one~ ~two~ ~three~ four versions of it. I'm not sure if we had a great reason for that, and now that we're up to four, it's possible that we should just give up.)

So I'm wondering if you'd expect this to be necessary for us to change in our repo or if it would be enough for us to manually publish `failureaccess:1.0.3` and then update `guava` to depend on that.

Alternatively, I assume that we could set up Maven to just _not_ automatically deploy `failureaccess` (while still giving us some way to do so when desired).

guava-testlib/src/module-info.java
```diff
@@ -0,0 +1,30 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+/**
+ * Guava Testlib
+ */
+open module com.google.common.testlib {
```
cpovirk, line 18: It seems fine to be more permissive for the test library here by using `open module`: The stakes are just lower for `guava-testlib` than for `guava`. So, at this point, I'm just curious as to whether this ends up being necessary for something in particular or whether you left it this way to be conservative in general. I think that _our_ tests pass even without `open`, assuming that I tested it right.

android/pom.xml
```diff
@@ -6,7 +6,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>com.google.guava</groupId>
   <artifactId>guava-parent</artifactId>
-  <version>HEAD-android-SNAPSHOT</version>
+  <version>1.0-HEAD-android-SNAPSHOT</version>
```
cpovirk, line 9: I assume that you don't know of any specific harm that would arise if we were to modularize the `android` flavor, as well? There is probably _some_ risk, since Android-focused tools are less likely to be aware of modules than JVM-focused tools, but I'd expect it to be pretty small at this point. The upside would be that cross-platform libraries (which may declare a dependency on `guava-android` even when most of their users run on the JVM) would still be able to use a properly modularized Guava.

I ask because our internal setup actually makes it _easier_ for me to make the change in both places. This wouldn't require any changes on your part; in my internal copy of this PR, our internal setup _already_ made the changes to `android` :) But in the unlikely event that anyone knows of reasons for me to undo that, I can do so.

guava/src/module-info.java
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Google Guava
```
cpovirk, line 18: I am noting for the record that we still get non-modular _Javadoc_ output. That's good, since it avoids breaking any links, as would happen with the modular directory structure. (It also means that we don't need to think hard about what to put in this Javadoc comment :))

cpovirk (commented):
Thanks. I'm probably going to see if I can submit the change to the format of the snapshot version numbers first, and then I may see what else I can carve off before the modularization itself.

pom.xml
sgammon: I understand, and that makes sense. It has been re-added here because `1.0.3` cannot be resolved from Maven Central until it is published. After that, it may be removed again.

`failureaccess` last was released when `Automatic-Module-Name` was added. This attribute causes problems with `jlink` and other use cases, so a release must take place to make `failureaccess` a true module. Thus, one more release probably needs to happen.

It just needs to be present in the reactor project order so it can be noticed by `install` in CI, which needs to build it instead of fetching it through Maven.

If I've missed something here and we _don't_ need to update `failureaccess` for whatever reason, it would simplify the PR a lot. That JAR can't be on the modulepath without `module-info.class`, and can't be on the classpath because of encapsulation and split packages.

guava-testlib/src/module-info.java
sgammon: Left it this way to be conservative in general :) although happy to drop as well.

android/pom.xml
sgammon: > I assume that you don't know of any specific harm that would arise if we were to modularize the android flavor, as well?

I am not aware of any problem this would cause. Android supports up to JVM 21 bytecode, and IIRC modules are ignored past the `javac` phase.

> There is probably some risk [for Android...]

Yeah, but keep in mind we are packaging the module within the version root at `META-INF/versions/9/*`, so a badly behaving tool would need to (1) be unaware or lack support for modules, _and_ would have to (2) scan JARs on the modulepath blindly, _and_ would have to (3) violate JDK 8 spec (by not skipping that as an invalid class name), _and_ would have to (4) have updated Guava to the latest version, implying it is maintained.

This fear keeps me up at night, too, but I think that actually affords quite a bit of protection. Even ancient tools which were released pre-JDK8 should have no issue since classes were not valid in `META-INF/*` at all back then. Thus, the risky part would be shipping the `module-info.class` at the root of the JAR, which probably isn't even that risky. That could be done in a few versions once confidence is in place, or when Java 9 becomes the minimum.

sgammon:
@cpovirk Responses in :) also, CI run should pass on approval

sgammon:
@cpovirk I'm keeping this PR rebased and up to date as you pluck things off. If you'd rather I keep this static, for reference, just let me know and I can restore it

cpovirk:
Ah, sorry, I don't mean to make work for you. I do have what's left pending in two internal CLs (one to bump the version of `failureaccess` and one to do the rest). I can turn those into PRs so that there's a cleanly mergeable PR available so that you don't need to deal with the merge conflicts that I'm creating. Since I changed `failureaccess` around to keep it "separate," both of those PRs will fail CI until I release `failureaccess-1.0.3`. I hope to get back to that all soon; I just need to set aside time for some routine licensing reviews.

cpovirk:
(I should add that it was very valuable to have it in a single, end-to-end PR so that I could see how it all comes together and passes tests!)

sgammon:
@cpovirk Oh no I don't mind, I just want to provide a clean view of the changes. So far merging has been easy because I'm just (mostly) dropping things already merged in to `master`, rendering this PR smaller as we go.

Carry on :) very excited for this change

cpovirk:
[Fixed](https://github.com/google/guava/commit/1d19ee019bba6ef21e1fb38b63fd0d144c7db305).

PR #7128, opened 2024-03-26
Fix Sonar rule S2293 The diamond operator ('<>') should be used

jlerbsc:
We propose the fix to violation s2293 The diamond operator ("<>") should be used identified by SonarQube.

Java uses angular brackets (< and >) to provide a specific type (the "type argument") to a generic type. For instance, List is a generic type, so a list containing strings can be declared with List<String>.
Prior to Java 7, the type argument had to be provided explicitly for every occurrence where generics were used. This often caused redundancy, as the type argument would have to be provided both when a field is declared and initialized.
Java 7 introduced the diamond operator (<>) to reduce the code’s verbosity of generics code. The type argument between the angular brackets should be omitted if the compiler can infer it. For instance, instead of having to declare a List's type in both its declaration and its constructor, you can now simplify the constructor declaration with <>, and the compiler will infer the type.

This patch has been automatically produced by our java code remediation solution, available free of charge for all open source projects (https://www.indepth.fr).

We believe that this PR can improve the quality of the Guava project code to a certain extent. Your feedback will also be very useful for us to know if our solution produces quality code or if we need to improve the way it works.

Thank you for your feedback.

### DIFF
--- a/android/guava-testlib/src/com/google/common/collect/testing/AbstractIteratorTester.java
+++ b/android/guava-testlib/src/com/google/common/collect/testing/AbstractIteratorTester.java
@@ -137,12 +137,12 @@ abstract class AbstractIteratorTester<E extends @Nullable Object, I extends Iter
      * The elements to be returned by future calls to {@code next()}, with the first at the top of
      * the stack.
      */
-    final Stack<E> nextElements = new Stack<E>();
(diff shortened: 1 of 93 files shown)


jlerbsc:
There are some remaining code formatting errors.

PR #7131, opened 2024-03-29
Uses HMAC-SHA256 instead of HMAC-MD5

Shivam7-1:
## Description:
This PR updates the cryptographic algorithm used in the `hmacMd5` method to address security vulnerabilities associated with MD5.

## Changes:
- Replaced the usage of HMAC-MD5 with HMAC-SHA256 in the `hmacMd5` method to enhance security.

### DIFF
--- a/guava/src/com/google/common/hash/Hashing.java
+++ b/guava/src/com/google/common/hash/Hashing.java
@@ -35,6 +35,7 @@ import java.util.zip.Checksum;
 import javax.annotation.CheckForNull;
 import javax.crypto.spec.SecretKeySpec;
 
+
 /**
  * Static methods to obtain {@link HashFunction} instances, and other static hashing-related
  * utilities.
@@ -302,9 +303,10 @@ public final class Hashing {
    * @param key the key material of the secret key
    * @since 20.0
    */
-  public static HashFunction hmacMd5(byte[] key) {
-    return hmacMd5(new SecretKeySpec(checkNotNull(key), "HmacMD5"));
-  }
+public static HashFunction hmacSha256(byte[] key) {
+    return hmacSha256(new SecretKeySpec(checkNotNull(key), "HmacSHA256"));
+}
+
 
   /**
    * Returns a hash function implementing the Message Authentication Code (MAC) algorithm, using the


cpovirk:
Do you know the latest on whether HMAC-MD5 is vulnerable? I know that, _many_ years ago, I heard that [HMAC-MD5 was safe even though MD5 was not](https://rdist.root.org/2009/10/29/stop-using-unsafe-keyed-hashes-use-hmac/#:~:text=because%20a%20collision%20alone%20is%20not%20enough%20to%20compromise%20it.). But that could easily have changed in the meantime.

If it is vulnerable nowadays, we'd probably respond by deprecating the method (as we did for [`Hashing.md5`](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/hash/Hashing.html#md5())). I don't think we'd change the behavior in place because users may be relying on it to remain consistent.

(Also, once we figure this out: It looks like there's another overload of the method (accepting a `Key`) that we'd also want to change.)

cpovirk:
https://en.wikipedia.org/wiki/HMAC#Security suggests no major developments since the doc I cited above. (See also https://crypto.stackexchange.com/a/9340/115834.)

I could still see an argument for deprecating on the grounds that "for a new protocol design, a ciphersuite with HMAC-MD5 should not be included." But given how bad plain MD5 is, and given that we also use deprecation for `murmur3_32` (because we got its implementation slightly wrong), I'd want to draw a clear distinction between how bad those two are and how bad `hmacMd5​` is. And I'd worry that deprecating `hmacMd5​` would muddy that.

Still, I think we could at least add a comment to the method to warn user about the advice to note use HMAC-MD5 in a "new protocol design." I'm not sure how many Guava users are designing a protocol for authenticating encrypted data, and I would hope that any such users have a better source of advice that me :) (In contrast, it's easier to imagine that a Guava users might design a system that uses a more "normal" hash and then might pick MD5 because it's familiar.) But adding a comment is easy enough that I think I'll do it now.

Shivam7-1:
Hi @cpovirk Thanks For Review
Do I need to make any changes from my side ?

PR #7155, opened 2024-04-12
Instead of Using HmacSHA1 Uses SHA-256 {patch}

Shivam7-1:
Description:

This PR addresses security vulnerabilities related to the use of weak cryptographic algorithms in the codebase. Specifically, it replaces the usage of the weak algorithms MD5 and HmacSHA1 with more secure alternatives.

Changes Made:

1. MessageDigestHashFunction:
   - Replaced algorithm parameter with a hardcoded value for SHA-256 to enforce the use of a secure hashing algorithm.
   - Removed support for weak algorithms like MD5.

2. Updated hmacSha1 Method:
   - Renamed method to hmacSha256 to reflect the use of HMAC with SHA-256.
   - Updated the cryptographic algorithm from HmacSHA1 to HmacSHA256 for enhanced security.

Reasoning:

Using weak cryptographic algorithms like MD5 and HmacSHA1 can leave data vulnerable to decryption attacks. By replacing them with stronger alternatives like SHA-256 and HMAC with SHA-256, we improve the overall security posture of the application.

### DIFF
--- a/guava/src/com/google/common/hash/Hashing.java
+++ b/guava/src/com/google/common/hash/Hashing.java
@@ -334,9 +334,10 @@ public final class Hashing {
    * @param key the key material of the secret key
    * @since 20.0
    */
-  public static HashFunction hmacSha1(byte[] key) {
-    return hmacSha1(new SecretKeySpec(checkNotNull(key), "HmacSHA1"));
-  }
+  public static HashFunction hmacSha256(byte[] key) {
+    return hmacSha256(new SecretKeySpec(checkNotNull(key), "HmacSHA256"));
+}
+
 
   /**
    * Returns a hash function implementing the Message Authentication Code (MAC) algorithm, using the


cpovirk:
I have a question similar to https://github.com/google/guava/pull/7131#issuecomment-2027396550: Do you know of resources that suggest avoiding HMAC-SHA-1? The situation seems similar to HMAC-MD5 (where MD5 is broken but HMAC-MD5 is not) but seemingly not as bad: SHA-1 is broken, and HMAC-SHA-1 is not, and unlike HMAC-MD5, I haven't found any resources that recommend against HMAC-SHA-1.

Shivam7-1:
Hi @cpovirk 

Thank you for your question regarding HMAC-SHA-1. While it is true that HMAC-SHA-1 has not been compromised to the same extent as SHA-1 itself, the trend in the security community is to move towards more robust algorithms due to the potential for future vulnerabilities. Here are some points to consider:

- **SHA-1 Vulnerabilities**: SHA-1 has been demonstrated to be vulnerable to collision attacks, which, although not directly affecting HMAC-SHA-1, do raise concerns about the hash function's overall security integrity².
-Given the vulnerabilities found in SHA-1, it is considered good security practice to preemptively replace it with stronger alternatives like SHA-256, even in the context of HMAC².
- **Recommendations**: Some security resources and discussions suggest that while HMAC-SHA-1 may still be secure, using a stronger hash like SHA-256 is recommended over SHA-1 due to the latter's demonstrated weaknesses²⁴.

By updating to HMAC-SHA-256, we are not only addressing current security concerns but also future-proofing our application against potential advancements that could exploit SHA-1's vulnerabilities. This proactive approach aligns with best practices for cryptographic security.

References Which i used
(1) How Secure is it to Use HMAC with SHA-1? | IToolkit. https://itoolkit.co/blog/2023/08/how-secure-is-it-to-use-hmac-with-sha-1/.
(2) HMAC and SHA-1 weakness? - Cryptography Stack Exchange. https://crypto.stackexchange.com/questions/40129/hmac-and-sha-1-weakness.
(3) NVD - CVE-2020-15098. https://nvd.nist.gov/vuln/detail/CVE-2020-15098.
(4) SHA-1 & HMAC OTP Frequently Asked Questions - Initiative for Open .... https://openauthentication.org/wp-content/uploads/2015/09/FAQ1.pdf.
(5) Differential Power Analysis of HMAC SHA-1 and HMAC SHA-2 in ... - Springer. https://link.springer.com/chapter/10.1007/978-3-319-25915-4_19.

cpovirk:
- I don't have access to the Springer paper, but I would expect for some of the other resources to have reflected any practical attack that was identified from 9 years ago.
- The itoolkit asserts that HMAC-SHA-1 is risky but doesn't back it up.
- At least one of the other links explicitly says that it still considers HMAC-SHA-1 to be secure.

Based on that, I don't think there's anything for us to do here.

Issue #7255, opened 2024-06-13
Implement of RateLimiter.setRate can be simplified


xianwdong:
```
public final void setRate(double permitsPerSecond) {
    checkArgument(
        permitsPerSecond > 0.0 && !Double.isNaN(permitsPerSecond), "rate must be positive");
    synchronized (mutex()) {
      doSetRate(permitsPerSecond, stopwatch.readMicros());
    }
  }
```
when permitsPerSecond is NaN，NaN > 0.0 return false, so check "!Double.isNaN(permitsPerSecond)" is redundant

PR #7259, opened 2024-06-16
Simplified RateLimiter.setRate

moinb:
- This is fix for Issue #7255
- Here setRate method takes double as argument so the condition '!Double.isNaN(permitsPerSecond)' is always 'true' when reached and it is redundant check.
- Reference: https://docs.oracle.com/en/java/javase/17/docs//api/java.base/java/lang/Double.html#isNaN(double)


moinb:
Deleted as google-cla wasn't reflecting on this PR. Re-opened a new pull request - https://github.com/google/guava/pull/7260

PR #7271, opened 2024-06-24
Rename field scale to quantileScaleValue to comply with Java S1700 standard suggested by SonarLint

ASangelina:
This PR aims to comply with the Java S1700 standard as suggested by SonarLint, which recommends avoiding field names that are identical to their enclosing class names. The following change was made:

- The field private final int scale in the Scale class was renamed to private final int quantileScaleValue to provide a more meaningful name and avoid conflicts.

**Before**
`public static final class Scale {
    private final int scale;
    // other members and methods
}
`
**After**
`public static final class Scale {
    private final int quantileScaleValue;
    // other members and methods
}
`

These changes help improve code readability and maintainability by following recommended best practices.

Please review the changes and provide your feedback.


cpovirk:
I don't think there's a problem with the current naming.

PR #7272, opened 2024-06-24
Fix single | to || for conditional checks to comply with Java S2638 standard

ASangelina:
This PR addresses the Java S2638 standard, which recommends using the short-circuit operator || instead of the bitwise operator | for conditional checks. The following change was made:

**Before:**
`if (localInputFuture == null | localExceptionType == null | localFallback == null
    // This check, unlike all the others, is a volatile read
    || isCancelled()) {
  return;
}
`
**After:**
`if (localInputFuture == null || localExceptionType == null || localFallback == null
    // This check, unlike all the others, is a volatile read
    || isCancelled()) {
  return;
}
`

The use of || ensures that the evaluation short-circuits, improving performance and avoiding potential null pointer exceptions. This change helps to improve the safety and efficiency of the code by following recommended best practices.

Please review the changes and provide your feedback.

### DIFF
--- a/.github/workflows/maven-publish.yml
+++ b/.github/workflows/maven-publish.yml
@@ -1,0 +1,34 @@
+# This workflow will build a package using Maven and then publish it to GitHub packages when a release is created
+# For more information see: https://github.com/actions/setup-java/blob/main/docs/advanced-usage.md#apache-maven-with-a-settings-path
+
+name: Maven Package
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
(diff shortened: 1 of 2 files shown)


cpovirk:
This is [intentional](https://github.com/google/guava/blob/faf36af34739ddf747eb896ee448688e17e3882b/guava/src/com/google/common/util/concurrent/AbstractFuture.java#L70-L78), but we should note that in `AbstractCatchingFuture`, too.

I will add that, in general, I have found most Sonar checks to be unhelpful, so we are unlikely to accept PRs except in cases in which there are actual risks of bugs or significant performance improvements. I'm sure that there are various Sonar checks that meet that standard, and maybe Sonar offers a way to prioritize those. I haven't used it enough to know.

PR #7301, opened 2024-07-10
BugFix Implemented try-with-resources for URLClassLoader in the class…

dk2k:
… FinalizableReferenceQueue


dk2k:
CLA was provided and it's details include my email for github

eamonnmcmanus (approved):
Seems reasonable. Since finalizers are [deprecated for removal](https://openjdk.org/jeps/421), we expect that `FinalizableReferenceQueue` will eventually disappear too. But meanwhile this is an improvement. I'm going to run this against all of Google's internal tests just to be on the safe side, but I don't expect there will be any issues.

Because of the way the Guava source code is managed, this PR will be superseded by another one coming from our internal source repo. You will be credited.

Issue #7319, opened 2024-07-19
Lingering threads in multiple tests


aoli-al:
### Guava Version

514f212777da34df57fb9ce875ce77948bf95473

### Description

While running Guava tests, I noticed multiple threads created by different tests that were left open. 

Here is the incomplete list:
 
The thread pool was opened here, but the shutdown was not called. 

https://github.com/google/guava/blob/514f212777da34df57fb9ce875ce77948bf95473/guava-tests/test/com/google/common/cache/CacheBuilderTest.java#L496

In FuturesTest.java, multiple tests call `newSingleThreadExecutor()` without shutdown. 

https://github.com/google/guava/blob/514f212777da34df57fb9ce875ce77948bf95473/guava-tests/test/com/google/common/util/concurrent/FuturesTest.java#L444

In TestThread.java, calling `stop()` may throw `UnsupportedOperationException` for the new version of Java. Thus, the `join` method will not be called. 

https://github.com/google/guava/blob/514f212777da34df57fb9ce875ce77948bf95473/guava-tests/test/com/google/common/util/concurrent/TestThread.java#L76

### Example

```java
Any test mentioned above.
```

### Expected Behavior

The test should clean up threads. 

### Actual Behavior

Threads left open. 

### Checklist

- [X] I agree to follow the [code of conduct](https://github.com/google/.github/blob/master/CODE_OF_CONDUCT.md).

- [X] I can reproduce the bug with the [latest](https://github.com/google/guava/releases/latest) version of Guava available.

aoli-al:
#7320 addresses the lingering threads in `CacheBuilderTest` and `FutureTest`; however, the threads in `TestThread` are still not properly closed. 

I tried to call `thread.interrupt` in `TestThread::tearDown`, but it does not work for `UninterruptibleMonitorTest`. Shall I create another issue to track this or just leave this issue open?

cpovirk:
We probably should use `interrupt()` so that we at least get rid of the threads from `InterruptibleMonitorTest`. I think the threads from `UninterruptibleMonitorTest` are unavoidably stuck under newer versions of Java. That is not great, but it should be _mostly_ harmless, I hope. (If we see bigger problems in the future, we could consider trying to run those tests in a separate VM (which I assume they are _not_ run in currently) or even exclude them entirely from our external CI, leaving them to run internally (assuming that we don't see similar problems there :)).)

[...]

PR #7320, opened 2024-07-19
Cleanup threads for CacheBuilderTest and FuturesTest

aoli-al:
This PR addresses #7319. Cleanup threads for `CacheBuilderTest` and `FuturesTest`

### DIFF
--- a/guava-tests/test/com/google/common/cache/CacheBuilderTest.java
+++ b/guava-tests/test/com/google/common/cache/CacheBuilderTest.java
@@ -541,6 +541,8 @@ public class CacheBuilderTest extends TestCase {
     // notification.
     assertEquals(expectedKeys, Sets.union(cache.asMap().keySet(), removalNotifications.keySet()));
     assertTrue(Sets.intersection(cache.asMap().keySet(), removalNotifications.keySet()).isEmpty());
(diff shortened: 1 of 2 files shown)


cpovirk:
(Thanks, I expect to be interested in merging this if you can get the CLA sorted out!)

aoli-al:
Thanks! It seems that the CLA is cleared https://github.com/google/guava/pull/7320/checks?check_run_id=27685732789

ben-manes:
fwiw, technically these should shutdown if the test fails. Typically a try-finally would be ideal, but if you want to avoid that then overriding `setUp` and `tearDown` could work by creating a `Closer` per test instance and registering the shutdown with that.

aoli-al:
Will it be okay to make `service` a class field and then initialize it in each test.  `tearDown` should call `service.shutdown` if service is not null?

cpovirk:
I think the most important thing is shutting down when the test succeeds so that there aren't threads hanging around during some later test run that someone tries to attach a debugger to. And I think we're inconsistent about whether we bother to try to shut down in case of failure.

Let me see whether the internal reviewer is concerned or not, hopefully tomorrow.

Issue #6985, opened 2024-02-12
Support for weak values in `Multimap`


mlichtblau:
### API(s)

```java
`com.google.common.collect.Multimap`, `com.google.common.collect.MultimapBuilder`
```

### How do you want it to be improved?

Support `.weakValues()` similar to `com.google.common.collect.MapMaker`, allowing for individual values to be garbage collected. When all values are garbage collected accessing the value should return an empty collection.

### Why do we need it to be improved?

Currently there is no `.weakValues()` option on the `MultimapBuilder`, but it would be just as useful there as in normal map.

### Example

```java
val multimapBuilder = MultimapBuilder.hashKeys().weakValues().hashSetValues();
```
```


### Current Behavior

Currently there is no way to build a `Multimap` with weak values.

### Desired Behavior

Allow the same behavior like in `MapMaker.weakValues()` in `MultimapBuilder`

### Concrete Use Cases

I want to create a type of notification component where subscribers can wait for the termination of a task with a given task id.
Oftentimes multiple subscribers are interested in subscribing to the same task.
Checking the status of the task in an external system is expensive so each thread polling the status would be too expensive.
My notification component would implement this interface:
```
CompletableFuture<Task> getFinishedTask(TaskId id);
```
In my design of the notification component it would be good to have a `Multimap<TaskId, CompletableFuture>` that keeps track of which task ids should be checked and which completable futures belong to those tasks.
However, as the `CompletableFuture` does not notify the upstream supplier whether the task is cancelled so it is not possible to cleanup tasks that have not yet completed, but are not referenced anymore.
Being able to use weak values in a `Multimap` would allow the futures to garbage collected and prevent the notification component from continuing checking the task once all futures have been garbage collected anyways.

### Checklist

- [X] I agree to follow the [code of conduct](https://github.com/google/.github/blob/master/CODE_OF_CONDUCT.md).

- [X] I have read and understood the [contribution guidelines](https://github.com/google/guava/wiki/HowToContribute#feature-requests).

- [X] I have read and understood [Guava's philosophy](https://github.com/google/guava/wiki/PhilosophyExplained), and I strongly believe that this proposal aligns with it.

anmol797:
Hi , is it still open ? can i work on that ? would love to contribute

bzsil8989:
Hi , Is it still available to contribute ?

YongGoose:
> Hi , Is it still available to contribute ?

See https://github.com/google/guava/pull/7345#issuecomment-2284813797

PR #7345, opened 2024-08-03
Added weakValues() to MultimapBuilder.java

Ayush-Thakur-geek:
Fixes for issue id #6985


netdpb:
Hi @Ayush-Thakur-geek. Thanks for your PR. A few high-level comments:

1. We haven't yet decided to support weak values in `Multimap`s at all (#6985), so a PR is premature at this point.
2. Your PR appears to add a method that has no effect. The request is to add support for weak values, not simply a method called `weakValues`.

As I said, we would have to decide that the feature is worth the trouble to design and implement before we'd accept implementations anyway. I'll close this PR for now.

PR #7371, opened 2024-08-23
Specify `@InlineMe` for `{Doubles,Floats}#compare`

Stephan202:
This mirrors the changes in 04c1b7a8fd353e4f797aa7002d44174219358979 for
similar `#compare` methods.

### DIFF
--- a/android/guava/src/com/google/common/primitives/Doubles.java
+++ b/android/guava/src/com/google/common/primitives/Doubles.java
@@ -25,6 +25,7 @@ import static java.lang.Double.POSITIVE_INFINITY;
 import com.google.common.annotations.GwtCompatible;
 import com.google.common.annotations.GwtIncompatible;
 import com.google.common.base.Converter;
+import com.google.errorprone.annotations.InlineMe;
 import java.io.Serializable;
 import java.util.AbstractList;
 import java.util.Arrays;
@@ -89,6 +90,7 @@ public final class Doubles extends DoublesMethodsForWeb {
    * @return a negative value if {@code a} is less than {@code b}; a positive value if {@code a} is
    *     greater than {@code b}; or zero if they are equal
    */
+  @InlineMe(replacement = "Double.compare(a, b)")
   public static int compare(double a, double b) {
     return Double.compare(a, b);
   }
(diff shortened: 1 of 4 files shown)


Stephan202:
Super fast @cpovirk! :rocket:

Stephan202:
Hey @kluever! Inside Picnic, by default, all Java builds run with Error Prone and `-Werror`, so `@InlineMe` is an effective way for library authors to nudge us away from a (soft-)deprecated API:
- We can't accidentally miss the relevant release note during an upgrade. (Cause usage of an `@InlineMe`-annotated method will cause the build to fail. (Likewise for usage of `@Deprecated`-annotated methods, but the `#compare` methods aren't annotated as such.))
- Thanks for Error Prone's patching feature, each team can accept the inline suggestion simply by running a shared `patch.sh` script: at scale this can save significant developer time.

Our setup also runs [Refaster as an Error Prone check](https://github.com/PicnicSupermarket/error-prone-support/blob/f6e9dbb996df2e90a8af7f9bf6ea61887cfc6a4b/refaster-runner/src/main/java/tech/picnic/errorprone/refaster/runner/Refaster.java), which can achieve similar behavior. That's why PicnicSupermarket/error-prone-support#1293 triggered me to file this pull request: I noticed that we could phase out only 5 out of 7 similar Refaster rules; this PR covers the remainder.

I think it'd be great if more libraries adopted `@InlineMe`; one could even imagine a future where tools such as [Renovate](https://docs.renovatebot.com) and [Dependabot](https://github.com/dependabot) are configured to automatically apply the suggested changes.

Stephan202:
(For anybody interested in what the Picnic build setup concretely looks like: [this Maven parent](https://github.com/PicnicSupermarket/error-prone-support/blob/f6e9dbb996df2e90a8af7f9bf6ea61887cfc6a4b/pom.xml) and [this patch script](https://github.com/PicnicSupermarket/error-prone-support/blob/ca5c3dd3b40a4ae809a9c6a1d654c54818813303/apply-error-prone-suggestions.sh) should give one an idea. Not 100% identical, but close enough.)

kluever:
@Stephan202 Amazing --- that's great to hear! A few years ago, after implementing `@InlineMe`, I spent some time open-sourcing it, and I've always been curious if anyone anyone externally was benefiting from it --- so this anecdote is very helpful. Thank you!

Inside of Google, we actually have the patches produced by `@InlineMe` automatically applied, approved, and submitted across the codebase --- so adding the annotation triggers a very powerful pipeline. I realize that setup doesn't exist externally, but it's still reassuring to hear that it's been useful, even in it's current state.

Issue #1855, opened 2014-10-31
Immutable collections do not guard against wrong Collection.toArray() method


gissuebot:
_[Original issue](https://code.google.com/p/guava-libraries/issues/detail?id=1855) created by **phwendler** on 2014-09-24 at 06:41 AM_

---

ImmutableList etc. can be changed after construction if a collection is passed to copyOf() that keeps a reference to the array returned by toArray(). If the resulting array is of correct size and has more than one element, it is directly used as the backing array of the immutable collection. I attached a file that shows this. The same happens in other places as well, for example in Ordering.immutableSortedCopy().

Such a collection violates the contract of Collection.toArray() which explicitly mentions that the collection may not retain a reference to the array. The only way I could imagine this to happen accidentally might be someone who needs to convert a lot of data back and forth between a List and an array, and implements something like Arrays.asList() but returning the internal array directly in toArray(). Another possibility would be some third-party code that maliciously tries to mangle with internal data structures.

The fix for this would be to do a second copy in these cases, which would probably be a major performance decrease for many users. In our project we do not need this additional protection as we do not have untrusted code, and would certainly prefer the speed of the current solution instead. I guess this is true for a vast majority of users.

A variant of this fix would be to whitelist collections from the java.\* and com.google.\* packages which are known to be well-behaved in this regard, and do a second copy only for other collections. This would be quite ugly and still costs performance for all users that pass in their own collections.

Thus I suggest to keep the current behavior, and instead document this.
Currently the wiki states that Guavas immutable collection are "Safe for use by untrusted libraries", which I would interpret for example such that it is safe to take an immutable collection returned by some untrusted library and use it directly.
This is not true, regardless of whether the library directly returns an immutable collection, or whether it returns an arbitrary collection and my code calls copyOf().
Instead, users retrieving collections from untrusted sources and copying them into an immutable collection need to do a second copy themselves with ImmutableList.copyOf(collection.toArray(new T[0])). (Using toArray() would lose the element type, copying into a List would require yet another copy, and using copyOf(copyOf()) of course does not work because the second call does not copy again.)

PR #7378, opened 2024-08-27
Use defensive copying in `Immutable*Array.copyOf(*Stream)`

perceptron8:


perceptron8:
This looks similar to #1855, so is this a design decision rather than an oversight?

perceptron8:
There's one subtle difference. In case of `Collection.toArray`, "safety" is documented:
> The returned array will be "safe" in that no references to it are maintained by this collection.
(https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#toArray--)

In case of `*Stream.toArray`, it is not.
> Returns an array containing the elements of this stream.
(https://docs.oracle.com/javase/8/docs/api/java/util/stream/DoubleStream.html#toArray--)
(https://docs.oracle.com/javase/8/docs/api/java/util/stream/IntStream.html#toArray--)
(https://docs.oracle.com/javase/8/docs/api/java/util/stream/LongStream.html#toArray--)

cpovirk:
Interesting. My assumption would be that the `*Stream` methods are likewise intended to not maintain a reference. Maybe it's just not documented because a `Stream` that is being collected is not likely to have a preexisting backing array lying around the way that a `Correction` normally does, so there's less need to say "Don't do the natural but unsafe thing" explicitly?

perceptron8:
Makes sense. I think that `toArray()` "safety" may be derived from terminal operation's definition (https://docs.oracle.com/javase/8/docs/api/java/util/stream/package-summary.html#StreamOps).

As you said almost 10 years ago:
> The problem of malicious implementations is one that we've intentionally decided not to tackle.
(https://github.com/google/guava/issues/1855#issuecomment-61355345)

In that case it's my oversight and this PR definitely shouldn't be merged, as it introduces unnecessary overhead.

Thanks for taking the time to look at this!

PR #7387, opened 2024-09-03
Fixed flaky test-case

hermya:
Fixed flaky test case in the following class: guava-tests/test/com/google/common/collect/MultimapsCollectionTest.java

Assertion called in: guava-testlib/src/com/google/common/collect/testing/testers/MapToStringTester.java

POINT OF FAILURE ->
In MapToStringTester class, testToString_formatting() asserts equality in map.toString() and an expected value of map toString() using a LinkedHashMap. This is inconsistent as key ordering obtained in toString() in one call, and the key order observed while iterating through entries in another call can be different.

Fix:
Added a method that retrieves the key order from map.toString() and uses that key order in expected value calculation.

Fixed using NonDex, plugin added in parent pom.
Suggested command to check flaky tests :
> mvn nondex:nondex

For the particular test class:
>mvn -pl ./guava-tests nondex:nondex -Dtest=com.google.common.collect.MultimapsCollectionTest -DnondexRuns=10 -DnondexRuns=10

For more information: https://github.com/TestingResearchIllinois/NonDex

### DIFF
--- a/guava-testlib/src/com/google/common/collect/testing/testers/MapToStringTester.java
+++ b/guava-testlib/src/com/google/common/collect/testing/testers/MapToStringTester.java
@@ -14,22 +14,26 @@
 
 package com.google.common.collect.testing.testers;
 
-import static com.google.common.collect.testing.features.CollectionFeature.NON_STANDARD_TOSTRING;
-import static com.google.common.collect.testing.features.CollectionSize.ONE;
-import static com.google.common.collect.testing.features.CollectionSize.ZERO;
-import static com.google.common.collect.testing.features.MapFeature.ALLOWS_NULL_KEYS;
-import static com.google.common.collect.testing.features.MapFeature.ALLOWS_NULL_VALUES;
-
 import com.google.common.annotations.GwtCompatible;
 import com.google.common.collect.testing.AbstractMapTester;
 import com.google.common.collect.testing.features.CollectionFeature;
 import com.google.common.collect.testing.features.CollectionSize;
 import com.google.common.collect.testing.features.MapFeature;
+import org.junit.Ignore;
+
+import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.Set;
-import org.junit.Ignore;
+import java.util.Objects;
+
+import static com.google.common.collect.testing.features.CollectionFeature.NON_STANDARD_TOSTRING;
+import static com.google.common.collect.testing.features.CollectionSize.ONE;
+import static com.google.common.collect.testing.features.CollectionSize.ZERO;
+import static com.google.common.collect.testing.features.MapFeature.ALLOWS_NULL_KEYS;
+import static com.google.common.collect.testing.features.MapFeature.ALLOWS_NULL_VALUES;
 
 /**
  * A generic JUnit test which tests {@code toString()} operations on a map. Can't be invoked
@@ -42,6 +46,8 @@ import org.junit.Ignore;
 @Ignore // Affects only Android test runner, which respects JUnit 4 annotations on JUnit 3 tests.
 @SuppressWarnings("JUnit4ClassUsedInJUnit3")
 public class MapToStringTester<K, V> extends AbstractMapTester<K, V> {
+
+  private static final Integer LENGTH_OF_COMMA_AND_SPACE = 2;
   public void testToString_minimal() {
     assertNotNull("toString() should not return null", getMap().toString());
   }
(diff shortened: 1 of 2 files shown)


cpovirk:
I have thought on and off for years about finding or writing a tool like NonDex. I hope to be able to someday give it a try!

`HashMap` ordering is another topic near and dear to our team's heart. We've made some changes to the JDK we use for testing to make it randomize the order in which we iterate over `HashMap` buckets. Each time a JVM starts, it picks a random seed that it uses throughout that VM.

Our approach has some advantages and some disadvantages relative to the NonDex scheme. The main _disadvantage_ is that sometimes you have to run a test multiple times in order to detect the hash-ordering dependency. The main _advantage_ is that it is a good match for the risks we see in practice: We have seen changes in iteration order when we upgrade the JVM, but we have not seen changes in iteration order from one iteration over a collection to the next iteration over the same unmodified collection in the same VM. We certainly can't rule out that the latter might someday happen. But given how annoying hash-ordering bugs can be to track down and fix, we don't want to inflict them upon users except in cases in which we think the iteration order might change "for real" in the future.

Given that policy, I'm not too worried about risks like the one you highlight here. If the fix were trivial, I might take it anyway, just to be robust. But it makes the test more complex (and _maaaaaaybe_ slower in some cases, but that's not something I'd consider a serious concern). And, in the absence of NonDex-style bytecode rewriting, I'd expect for the author of a `Map` or `Multimap` class to _want_ to know if the `toString()` ordering and the `entrySet()` ordering differ when no modifications have been made to the multimap, since that could in theory happen accidentally for reasons unrelated to hash ordering. (We could perhaps mostly resolve _that_ concern by having a second test that runs only if `KNOWN_ORDER` is set. But there may still be complexities resulting from how a `Multimap` can have ordered _keys_ but not ordered _values_ and vice versa.)

So, while I like the general idea here, and while I could probably file a dozen NonDex feature requests if you're interested and if I find the time, I think that this particular hash-ordering dependency is not worth worrying about. What do you think?

hermya:
I read your reply, correct me if I'm wrong but, 

Iteration order in a `HashMap` is inconsistent in the same session itself. This testcase wants to assert the output of `toString()` method. Under the hood, `HashMap.toString()` may not give consistent output every time, which is expected but defies the purpose of this test specifically (*You can try this out using the change in* `pom.xml` *and the **mvn** commands I provided in last message*). That's why I proposed the reuse of `key order` extracted from the output of `toString()` called earlier, for a cleaner assertion. Also, I agree that for huge maps, there can be significant impact on time taken in this test execution. 

In my opinion, `HashMap.toString()` should be avoided in test cases altogether, as its assertion gets complicated in many cases.

- for a map with two elements, one can check both permutations of toString() like
```
var exp = map.toString();
Assertions.assertTrue("{A=1, B=2}".equals(exp) || "{B=2, A=1}".equals(exp));
```
but what about map with 20 elements? (`20!` checks)
- What about a Data Structure like `Map<String, Map<String, . . .>>` 

For all such situations, `LinkedHashMap.toString()` or `TreeMap.toString()` is the best, as their keyOrder is fixed. Otherwise, a consistent toString() method feels worthy, and that's why I proposed reusing of key order.
Finally, I'd say that nondex can be used to scrape such "flaky" tests as a part of build process (May be not all builds, but definitely a weekly/monthly check)

cpovirk:
I think we agree that there's nothing in the contract for `HashMap` that guarantees that iteration order will remain the same, even within the same VM and even if the `HashMap` contents aren't modified. My understanding is that, with the historical implementations of `HashMap`, that's turned out to be the case anyway. I understand that NonDex can rewrite the bytecode to demonstrate what would happen if the ordering _were_ to change in that case. But are you aware of any cases in which the ordering does change in practice (within the same VM and without changes to the contents)?

hermya:
It's a tough call. Since morning, I am trying to run functions to obtain indeterministic collections or arrays (Sets, Lists/Arrays, something obtained from java reflection, mappers, etc). In most environments without nondex shuffling, their will be no change in the key order up to a billion calls. My assumption about indeterminism lies on :<br/> **"This class makes no guarantees as to the order of the map; in particular, it does not guarantee that the order will remain constant over time."** from [here](https://docs.oracle.com/javase/8/docs/api/java/util/HashMap.html).<br/> The part of **constant over-time** seems open to interpretation. That being said, I also agree that I haven't seen changes in ordering without change in nondex seed (changing the VM env).

cpovirk:
That's true: The documentation is not clear about what it means. I see a couple historical JDK issues that cite that documentation when discussing what appear to be changes from one version of Java to another:
- https://bugs.openjdk.org/browse/JDK-7176960
- https://bugs.openjdk.org/browse/JDK-8322358

And I see another issue that cites that documentation when discussing something that's really more about identity hash codes (casting some doubt on the idea that maintainer responses on bug threads are universally reliable :smile:):
- https://bugs.openjdk.org/browse/JDK-8227672

It's also true that you can end up with entry _x_ before entry _y_ in the iteration order and then, through subsequent additions for different keys, end up with _y_ before _x_. So that's another way in which the order could be considered not to "remain constant."

I do agree that nothing in the contract actually guarantees that the order remains constant in the absence of changes to contents. Thanks for running tests for that, too. I'm comfortable taking that risk for now. If it comes back to bite us, you can bet that we'll be rushing to start randomizing the order per-iteration after all.

hermya:
Thanks for sharing these threads for reference!

hermya:
### Update

Sorry for a late update. Latest release of nondex works in two modes: **ONE** and **FULL**.
The way to use it would be through the option: `-DnondexMode=<FULL|ONE>`. For example:
```
mvn -pl ./guava-tests edu.illinois:nondex-maven-plugin:2.1.7:nondex -DnondexMode=ONE -Dtest=com.google.common.collect.MultimapsCollectionTest -DnondexRuns=10
```

According to the way you described **guava's testing-jdk** works, this test-case will **pass in ONE mode**. <br> This is because **ONE** mode modifies the iteration order of an iterable object during its first access only. Which means, after creation of a new map, its `map.toString()` OR `map.entrySet()` will return consistent key-ordering, for one VM.<br>On the other hand, **FULL** mode changes it at each '*touch*' i.e. each function call that internally iterates the map.

Hope this information is useful someday!

cpovirk:
Thanks! **ONE** mode sounds promising, and I will try to play around with nondex in that mode to see how it goes.

PR #7396, opened 2024-09-20
Create lintscan

sharkeshd:
Added Lint scanner for code Scanning


chaoren:
Could you please explain what this is and why we should add it?

PR #7425, opened 2024-10-09
fix flaky test testToStringWithNullValue and testToString_formatting

ljmcr:
This PR is a further discussion based on the previous flaky test PR: [https://github.com/google/guava/pull/7387](https://github.com/google/guava/pull/7387)
The two flaky tests are also detected by running with Nondex:
`mvn -pl ./guava-tests nondex:nondex -Dtest=com.google.common.collect.MultimapsCollectionTest -DnondexRuns=10`
The issue arises because the order of keys in a HashMap entry set is not fixed, leading to variations in the toString() output depending on the seed, which causes test failures.

To address this, I’ve explored three potential solutions:
1. Use a different data structure, such as TreeMap, to guarantee key order.
2. Sort the toString() output to achieve a consistent order.
3. Use a comparison method other than assertEquals() for evaluating the results.、

Since these tests are intended to verify the toString() output specifically for HashMap, changing the data structure may alter the test's intent. With efficiency in mind, my proposed solution is to sort the toString() output to ensure consistency.

I would appreciate any feedback on this approach or other suggestions for addressing these flaky tests. Thank you!

### DIFF
--- a/android/guava-testlib/src/com/google/common/collect/testing/google/MultimapToStringTester.java
+++ b/android/guava-testlib/src/com/google/common/collect/testing/google/MultimapToStringTester.java
@@ -14,18 +14,21 @@
 
 package com.google.common.collect.testing.google;
 
-import static com.google.common.collect.testing.features.CollectionFeature.NON_STANDARD_TOSTRING;
-import static com.google.common.collect.testing.features.CollectionSize.ONE;
-import static com.google.common.collect.testing.features.CollectionSize.ZERO;
-import static com.google.common.collect.testing.features.MapFeature.ALLOWS_NULL_KEYS;
-import static com.google.common.collect.testing.features.MapFeature.ALLOWS_NULL_VALUES;
+import java.util.Arrays;
+import java.util.regex.Pattern;
+
+import org.junit.Ignore;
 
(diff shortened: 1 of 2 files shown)


ljmcr:
> Hi @ljmcr. Please sign the CLA. We can't really interact with your PR until you've done so. Thanks.

Hi, I think I have already signed the CLA, the checkbox in the pipeline for CLA is passed in my perspective, can you confirm again, thank you!

netdpb:
> > Hi @ljmcr. Please sign the CLA. We can't really interact with your PR until you've done so. Thanks.
> 
> Hi, I think I have already signed the CLA, the checkbox in the pipeline for CLA is passed in my perspective. Can you confirm again? thank you!

It looks like it worked. Thanks!

netdpb:
I feel like the same reasons that led us to close #7387 apply here as well. @cpovirk, what do you think?

cpovirk:
Yes, I think it's good to have something like this available if we ever do see the level of randomization that Nondex implements. Until then, I think we'd rather have a test failure alert us if our collections or the JDK collections begin to iterate in a different order than we'd expect, since that is likely to affect our users. At that point, we could take action, including applying this patch.

Issue #2193, opened 2015-10-15
mvn compile on root of repo fails


amshali:
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:2.10:unpack-dependencies (unpack-guava-sources) on project guava-gwt:
Artifact has not been packaged yet. When used on reactor artifact, unpack should be executed after packaging: see MDEP-98. -> [Help 1]
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
[ERROR]
[ERROR] After correcting the problems, you can resume the build with the command
[ERROR]   mvn <goals> -rf :guava-gwt

cpovirk:
Does it work if you use `mvn install`?

We should really have some documentation that says to do that, probably on https://github.com/google/guava/wiki/HowToContribute. I had been thinking that we did, but if so, I can't find it.

kevinb9n:
We should have this and readily findable.

cpovirk:
One thing that this could address is https://github.com/google/guava/issues/2235: how to run a specific test. (Arguably that's more of a Maven question, but (a) it would be nice to lower the barrier anyway, and (b) one common "Maven way" of doing it doesn't work for Guava, as noted in that thread.)

cpovirk:
If https://github.com/google/guava/pull/6326 and its followup go through, then we'll be able to recommend `./mvnw clean install`, which will save people from having to install Maven (and possibly find that [Guava doesn't build with the version they have](https://github.com/google/guava/commit/36f7898c3bcd232f44fbef0538f24a6b8403d023)).

cpovirk:
Apparently we've also had https://github.com/google/guava/wiki/ContributorSetUp#build-and-test for a long time now. (It's probably obsolete.) I'm not sure whether we have linked to it from anywhere.... (It's possible that my link here will make the world worse :))

PR #7434, opened 2024-10-14
Replace deprecated Throwables.propagate with throwIfUnchecked and direct exception throwing in SubscriberRegistry, MoreExecutors

panic08:
The suggestion here is to change the use of the deprecated `Throwables.propagate()` in the `SubscriberRegistry.java`, `MoreExecutors.java` classes to the alternative that is recommended in the documentation for this method: 

```
Throwables.throwIfUnchecked(e);
throw new RuntimeException(e.getCause());
```

### DIFF
--- a/guava/src/com/google/common/eventbus/SubscriberRegistry.java
+++ b/guava/src/com/google/common/eventbus/SubscriberRegistry.java
@@ -239,7 +239,8 @@ final class SubscriberRegistry {
     try {
       return flattenHierarchyCache.getUnchecked(concreteClass);
     } catch (UncheckedExecutionException e) {
-      throw Throwables.propagate(e.getCause());
+      Throwables.throwIfUnchecked(e);
+      throw new RuntimeException(e.getCause());
     }
   }
 
--- a/guava/src/com/google/common/util/concurrent/MoreExecutors.java
+++ b/guava/src/com/google/common/util/concurrent/MoreExecutors.java
@@ -808,7 +808,8 @@ public final class MoreExecutors {
     } catch (IllegalAccessException | ClassNotFoundException | NoSuchMethodException e) {
       throw new RuntimeException("Couldn't invoke ThreadManager.currentRequestThreadFactory", e);
     } catch (InvocationTargetException e) {
-      throw Throwables.propagate(e.getCause());
+      Throwables.throwIfUnchecked(e);
+      throw new RuntimeException(e.getCause());
     }
   }
 


cpovirk:
I wanted to make sure that I came back and said something here because this PR prompted a bunch of activity :)

- The PR had a subtle error that we've seen before in each file: Instead of `throwIfUnchecked(e.getCause())`, it had `throwIfUnchecked(e)`.
   - That actually causes a build error in one case (the `InvocationTargetException` case) because our build runs https://errorprone.info/bugpattern/ThrowIfUncheckedKnownChecked. (Sadly, running our build is still slightly more complex than it should be, but [`./mvnw clean install`](https://github.com/google/guava/issues/2193#issuecomment-1429885133) should normally do the trick.)
   - It's _not_ a build error in the other case (the `UncheckedExecutionException` case) because Error Prone doesn't have a checker for a known _unchecked_ exception. (There would probably have been a test error, at least.) I'm proposing adding such a checker in https://github.com/google/error-prone/pull/4625.
   - I think it was fear of that kind of problem that had led us to employ some awkward code elsewhere, so I tried out a new pattern for avoiding such problems in https://github.com/google/guava/pull/7442.
- The existing behavior in each class was questionable:
   - `SubscriberRegistry` was rethrowing an exception that could (in rare cases) come from another thread. This can be confusing, since the stack trace in that thread doesn't necessarily match the stack trace in the thread that eventually catches the exception. I ended up changing that in https://github.com/google/guava/pull/7449, and I just sent up a followup in https://github.com/google/guava/pull/7453.
   - `MoreExecutors` was more reasonable, but I've come to view "`throw new RuntimeException(e)`" as a mild code smell: While it's very rarely actually _important_ to throw a _specific subclass_ of `RuntimeException` instead, "`throw new RuntimeException(e)`" has historically sometimes indicated that the author just didn't want to think about what might have caused the exception or what should actually be done about it. We ended up adding a comment and using `UndeclaredThrowableException`, and then I made similar changes in some other files in https://github.com/google/guava/pull/7451, plus https://github.com/google/truth/pull/1353 in lieu of a similar change elsewhere (which then tipped me off to search for some other Java 7 / Android updates in https://github.com/google/guava/pull/7446 / https://github.com/google/truth/pull/1352).
- Given how much I ended up jumping from place to place, I figured that I'd follow up on some other existing warnings, for which you can see further changes in https://github.com/google/guava/commits/master/.

PR #7483, opened 2024-10-24
Customize redirect template to preserve the URL anchor.

chaoren:
Extracted the default redirect HTML via `wget guava.dev/api` and replaced the destination URL with `{{ page.redirect.to }}` and just added `+location.hash` to the JavaScript redirect URL.

See https://github.com/jekyll/jekyll-redirect-from#customizing-the-redirect-template.


cpovirk (approved):
Nice, thank you!

PR #7494, opened 2024-10-27
Optimize bloomfilter issue 7346

Romain-E:
### Context
This PR optimizes the calculation of the number of hash functions and bits used in BloomFilter, reducing redundant computations and improving readability.

### Changes
Implemented a more efficient algorithm for `optimalNumOfBits`
Replaced hash calculations in `optimalNumOfHashFunctions`
Updated tests to reflect these changes
### Impact
Improves performance when using `BloomFilter`.

### Tests affected
Tests have been modified to validate the new calculation methods.

### DIFF
--- a/guava-tests/test/com/google/common/hash/BloomFilterTest.java
+++ b/guava-tests/test/com/google/common/hash/BloomFilterTest.java
@@ -255,15 +255,16 @@ public class BloomFilterTest extends TestCase {
   /** Tests that we never get an optimal hashes number of zero. */
   public void testOptimalHashes() {
     for (int n = 1; n < 1000; n++) {
-      for (int m = 0; m < 1000; m++) {
-        assertTrue(BloomFilter.optimalNumOfHashFunctions(n, m) > 0);
+      for (double p = 0.1; p > 1e-10; p/=10) {
+        assertTrue(BloomFilter.optimalNumOfHashFunctions(p) > 0);
       }
     }
   }
 
   // https://github.com/google/guava/issues/1781
   public void testOptimalNumOfHashFunctionsRounding() {
-    assertEquals(7, BloomFilter.optimalNumOfHashFunctions(319, 3072));
+    double p = 0.03;
+    assertEquals(5, BloomFilter.optimalNumOfHashFunctions(p));
   }
 
   /** Tests that we always get a non-negative optimal size. */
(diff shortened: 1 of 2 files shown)


guava/src/com/google/common/hash/BloomFilter.java
```diff
@@ -505,18 +511,16 @@ public int hashCode() {
...
    *
-   * @param n expected insertions (must be positive)
-   * @param m total number of bits in Bloom filter (must be positive)
+   * @param p desired false positive probability (must be between 0 and 1, exclusive)
    */
   @VisibleForTesting
-  static int optimalNumOfHashFunctions(long n, long m) {
-    // (m / n) * log(2), but avoid truncation due to division!
-    return max(1, (int) Math.round((double) m / n * Math.log(2)));
+  static int optimalNumOfHashFunctions(double p) {
```
rohitnandi12, line 521: The change here replaces the optimalNumOfHashFunctions method, which originally took parameters n (expected insertions) and m (total number of bits), with a new version that instead uses p (desired false positive probability). While this new approach aligns with cases where p is known, it reduces flexibility for use cases where m and n are more readily available or preferred for direct configuration.

To maintain backward compatibility and allow both methods of calculation, it would be more effective to overload optimalNumOfHashFunctions rather than replacing it. Overloading the method would allow users to specify either n and m or p as needed, based on their use case, without introducing breaking changes.

eamonnmcmanus: I'm not sure what this comment is referring to. This method is not part of the public API, so why would we be concerned with backward compatibility?

guava-tests/test/com/google/common/hash/BloomFilterTest.java
```diff
@@ -255,15 +255,16 @@ public void testNullPointers() {
...
     for (int n = 1; n < 1000; n++) {
-      for (int m = 0; m < 1000; m++) {
-        assertTrue(BloomFilter.optimalNumOfHashFunctions(n, m) > 0);
+      for (double p = 0.1; p > 1e-10; p/=10) {
+        assertTrue(BloomFilter.optimalNumOfHashFunctions(p) > 0);
       }
     }
   }
 
   // https://github.com/google/guava/issues/1781
```
eamonnmcmanus, line 264: Is this still relevant?

guava-tests/test/com/google/common/hash/BloomFilterTest.java
```diff
@@ -255,15 +255,16 @@ public void testNullPointers() {
...
   public void testOptimalNumOfHashFunctionsRounding() {
```
eamonnmcmanus, line 265: Should this name be changed?

I realize that the existing test doesn't do this, but I think it would be good to spell out the calculation that goes from 0.03 to 5 in a comment.

eamonnmcmanus (changes requested):

PR #7549, opened 2024-12-13
Update FarmHashFingerprint64.java

gopinath-vasalamarri:
Remove the data dependency between loads and both the rotates.  

We can rewrite the code in such a way that rotates can start computing as soon as the required loads are present instead of waiting for both the loads to retire.

### DIFF
--- a/guava/src/com/google/common/hash/FarmHashFingerprint64.java
+++ b/guava/src/com/google/common/hash/FarmHashFingerprint64.java
@@ -120,9 +120,9 @@ final class FarmHashFingerprint64 extends AbstractNonStreamingHashFunction {
       long mul = K2 + length * 2L;
       long a = load64(bytes, offset) + K2;
       long b = load64(bytes, offset + length - 8);
-      long c = rotateRight(b, 37) * mul + a;
-      long d = (rotateRight(a, 25) + b) * mul;
-      return hashLength16(c, d, mul);
+      long c = rotateRight(b, 37) * mul;
+      long d = rotateRight(a, 25) * mul;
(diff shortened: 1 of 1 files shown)


cpovirk:
This might be a good improvement, but our existing (internal) benchmarks are kind of questionable, and our fleet-wide profiling shows a negligible amount of CPU in these functions for us, so we are unlikely to improve the benchmarks enough to see if this would help. If you want to try to put something together with JMH, you could, but I'm not even sure I'd guarantee that we'd end up merging this, given that it makes the implementation code different from the C++ implementation. (Now, if you could get the _C++_ implementation updated to match this because this change showed benefits there, then we might follow.)

gopinath-vasalamarri:
Thank you Chris for the suggestions.

One meta question on the comment -- is there a list somewhere, of functions in Guava, which consume the highest amount of CPU ?.

cpovirk:
I don't think we've published such a list. Setting aside any confidentiality issues, there's the additional complication that some methods "consume a lot of CPU" because they call into arbitrary user code (like `ListenableFutureTask.run`). With enough effort, we could probably find a list of methods that consume the most CPU _inside `com.google.common`, though even that wouldn't provide the full picture, since some of our code could get faster by avoiding calling user code so much (e.g., by performing fewer hashings of user objects or equality comparisons on user objects) :) And then sometimes the best way to address a lot of CPU time spent in one of our methods is to change a few big callers to do something else entirely.

There might still be something useful that we could provide here, but those are some of the reasons that it probably won't become a priority, either.

gopinath-vasalamarri:
Got it,  yeah I agree,  ROI seems to be low.

PR #7550, opened 2024-12-14
Update BloomFilterStrategies.java

gopinath-vasalamarri:
Remove imul instruction from murmur operations of 32bit hash functions -- reduces the number of instructions by 1 CPU cycle.

Before -- https://godbolt.org/z/GWvnEoPa3
After -- https://godbolt.org/z/4zEoM5G1W


lowasser:
I see the difference in bytecode, but is there a difference in benchmarks?  I'd tend to expect both versions to get optimized to the same thing.

gopinath-vasalamarri:
yeah,  running a BM is good idea.   By any chance,  do we have one for testing the performance of the strategies ?.   if so,  could you please point me out on how to run them.   


thanks

PR #7598, opened 2025-01-03
DO NOT MERGE

cpovirk:
Test what happens if a JDK that Maven wants is not provided by
`setup-java`.


cpovirk:
Great, it failed as hoped.

PR #7716, opened 2025-03-10
Modify the equals method and do not call the size method

liuyuyan2717:
For the difference method of Sets, which returns a SetView that inherits AbstractSet, I want to override its equals method to reduce the running time of this method. The main operation is to delete the call to the size method.


chaoren:
Please read https://github.com/google/guava/wiki/HowToContribute#code-contributions before submitting pull requests. Please justify this change by filing a feature request with all the relevant information first. Why is this change necessary? Real world examples where this makes a difference?

Issue #7769, opened 2025-04-09
MethodHandle.invoke and MethodHandle.invokeExact are only supported starting with Android O (--min-api 26)


JordanLongstaff:
### Guava Version

33.4.7-jre

### Description

The latest version of Guava uses `MethodHandle` which is only available in Android API 26 or later. My app supports API 21 at the minimum, and it uses the latest release of Guava to patch a security vulnerability, only now it won't build because Guava uses methods that are not supported in the minimum SDK version. And I have no desire to increase it in my app.

### Example

```java
Any Android app with minimum SDK version lower than 26, using the latest version of Guava, would suffice. Here's a PR build from my repo that failed:
https://github.com/JordanLongstaff/ArtemisAgent/actions/runs/14347309233/job/40219434611?pr=189
```

### Expected Behavior

App should still build as normal.

### Actual Behavior

App cannot build because of `MethodHandle`. See the link in the example for more info.

### Platforms

Android

### Checklist

- [x] I agree to follow the [code of conduct](https://github.com/google/.github/blob/master/CODE_OF_CONDUCT.md).

- [x] I can reproduce the bug with the [latest](https://github.com/google/guava/releases/latest) version of Guava available.

cpovirk:
Sorry, our internal toolchain must handle things differently, as we're and to build for minSdkVersion 21 there. (We should ideally set up an external test for this.) Thanks for the fast report of this. I'll look when I start with in an hour or two

cpovirk:
I haven't gotten the build running locally yet, though I made a little progress:

```
/tmp/tmp.wPQ8gN60RZ/ArtemisAgent/keystore.properties (No such file or directory)
```

So:

```
printf '%s=X\n' keyAlias keyPassword storePassword storeFile > keystore.properties
```

But now:

```
* What went wrong:
A problem occurred configuring project ':IAN:annotations:konsist'.
> Failed to apply plugin 'org.gradle.software-reporting-tasks'.
   > Could not create task ':IAN:annotations:konsist:outgoingVariants'.
      > Could not create task of type 'OutgoingVariantsReportTask'.
         > Could not create an instance of type org.gradle.api.tasks.diagnostics.internal.configurations.ConfigurationReportsImpl.
            > Could not create an instance of type org.gradle.api.reporting.internal.DefaultReportContainer.
               > Type T not present
```

[...]

PR #7775, opened 2025-04-10
DO NOT MERGE: AGP integration test.

cpovirk:
This detects the problem of https://github.com/google/guava/issues/7769.

To actually merge this, I'd want to:

- Make it work with the version of Gradle we already use, 8.5, or else
  upgrade the somewhat-standard version of Gradle that's used across a
  handful of Google project, including ours.

- Integrate it into our CI and release processes, including testing with
  the version that we're building, as [in our Gradle integration tests
  for our module
  metadata](https://github.com/google/guava/blob/97c96e00bd8940b45399979cbd4162d3ef362532/integration-tests/gradle/build.gradle.kts#L1).

- Maybe put some patterns into `.gitignore`, as possibly I should have
  along with the previous Gradle integraiotn tests?


cpovirk:
(I'd ideally also want to not make our CI take significantly longer than it already does, as tracked in internal issue 334979149.)

Issue #7855, opened 2025-07-01
testlib: add tests for Map.entrySet().stream()


vlsi:
* `map.entrySet().stream().toList()` should return the same elements
* `spliterator().characteristics()` should probably align with map/set characteristic. For instance, if the map permits nulls, and there are nulls, then `Spliterator.NONNULL` does not sound right.

See https://github.com/ben-manes/caffeine/issues/1883

ben-manes:
I see the mistake exists in every popular non-jdk `ConcurrentMap` implementation that I can find: Guava, Eclipse Collections, Apache Commons' and Spring's ConcurrentReferenceHashMap, JCTools, etc. Some of these implementations predated Java 8's addition and all inherit the default of `SIZED` which is unsafe in their concurrent collection. Including additional test coverage would be very nice, though admittedly I've been unsuccessful in getting testlib adoption when reporting the bugs it finds.

vandan09:
Hi @vlsi,
I’d like to work on this and add test coverage for Map.entrySet().stream().toList() and the corresponding Spliterator characteristics.
Let me know if there are any specific map implementations you'd like covered or testing strategies you'd recommend.
Thanks!

ShaikSalmaAga:
Hi @vlsi,

I hope you are doing well.

I'd love to contribute to this issue. I'm planning to update the `entrySpliterator()` implementation by removing the incorrect use of `Spliterator.NONNULL`, and I'll also add test coverage to ensure the characteristics reflect the actual behavior especially when null values may be present.

Please let me know if there's anything specific you'd like me to focus on or consider while working on this.

Thanks so much.

chaoren:
[`MapTestSuiteBuilder`](https://github.com/google/guava/blob/master/guava-testlib/src/com/google/common/collect/testing/MapTestSuiteBuilder.java) creates a derived test suite for `entrySet()` using [`SetTestSuiteBuilder`](https://github.com/google/guava/blob/master/guava-testlib/src/com/google/common/collect/testing/SetTestSuiteBuilder.java): https://github.com/google/guava/blob/162544b946eb9e69151fd5dd22355a963450588d/guava-testlib/src/com/google/common/collect/testing/MapTestSuiteBuilder.java#L164-L167

[...]

PR #7878, opened 2025-07-17
Add test for Map.entrySet().stream()

ASR1015:
Summary
This PR adds a unit test `testEntrySetStream()` in `MapEntrySetTester` that validates 
the stream count from `Map.entrySet()` matches the size of the map.

Why
Improves test coverage for stream operations on map entry sets.

Notes
- Code builds and all tests pass.
- Ready for review.


ASR1015:
Hi! I've signed the CLA and re-triggered the check. Everything should be in place now. Kindly review when possible. Thank you!

chaoren:
Hi, I think it'd be easier for us to handle this ourselves.

ASR1015:
Thank you for the feedback. Would you mind pointing out the specific issues in the PR so I can learn and improve future contributions?

chaoren:
The test you're adding is already covered indirectly via derived test suites. See my comment here: https://github.com/google/guava/issues/7855#issuecomment-3085056859. The actual issue is something else. You also added the test to the wrong place and the change isn't formatted correctly.

PR #7907, opened 2025-07-29
Add basic test for RateLimiter.tryAcquire() burst behavior

Daksh021:
This pull request adds a test case for `RateLimiter.tryAcquire()` to observe its behavior under a short burst of requests.

 What the Test Covers

- Creates a RateLimiter with 1000 permits/second.
- Fires 20 tryAcquire() calls in rapid succession.
- Asserts that:
  - Some requests are allowed initially (due to stored permits).
  - Some requests are denied once the limiter catches up to the rate limit.

 Rationale

Although testImmediateTryAcquire() checks permit exhaustion for two consecutive calls, this test simulates a higher-throughput scenario to validate how  tryAcquire() behaves under short bursts — closer to real-world usage.

Let me know if any revisions are needed!

### DIFF
--- a/guava-tests/test/com/google/common/util/concurrent/RateLimiterTest.java
+++ b/guava-tests/test/com/google/common/util/concurrent/RateLimiterTest.java
@@ -24,7 +24,7 @@ import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.junit.Assert.assertThrows;
-
+import static org.junit.Assert.assertTrue;
 import com.google.common.collect.ImmutableClassToInstanceMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.testing.NullPointerTester;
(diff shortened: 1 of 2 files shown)


netdpb:
How is this different from #7903?

Daksh021:
@netdpb You are right that testImmediateTryAcquire() already tests the basic behavior of the rate limiter with just two calls: the first one passes, the second fails due to the 1 permit/sec rate.

The test I added tries to simulate a more realistic burst of traffic. For example:
RateLimiter limiter = RateLimiter.create(1000); // 1 permit per millisecond
int allowed = 0, denied = 0;
for (int i = 0; i < 20; i++) {
    if (limiter.tryAcquire()) allowed++;
    else denied++;
}
In this case:
Some requests should pass due to stored or fast-available permits.
The rest should fail, because they are sent faster than 1/ms.
The idea was to verify that tryAcquire() behaves as expected under short, fast bursts — which could happen in real services — and not just in simple two-call sequences.
That said, if this isn't adding meaningful coverage or protection against real regressions, I’m happy to  remove it . Its no different just the mistakes you pointed out I had rectified it .

Daksh021:
> How is this different from #7903?
> @netdpb You are right that testImmediateTryAcquire() already tests the basic behavior of the rate limiter with just two calls: the first one passes, the second fails due to the 1 permit/sec rate.
> 
> The test I added tries to simulate a more realistic burst of traffic. For example: RateLimiter limiter = RateLimiter.create(1000); // 1 permit per millisecond int allowed = 0, denied = 0; for (int i = 0; i < 20; i++) { if (limiter.tryAcquire()) allowed++; else denied++; } In this case: Some requests should pass due to stored or fast-available permits. The rest should fail, because they are sent faster than 1/ms. The idea was to verify that tryAcquire() behaves as expected under short, fast bursts — which could happen in real services — and not just in simple two-call sequences. That said, if this isn't adding meaningful coverage or protection against real regressions, I’m happy to remove it . Its no different just the mistakes you pointed out I had rectified it . I am kind of newbie in open source so if I am troubling you. I am sorry. I just tried to simulate the whole real world scenario because there is no solution to handle all the incoming requests at once it will cause system overload .Either we can create a waiting room type of thing where the denied requests are present and as the servers are free we will send them to fetch necessary info but here the problem is the new requests that come at that point of time will have to wait which will cause latency and bad user experience. If my solution and the code does not contribute anything I am happy to close the PR. I tried my best to solve this problem .

One more approach I thought If, say, 4 requests are in the waiting queue and 4 new ones arrive at the same time, we can allow 2 from the queue and 2 from the new batch balancing fairness and responsiveness.

This approach avoids starvation of old requests while still keeping latency low for fresh ones  something rate limiters alone don’t manage.

Of course, this goes beyond the scope of RateLimiter.tryAcquire() in Guava, but I just wanted to share the thinking that led me to write the test in this PR.

If I caused any confusion or trouble during this process, I sincerely apologize.

netdpb:
This change is identical to the one in #7903. I think you created two PRs by mistake.

PR #7929, opened 2025-08-09
Add project icon for `IntelliJ IDEA`

Pankraz76:
Add project icon for `IntelliJ IDEA` to match other high tier projects, as well.

image

image

- https://github.com/spring-projects/spring-framework/pull/34444
- https://github.com/apache/maven/pull/2283
- https://github.com/diffplug/spotless/pull/2522

https://github.com/google/google-java-format/pull/1264


cpovirk:
Do you know where we can find information about using the Google logo in locations like this? It may well be fine, but I'd want to have a page somewhere that clarifies that we can copy it here, and I'm not sure how quickly we can find such a thing.

Pankraz76:
> clarifies that we can copy it here

did not get this, not sure what you mean.

Yes im wondering too, how this is moved to other repos as well.

Seeing the nice copy tool, this might be something good for all the google projects.

cpovirk:
Sorry, I mean that the part of Google responsible for the logos may have rules for which projects are allowed to use them. My _guess_ is that it's not an issue, but I've seen rules for "similar" things, like which projects have to present themselves on GitHub as "not an official Google project."

PR #7916, opened 2025-08-06
Add `staticanalysis.UnnecessaryParentheses`

Pankraz76:
PoC for:

- https://github.com/google/guava/issues/7917

```
[WARNING] Changes have been made to guava-gwt/src-super/com/google/common/cache/super/com/google/common/cache/LocalCache.java by:
[WARNING]     org.openrewrite.staticanalysis.MissingOverrideAnnotation
[WARNING] Please review and commit the results.
[WARNING] Estimate time saved: 2h 30m
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for Guava Maven Parent 999.0.0-HEAD-jre-SNAPSHOT:
[INFO] 
[INFO] Guava Maven Parent ................................. SUCCESS [  0.270 s]
[INFO] Guava: Google Core Libraries for Java .............. SUCCESS [  1.327 s]
[INFO] Guava BOM .......................................... SUCCESS [  0.028 s]
[INFO] Guava Testing Library .............................. SUCCESS [  0.289 s]
[INFO] Guava Unit Tests ................................... SUCCESS [  0.143 s]
[INFO] Guava GWT compatible libs .......................... SUCCESS [ 43.500 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  45.870 s
[INFO] Finished at: 2025-08-06T11:50:37+02:00
[INFO] ------------------------------------------------------------------------
```

PR #7930, opened 2025-08-09
Apply `UnnecessaryParentheses`

Pankraz76:
enabler for:

- https://github.com/google/guava/pull/7934

### DIFF
--- a/android/guava-tests/test/com/google/common/collect/ListsTest.java
+++ b/android/guava-tests/test/com/google/common/collect/ListsTest.java
@@ -81,7 +81,7 @@ public class ListsTest extends TestCase {
   private static final class RemoveFirstFunction implements Function<String, String>, Serializable {
     @Override
     public String apply(String from) {
-      return (from.length() == 0) ? from : from.substring(1);
+      return from.isEmpty() ? from : from.substring(1);
     }
   }
 
--- a/android/guava-testlib/src/com/google/common/collect/testing/Helpers.java
+++ b/android/guava-testlib/src/com/google/common/collect/testing/Helpers.java
@@ -262,7 +262,7 @@ public class Helpers {
     @Override
     @SuppressWarnings("unchecked") // no less safe than putting it in the map!
     public int compare(Entry<K, V> a, Entry<K, V> b) {
-      return (keyComparator == null)
+      return keyComparator == null
           ? ((Comparable) a.getKey()).compareTo(b.getKey())
           : keyComparator.compare(a.getKey(), b.getKey());
     }
@@ -440,7 +440,7 @@ public class Helpers {
       public int hashCode() {
         K k = getKey();
         V v = getValue();
-        return ((k == null) ? 0 : k.hashCode()) ^ ((v == null) ? 0 : v.hashCode());
+        return (k == null ? 0 : k.hashCode()) ^ (v == null ? 0 : v.hashCode());
       }
 
       @Override
--- a/android/guava-testlib/src/com/google/common/collect/testing/MapInterfaceTest.java
+++ b/android/guava-testlib/src/com/google/common/collect/testing/MapInterfaceTest.java
@@ -687,7 +687,7 @@ public abstract class MapInterfaceTest<K extends @Nullable Object, V extends @Nu
     Set<Entry<K, V>> entriesToRetain =
         singleton(mapEntry(originalEntry.getKey(), originalEntry.getValue()));
     if (supportsRemove) {
-      boolean shouldRemove = (entrySet.size() > entriesToRetain.size());
+      boolean shouldRemove = entrySet.size() > entriesToRetain.size();
       boolean didRemove = entrySet.retainAll(entriesToRetain);
       assertEquals(shouldRemove, didRemove);
       assertEquals(entriesToRetain.size(), map.size());
--- a/android/guava-testlib/src/com/google/common/collect/testing/MinimalSet.java
+++ b/android/guava-testlib/src/com/google/common/collect/testing/MinimalSet.java
@@ -80,7 +80,7 @@ public class MinimalSet<E extends @Nullable Object> extends MinimalCollection<E>
   public int hashCode() {
     int hashCodeSum = 0;
     for (Object o : this) {
-      hashCodeSum += (o == null) ? 0 : o.hashCode();
+      hashCodeSum += o == null ? 0 : o.hashCode();
     }
     return hashCodeSum;
   }
--- a/android/guava-testlib/src/com/google/common/collect/testing/TestsForSetsInJavaUtil.java
+++ b/android/guava-testlib/src/com/google/common/collect/testing/TestsForSetsInJavaUtil.java
@@ -206,7 +206,7 @@ public class TestsForSetsInJavaUtil {
             new TestEnumSetGenerator() {
               @Override
               public Set<AnEnum> create(AnEnum[] elements) {
-                return (elements.length == 0)
+                return elements.length == 0
                     ? EnumSet.noneOf(AnEnum.class)
                     : EnumSet.copyOf(MinimalCollection.of(elements));
               }
--- a/android/guava-testlib/src/com/google/common/collect/testing/google/SetGenerators.java
+++ b/android/guava-testlib/src/com/google/common/collect/testing/google/SetGenerators.java
@@ -179,7 +179,7 @@ public class SetGenerators {
 
   public static class ImmutableSortedSetExplicitComparator extends TestStringSetGenerator {
 
-    private static final Comparator<String> STRING_REVERSED = Collections.reverseOrder();
+    private static final Comparator<String> STRING_REVERSED = Comparator.reverseOrder();
 
     @Override
     protected SortedSet<String> create(String[] elements) {
@@ -194,7 +194,7 @@ public class SetGenerators {
     @SuppressWarnings("CanIgnoreReturnValueSuggester")
     @Override
     public List<String> order(List<String> insertionOrder) {
-      sort(insertionOrder, Collections.reverseOrder());
+      sort(insertionOrder, Comparator.reverseOrder());
       return insertionOrder;
     }
   }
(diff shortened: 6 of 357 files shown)


Pankraz76:
if you consider this let me know i need to revert the maven suggestions then if these could not be applied as well.

Pankraz76:
[INFO] Tests run: 0, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for Guava Maven Parent 999.0.0-HEAD-jre-SNAPSHOT:
[INFO] 
[INFO] Guava Maven Parent ................................. SUCCESS [  0.114 s]
[INFO] Guava: Google Core Libraries for Java .............. SUCCESS [  3.942 s]
[INFO] Guava BOM .......................................... SUCCESS [  0.000 s]
[INFO] Guava Testing Library .............................. SUCCESS [01:40 min]
[INFO] Guava Unit Tests ................................... SUCCESS [07:29 min]
[INFO] Guava GWT compatible libs .......................... SUCCESS [  6.779 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  09:20 min
[INFO] Finished at: 2025-08-09T14:33:11+02:00
[INFO] ------------------------------------------------------------------------

android/guava-tests/test/com/google/common/base/ObjectsTest.java
```diff
@@ -46,7 +46,7 @@ public void testEqual() throws Exception {
 
     // test distinct string objects
     String s1 = "foobar";
-    String s2 = new String(s1);
+    String s2 = s1;
     assertTrue(Objects.equal(s1, s2));
```
Stephan202, line 50: Here and in some test below distinct _instances_ are created on purpose. These changes should be reverted.

android/guava-tests/test/com/google/common/collect/ListsTest.java
```diff
@@ -81,7 +81,7 @@ public class ListsTest extends TestCase {
   private static final class RemoveFirstFunction implements Function<String, String>, Serializable {
     @Override
     public String apply(String from) {
-      return (from.length() == 0) ? from : from.substring(1);
+      return (from.isEmpty()) ? from : from.substring(1);
```
Stephan202, line 84: The wrapping parens can now be dropped as well.

Stephan202 (commented):
The `pom.xml` changes seem unrelated to the application of the Refaster rules. Just some drive-by comments.

android/guava-tests/test/com/google/common/collect/ListsTest.java
Pankraz76: yes, this is a new recipe then, as the goal is not to touch this, just leveraging rewrite for doing all this kind of trivial stuff. Thats exactly what its about. Thanks for demonstrating why its a good invest to spend some CPU time avoiding this upfront. No offence given, really appreciate your feedback and guidance. This is just the perfect example what this is all about.

@cpovirk
@romani

- https://docs.openrewrite.org/recipes/staticanalysis/unnecessaryparentheses

Pankraz76: > can now be dropped as well

Assuming its just as obsolete, as it was before, indicating the broken window effect bleeding the code, due to absents of (proper) quality gate.

Pankraz76:
> The `pom.xml` changes seem unrelated

yes sorry, had dirty working copy including:

- https://github.com/google/guava/pull/7916

Will undo/assuming we gonna merge this dedicated:

- https://github.com/google/guava/pull/7932

android/guava-tests/test/com/google/common/base/ObjectsTest.java
Pankraz76: thx, reverted assuming it as SSOT and is replicated somehow:

image

Pankraz76:
of course this change should not be merges this way, considering simple replication of the `UnnecessaryParentheses` in dedicated branch, pulling this upfront.

cpovirk:
I'd be happy to take the `UnnecessaryParentheses` changes if you're interested in creating a PR for just them. Some of the library changes look like things that we can't take for Android reasons, but `length() == 0` -> `isEmpty()` would also be good, as would (I think) `BigDecimal.valueOf`.

Pankraz76:
no pom changes, just running the recipe on the repo:

image

Pankraz76:
Can rebase at any time, but due to the code density, conflicts are likely to arise soon. This is another good example why the plugin is in need to avoid spending time on issues like this. As it will be happen very soon after its fixed **once** but not forever. Making the whole effort most efficient if fixed for the future as well.

It's either a one-time effort to fix it now, or dealing with it constantly—which is inefficient until it's properly resolved once.

cpovirk:
The best way to avoid spending time on this level of issue is often just to ignore it, sadly. I do think that the mass fix here (which looks to be more aggressive than Error Prone was in https://github.com/google/guava/pull/7696 and https://github.com/google/guava/pull/7771) is worthwhile, and I have sent it for review internally. Still, I think the cost of the first breakage from a build-time check here would outweigh the benefits of the backsliding protection, and I am happy to leave the checking off. If we want to clean up such errors in the future, I think the most efficient way to do that will be another mass run of the check.

Pankraz76:
> sadly



What about tools like Checkstyle? Looking at large-scale projects like PMD and Checkstyle, they don’t have this issue. Maven also uses Spotless to automate these tedious tasks and discussions, no offence. Its about fixing issues like you mentioning not about ignoring stuff. Its not an improvement leaving the flaws open considering this hindering to an easy access of the code.

Investing all the time spent on manual formatting fixes would be better spent on a single commit with a plugin.  

We’re past the era where formatting errors require:  
- **Worst case**: Being ignored  
- **Bad case**: Manual fixes  
- **Best case**: Automation via tools like Spotless  

Major projects like Maven, Quarkus, and Spring all rely on automation—because at scale, there’s no other practical way.

Pankraz76:
This could be made simply by merging and activating the `failOnDryRunResults` param afterwards making the commit easy and reproducible for everyone.

cpovirk:
There is a lot that can be said about both the importance of code simplifications/improvements and the way to make those improvements at the lowest cost (e.g., build errors/warnings vs. automated code-review comments vs. human action vs. periodic cleanups vs. presubmits). (I think that there may be a presentation of one perspective on this in https://storage.googleapis.com/gweb-research2023-media/pubtools/4365.pdf.) Our situation is more complicated still because our Maven build is sort of "secondary" in our normal workflow. Breakages that are delayed to that phase need to provide a lot of value before they pay off.

cpovirk:
As I should probably have anticipated, and as I hinted at in https://github.com/google/guava/pull/7934#issuecomment-3176037903, there's some disagreement over the value of parentheses on ternary conditions. (Opinions range from "always" to "never," with space for intermediate options like "only if the conditional is 'complex' (e.g., `foo == bar || foo instanceof Baz`)" or "not if there's a lot of parenthesization already on the line.") It seems to fall into the range of "not controversial enough that we need to force a rule to prevent discussions but controversial enough that a rule would annoy people." I'm making an attempt to revert those changes and take what's left, but I may end up erring in one direction or the other.

Pankraz76:
> As I should probably have anticipated, and as I hinted at in [#7934 (comment)](https://github.com/google/guava/pull/7934#issuecomment-3176037903), there's some disagreement over the value of parentheses on ternary conditions. (Opinions range from "always" to "never," with space for intermediate options like "only if the conditional is 'complex' (e.g., `foo == bar || foo instanceof Baz`)" or "not if there's a lot of parenthesization already on the line.") It seems to fall into the range of "not controversial enough that we need to force a rule to prevent discussions but controversial enough that a rule would annoy people." I'm making an attempt to revert those changes and take what's left, but I may end up erring in one direction or the other.

yes "unfortunately" thats true, theres always the need for some exception, using suppression serving both parties.

Having a little freedom, while not having to strive in chaos.

Pankraz76:
Assuming something much broader like `RemoveUnusedPrivateMethods` is someone most ppl. would agree upon to avoid. Even tho still some like to preserve making software archeology.

- https://github.com/google/guava/pull/7938

cpovirk:
I forgot to also post: A teammate referred me to https://google.github.io/styleguide/javaguide.html#s4.7-grouping-parentheses, which is the standard we're up against.

The Java changes in https://github.com/google/guava/pull/7938 look good except for the ones that appear to be removing serialization methods that Java calls automatically.

Pankraz76:
> serialization methods that Java calls automatically

Thanks very interesting. Need to ignore these, in the recipe, if this is actually true/good.

Pankraz76:
but again this effort is considered waste, as its not done, it will happen again.

We should try to stop starting and start finishing. As this obsolete stuff will occur anytime again if there is no quality gate imposed. Currently its bug/issue/flaw eldorado. I have fixed bugs in maven due to this kind of issue, so please dont argue its not important.

- https://github.com/apache/maven/pull/2402
- https://github.com/apache/maven/pull/2312
- https://github.com/apache/maven/pull/2310

Pankraz76:
> > serialization methods that Java calls automatically
> 
> Thanks very interesting. Need to ignore these, in the recipe, if this is actually true/good.

Is this something real or just some random google stuff? @timtebeek

https://github.com/search?q=repo%3Agoogle%2Fguava%20writeReplace&type=code

Inconsistent again. Its private. Is it called via reflection then or how should this work in runtime ? 

image



---


Not even the code format is working, its really sad to see this important project in such conditions.

image

Pankraz76:
- https://github.com/google/guava/pull/7944

timtebeek:
> > > serialization methods that Java calls automatically
> > 
> > 
> > Thanks very interesting. Need to ignore these, in the recipe, if this is actually true/good.
> 
> Is this something real or just some random google stuff? @timtebeek

This case was missed; fixed now and likely released later today: https://github.com/openrewrite/rewrite-static-analysis/commit/976d366036dc785107f199d8a3467a2675959469

Issue #7917, opened 2025-08-06
Add `rewrite` support for `errorprone.refasterrules`


Pankraz76:
## Add `rewrite` support for `errorprone.refasterrules`

I’d like to propose integrating [Google’s Error Prone](https://github.com/google/error-prone) and its Picnic extension (demonstrated in [Automating Away Bugs with Error Prone | PlatformCon 2023](https://www.youtube.com/watch?v=6llnrUtVlrE)) to enable automated bug fixes via `rewrite` rules. This would complement Checkstyle’s static analysis capabilities by addressing semantic bugs rather than stylistic issues.

## Motivation
Error Prone’s `refaster`/`rewrite` rules can automatically:
- Fix common bug patterns (e.g., `String.equals()` misuse)
- Modernize code (e.g., JDK migration helpers)
- Enforce best practices (e.g., null-check improvements)

Real-world adoptions show tangible benefits:
- [Apache Kafka’s integration](https://github.com/apache/kafka/pull/20219)
- [Spotless’s implementation](https://github.com/diffplug/spotless/pull/2576)

## Proposal
1. Add support for `errorprone.refasterrules`-based rewrites
2. Implement with opt-in adoption (no breaking changes)
3. Include suppression mechanisms for API constraints

## Discussion Points
- Need consensus on:
- Scope of auto-fixes
- Preferred suppression strategy
- Integration approach

## Next Steps
I’m happy to:
- Prepare a PoC demonstrating the value
- Collaborate on implementation strategy
- Address any concerns about compatibility

relates to:

- https://github.com/checkstyle/checkstyle/issues/16543

### 2. What's the best code you can write to accomplish that without the new feature?

- https://github.com/google/guava/pull/7916

### 3. What would that same code look like if we added your feature?

current state diff:

- https://github.com/google/guava/pull/7916

### (Optional) What would the method signatures for your feature look like?

```java

```

### Concrete Use Cases

https://github.com/checkstyle/checkstyle/issues/17487

### Checklist

- [x] I agree to follow the [code of conduct](https://github.com/google/.github/blob/master/CODE_OF_CONDUCT.md).

- [x] I have read and understood the [contribution guidelines](https://github.com/google/guava/wiki/HowToContribute#feature-requests).

- [x] I have read and understood [Guava's philosophy](https://github.com/google/guava/wiki/PhilosophyExplained), and I strongly believe that this proposal aligns with it.

[...]

PR #7932, opened 2025-08-10
Add `openrewrite.maven.BestPractices`

Pankraz76:
PoC for:

- https://github.com/google/guava/issues/7917
- https://github.com/google/guava/pull/7930


cpovirk:
Maven changes are a bit of a different case than Java changes, so my https://github.com/google/guava/pull/7916#issuecomment-3175212882 isn't as directly relevant here. I have a somewhat similar overall sentiment but without as much to say. I could try to come up with more, but I think the time would be better spent moving on to other Guava work.

PR #7934, opened 2025-08-11
Add `rewrite` support for `UnnecessaryParentheses`

Pankraz76:
Opt in ability plugin to complete:

- https://github.com/google/guava/pull/7930


cpovirk:
I am fairly confident that we won't want this check on. We may still be able to take part of https://github.com/google/guava/pull/7930, but even it is proving more controversial. (Yes, I take the point that it's nice to have rules for things precisely to avoid controversy :) But a concern is that some ternaries become harder to read for some people without parentheses, and people generally aren't so eager to remove the parentheses that we end up with actual controversy during cod review.) Anyway, I'll follow up on #7930 once I hear more from the review.

PR #7945, opened 2025-08-12
Add `rewrite` support for `maven.BestPractices`

Pankraz76:


cpovirk:
It sounds like this will remain a disagreement around the relative costs and benefits of these checks for our specific code base and build integration. If I think of anything to add to https://github.com/google/guava/pull/7916#issuecomment-3175212882 and https://github.com/google/guava/pull/7930#issuecomment-3175874817 in the future, I'll add it, but I think that's most of the story.

PR #7959, opened 2025-08-31
Apply maven BestPractices

Pankraz76:
Hi im wondering if there is any benefit and chance of using:

- https://github.com/google/guava/pull/7961

### DIFF
--- a/pmd.xml
+++ b/pmd.xml
@@ -1,0 +1,13 @@
+<?xml version="1.0"?>
+<ruleset name="DefinitionOfDone"
+         xmlns="http://pmd.sourceforge.net/ruleset/2.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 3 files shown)


guava/src/com/google/common/collect/MutableClassToInstanceMap.java
```diff
@@ -183,10 +183,6 @@ public void putAll(Map<? extends Class<? extends @NonNull B>, ? extends B> map)
     return Primitives.wrap(type).cast(value);
   }
 
-    private Object writeReplace() {
```
Pankraz76, line 186: but this is a bug right?

Pankraz76: fixed.

cpovirk:
I stand by https://github.com/google/guava/pull/7916#issuecomment-3175212882.

PR #7961, opened 2025-09-01
Run maven BestPractices

Pankraz76:


cpovirk:
This clearly means a lot to you, so I can take some parts that are unlikely to require any discussion, though not everything.

Pankraz76:
Thanks for consider.

PR #7966, opened 2025-09-03
Reduce allocations in TypeToken#getRawType

### LINKED ISSUE #7957: TypeToken#getRawType allocates a lot
### Guava Version

33.4.8-jre

### Description

TypeToken#getRawType is needlessly expensive for TypeToken instances where the runtime type is already a Class; it allocates an ImmutableSet builder, a TypeVisitor. Ideally for Class / ParameterizedType, no allocations would occur.

### Example

```java
TypeToken.of(Object.class).getRawType(); // ideally shouldn't allocate
new TypeToken<List<Object>>(){}.getRawType(); // ideally shouldn't allocate, or should allocate less
```

### Expected Behavior

I would not expect `getRawType` to allocate so heavily for trivial TypeTokens.

### Actual Behavior

Several GBs of allocations observed in a 3 minute profile from a production application, or 1% of allocations.

### Packages

com.google.common.reflect

### Platforms

_No response_

### Checklist

kilink:
Special case Class and ParameterizedType handling in getRawType to avoid allocating for the trivial / common case. Fixes #7957.

### DIFF
--- a/guava/src/com/google/common/reflect/TypeToken.java
+++ b/guava/src/com/google/common/reflect/TypeToken.java
@@ -191,10 +191,19 @@ public abstract class TypeToken<T> extends TypeCapture<T> implements Serializabl
    * </ul>
    */
   public final Class<? super T> getRawType() {
-    // For wildcard or type variable, the first bound determines the runtime type.
-    Class<?> rawType = getRawTypes().iterator().next();
-    @SuppressWarnings("unchecked") // raw type is |T|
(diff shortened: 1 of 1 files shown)


cpovirk:
Thanks, I'd somehow gotten into my head that the `Class` case was the only easy one, but as your comment and PR demonstrate, `ParameterizedType` is easy, too.

I will try to get this merged today. (It first has to go through our internal review process.)

Issue #7958, opened 2025-08-30
TypeToken class do not respect Object#hashCode() method contract.


d-william:
### Guava Version

33.4.8-jre

### Description

The TypeToken class do not respect Object#hashCode() method contract.
Especially when the implementation of the hashCode method of the wrapped type is not the same.
The javadoc indicates: 
> If two objects are equal according to the #equals(Object) equals method, then calling the hashCode method on each of the two objects must produce the same integer result.  

This sentence is not respected.

This happens when the type given to the TypeToken comes from another library (TypeUtils of lang3 for example).
Example: The user may retrieve a type from a given API (that internaly use lang3) and want to wrap it in a TypeToken.

Workaround: Normalize the type before creating the TypeToken.
```
Type normalized = new TypeResolver().resolveType(externalType)
TypeToken<?> token = TypeToken.of(normalized);
```

Solution proposal:
 - Normalize the type in the TypeToken#of(Type) method
 - OR Indicate in the javadoc of the TypeToken#of(Type) method the user may encounter this king of issue.
 - OR Make sure the two wrapped type use the same hashing method

### Example

```java
Type externalType = TypeUtils.parameterize(Class.class, String.class);
TypeToken<?> token2 = TypeToken.of(externalType);

System.out.println(Objects.equals(token1, token2)); // true
System.out.println(Objects.equals(token2, token1)); // true

System.out.println(token1.hashCode() == token2.hashCode()); // false
System.out.println(token1.hashCode()); // 550244190
System.out.println(token2.hashCode()); // -560007
```

### Expected Behavior

When two TypeToken are equal, their hashcodes are the same.

### Actual Behavior

When two TypeToken are equal, their hashcode may be different

### Packages

com.google.common.reflect

### Platforms

Any

### Checklist

- [x] I agree to follow the [code of conduct](https://github.com/google/.github/blob/master/CODE_OF_CONDUCT.md).

- [x] I can reproduce the bug with the [latest](https://github.com/google/guava/releases/latest) version of Guava available.

cpovirk:
`TypeToken` relies on the `Type` that it contains to implement `equals` and `hashCode`. Depending on how you create the `TypeToken` instance, the contained `Type` might be implemented by Guava, the JDK, or another library (as in your example).

[...]

PR #7973, opened 2025-09-06
Fix TypeToken hashCode contract violation 

lmcrean:
## Problem Statement

#7958: TypeToken violates the fundamental Object.hashCode() contract when wrapping Type instances from different implementations. The issue occurs when:

1. Two Type instances are semantically equivalent (Type.equals() returns true)
2. But come from different libraries/implementations (e.g., Guava vs Apache Commons Lang3)
3. These implementations have different hashCode() behaviors

This results in equal TypeToken instances having different hash codes, violating the contract:
> If two objects are equal according to equals(), then calling hashCode() on each must produce the same integer result.

## Solution

Normalize input Types in `TypeToken.of(Type)` using the existing TypeResolver infrastructure:

```java
// Before:
return new SimpleTypeToken<>(type);

// After:
return new SimpleTypeToken<>(new TypeResolver().resolveType(type));
```

TypeResolver.resolveType() reconstructs Types using Guava's consistent implementations, ensuring all TypeToken instances wrap Types with compatible equals/hashCode behavior.

## Impact

**Benefits:**
- [x] Fixes fundamental hashCode contract violation
- [x] Ensures HashMap/HashSet safety for TypeToken usage
- [x] Provides consistent behavior regardless of Type source
- [x] Uses existing, well-tested TypeResolver infrastructure

**Risk Assessment:**
- **Performance:** Minimal - TypeResolver is lightweight and already used in TypeToken
- **Compatibility:** 100% backward compatible - existing usage unaffected

## Testing for Maintainers

**New Test Coverage:**
- `TypeTokenTest.testHashCodeContractWithExternalTypes()` reproduces the original issue
- Creates external ParameterizedType with different hashCode implementation
- Verifies equals() works and hashCode contract is respected
- Test fails without fix, passes with fix

**Regression Testing:**
- All existing TypeToken tests continue to pass
- No changes to TypeToken's public API or documented behavior
- TypeResolver normalization is transparent to users

**Manual Verification:**
```bash
# Run the specific test
mvn test -Dtest=TypeTokenTest#testHashCodeContractWithExternalTypes

# Run full TypeToken test suite
mvn test -Dtest=TypeTokenTest
```

## Breaking Changes

**None.** 

This is a pure bug fix that:
- Maintains identical external API
- Preserves all existing TypeToken behavior
- Only affects the internal Type representation (normalized to consistent implementations)
- Has no user-visible changes except fixing the hashCode contract violation

The fix is invisible to correctly functioning code and only resolves the underlying contract violation that could cause issues in hash-based collections.

### DIFF
--- a/guava-tests/test/com/google/common/reflect/TypeTokenTest.java
+++ b/guava-tests/test/com/google/common/reflect/TypeTokenTest.java
@@ -40,11 +40,13 @@ import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
 import java.lang.reflect.WildcardType;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import junit.framework.TestCase;
 import org.jspecify.annotations.NullUnmarked;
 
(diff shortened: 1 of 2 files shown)


cpovirk:
As noted in https://github.com/google/guava/issues/7958#issuecomment-3239638119, the bug is in another library, rather than in Guava. Guava could be changed to work around many bugs in many libraries, but we'd have to have a highly unusual reason to go down that road.

lmcrean:
Thank you @cpovirk for the clarification, I had read that comment and now fully comprehend the externality of the issue.

Issue #6666, opened 2023-08-02
Look for other Maven artifacts that contain Guava classes, and list them in our metadata


cpovirk:
Just as the current metadata can detect conflicts with `google-collections`, it could detect conflicts with those other artifacts.

For example, I keep forgetting that we ourselves released [`guava-base`](https://repo1.maven.org/maven2/com/google/guava/guava-base/r03/), etc. for exactly one release. And others have done this (e.g., [diffplug](https://repo1.maven.org/maven2/com/diffplug/guava/guava-collect/19.0.0/)), too.

Here's [a person who recently got bit by this](https://twitter.com/dirkmahler/status/1667946899297431556).

ben-manes:
[guava-osgi](https://code.google.com/archive/p/guava-osgi) is another one ([stackoverflow](https://stackoverflow.com/questions/76533317/what-is-the-right-way-to-call-com-google-common-collect-lists-cartesianproduct?noredirect=1#comment134954933_76533317))

pan3793:
[guava-jdk5](https://mvnrepository.com/artifact/com.google.guava/guava-jdk5)

pan3793:
[hive-exec](https://mvnrepository.com/artifact/org.apache.hive/hive-exec)

cpovirk:
Oh, thanks, I should definitely have thought of guava-jdk5.

Which versions of hive-exec have this problem? I pulled hive-exec-0.9.0.jar, hive-exec-1.2.2-core.jar, hive-exec-2.3.9-core.jar, hive-exec-3.1.3-core.jar, hive-exec-4.0.0-beta-1.jar, and hive-exec-4.0.0-beta-1-fallbackauthorizer.jar, and only 4.0.0-beta-1.jar appeared to contain Guava classes, which thankfully were repackaged and thus not a problem (though I also see _other_ com.google classes in that jar that have _not_ been repackaged, all under com.google.protobuf or com.google.re2j).

Not that we can necessarily do much about hive-exec: Since it contains _additional_ classes beyond just Guava's, we can't say to replace its contents with Guava. At most, we might be able to say to replace Guava's contents with _it_, and I'm not sure that's likely to be a net win :(

pan3793:
@cpovirk for instance, hive-exec-2.3.7.jar

cpovirk:
Ah, thanks, I didn't pay close enough attention to that "core" suffix that I was looking at.

[...]

Issue #6217, opened 2022-10-16
Testlib: Add test that putIfAbsent replaces a null value 


ben-manes:
This was shown in a recent [Java Collections Puzzlers](https://www.youtube.com/watch?v=w6hhjg_gt_M). The JavaDoc for `Map.putIfAbsent` states the following.

> If the specified key is not already associated with a value (or is mapped to {@code null}) associates it with the given value and returns {@code null}, else returns the current value.

```java
@Test
public void putIfAbsent_nullExistingValue() {
  var map = new HashMap<Object, Object>();
  map.put("a", null);
  map.putIfAbsent("a", "b");
  assertThat(map).containsEntry("a", "b");
}
```

I found that multiple custom maps which permit null values due not honor this peculiarity. For example fastutils' adopted the testlib and  `Object2ObjectOpenHashMap` passes its suite, but it mistakenly does not replace the value as shown above.

kevinb9n:
Thanks Ben! Looks worth adding. If you or anyone else feels like making the PR we can merge it. If not, I hope we'd still get around to it sooner or later.

alvonellos:
@kevinb9n what would be the best way for me to get started contributing? This looks like a good first issue.

ben-manes:
Take a look at [MapPutIfAbsentTester](https://github.com/google/guava/blob/master/guava-testlib/src/com/google/common/collect/testing/testers/MapPutIfAbsentTester.java). These cases would be to allow null values so follow the pattern elsewhere for defining test matching constraints, e.g.

https://github.com/google/guava/blob/b337be608971b02a51f1a31a50a49c4755c6fd75/guava-testlib/src/com/google/common/collect/testing/testers/MapPutTester.java#L180-L184

It would also probably be good to review the default methods promoted from `ConcurrentMap` (e.g. replace) and the newer ones like computes and merge to see if the `ALLOWS_NULL_VALUES` cases are being covered. The tests should be run by [OpenJdk6MapTests](https://github.com/google/guava/blob/master/guava-testlib/test/com/google/common/collect/testing/OpenJdk6MapTests.java) which includes JDK maps that support or disallow null values, so verifying that yours works as expected means to see that your new tests intentionally fail and then go green when implemented. Otherwise, simply try to stay within the coding style of the surrounding code.

alvonellos:
Hey ben, I made the code change but I'm running into some issues getting guava to build on my system. I get an error: package sun.misc does not exist

any tips?

[...]

Issue #5756, opened 2021-10-18
unreleased resource : streams


hehaoNo1:
com/google/common/io/filebackedoutputstream     235line  

> FileOutputstream transfer=new FileOutputStream(temp)；

allanhaywood:
@hehaoNo1 Question for my edification. Transfer eventually gets assigned to out. Is your concern for when an exception may be hit in the try block?

PR #7974, opened 2025-09-07
Improve error messages for annotation methods on synthetic TypeVariables

lmcrean:
## Problem

Synthetic TypeVariables created by `TypeResolver` throw unhelpful `UnsupportedOperationException("methodName")` when annotation methods are called.

Addresses TODO b/147144588 in `TypeResolver.java:371-374`.

## Solution

Add helpful error messages explaining why annotations aren't supported and what to do instead.

**Changes:**
1. Convert TODO to NOTE documenting the intentional limitation
2. Add descriptive error for annotation methods with bug reference

## Impact

- Better developer experience through clear error messages
- No functional changes
- 2 files modified

## Testing

```bash
# Verify existing tests still pass
./mvnw test -f guava-tests/pom.xml -Dtest="TypeResolverTest,TypesTest"
```

All existing tests pass without modification, confirming backward compatibility.

## Example Error Message

Before:
```
UnsupportedOperationException: isAnnotationPresent
```

After:
```
UnsupportedOperationException: Annotation methods are not supported on synthetic
TypeVariables created during type resolution. The semantics of annotations on
resolved types with modified bounds are undefined. See b/147144588.
```

## Breaking Changes

**None.** This change only improves error messages. All existing behavior is preserved:
- Same exceptions thrown
- Same methods unsupported
- No new TypeVariable creation
- No API changes

### DIFF
--- a/guava/src/com/google/common/reflect/Types.java
+++ b/guava/src/com/google/common/reflect/Types.java
@@ -386,7 +386,9 @@ final class Types {
         if (methodName.equals("getAnnotatedBounds")
             || methodName.startsWith("getAnnotation")
             || methodName.startsWith("getDeclaredAnnotation")
-            || methodName.equals("isAnnotationPresent")) {
+            || methodName.equals("isAnnotationPresent")
+            || methodName.equals("getAnnotations")
+            || methodName.equals("getDeclaredAnnotations")) {
           throw new UnsupportedOperationException(
               "Annotation methods are not supported on synthetic TypeVariables created during type "
               + "resolution. The semantics of annotations on resolved types with modified bounds are "


lmcrean:
@cpovirk I took a shot at implementing the TypeVariable annotation support you identified in `TODO: b/147144588`. Does this approach align with what you had in mind? Happy to adjust if needed.

cpovirk:
Unfortunately, a big part of the problem is that I don't even know what behavior would make sense here :( I am a little fuzzy at the moment even on the exact circumstances that we end up using our own `TypeVariable` implementation in the first place (capture conversion? substitution (e.g., `resolveType`, `where`)?). Then there's the question of what should happen to annotations during such operations, especially given that the handling in javac itself has [changed over time](https://bugs.openjdk.org/browse/JDK-8323620). It's also possible that "proper" handling of annotations would mean returning our own `TypeVariable` implementation more often instead of the original JDK object (e.g., when substituting in `@Nullable E` for `E`?), and that could break existing users.

We say every so often that we should get together a list of issues for which it's reasonably like that we actually _would_ try to take a patch. The most recent such list that I have covers:
- https://github.com/google/guava/issues/6217
- https://github.com/google/guava/issues/6666
- https://github.com/google/guava/issues/5756
- https://github.com/google/guava/issues/5773

lmcrean:
@cpovirk thanks for the guidance, i have minimised the impact of this pr in light of what you said to focus on the error message. 

In the TypeResolver note I have used language to be more like "Until there's a clear specification for what annotations should mean on resolved TypeVariables..."

PR #7986, opened 2025-09-14
Fix resource leak in FileBackedOutputStream to prevent file handle exhaustion

### LINKED ISSUE #5756: unreleased resource : streams

lmcrean:
## Problem
FileBackedOutputStream leaks file handles when an IOException occurs during the transition from memory to file storage. The FileOutputStream created at line 242 is never closed if `write()` or `flush()` fails, leading to resource exhaustion.

## Solution
Use a success-flag pattern with nested try-finally to ensure the stream is closed only on failure, preserving it on success when it becomes the active output stream.

## Changes
- Added conditional cleanup logic in `FileBackedOutputStream.java`
- Stream is closed only when an exception occurs
- Original IOException is preserved and rethrown
- No changes to public API or behavior

## Testing
```bash
# Run specific tests
./mvnw test -Dtest=FileBackedOutputStreamTest -pl guava-tests

# Run with coverage for the affected class
./mvnw test -Dtest=FileBackedOutputStream* -pl guava-tests
```

All existing tests pass. Added regression tests to verify correct threshold-crossing behavior.

Fixes #5756

### DIFF
--- a/guava-tests/test/com/google/common/io/FileBackedOutputStreamTest.java
+++ b/guava-tests/test/com/google/common/io/FileBackedOutputStreamTest.java
@@ -171,4 +171,83 @@ public class FileBackedOutputStreamTest extends IoTestCase {
   private static boolean isWindows() {
     return OS_NAME.value().startsWith("Windows");
   }
+
+  /**
+   * Test that verifies the resource leak fix for Issue #5756.
+   * This test ensures that when switching from memory to file storage,
+   * the FileOutputStream is properly managed even if an exception occurs.
+   *
+   * Note: Direct testing of the IOException scenario during write/flush is
+   * challenging without mocking. This test verifies that normal operation
+   * still works correctly with the fix in place.
+   */
+  public void testThresholdCrossing_ResourceManagement() throws Exception {
+    // Test data that will cross the threshold
+    int threshold = 50;
+    byte[] beforeThreshold = newPreFilledByteArray(40);
+    byte[] afterThreshold = newPreFilledByteArray(30);
+
+    FileBackedOutputStream out = new FileBackedOutputStream(threshold);
+    ByteSource source = out.asByteSource();
+
+    // Write data that doesn't cross threshold
+    out.write(beforeThreshold);
+    assertNull(out.getFile());
+
+    // Write data that crosses threshold - this exercises the fixed code path
+    if (!JAVA_IO_TMPDIR.value().equals("/sdcard")) {
+      out.write(afterThreshold);
+      File file = out.getFile();
+      assertNotNull(file);
+      assertTrue(file.exists());
+
+      // Verify all data was written correctly
+      byte[] expected = new byte[70];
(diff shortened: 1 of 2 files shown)


guava/src/com/google/common/io/FileBackedOutputStream.java
```diff
@@ -238,12 +238,26 @@ private void update(int len) throws IOException {
...
+        try {
+          transfer.write(memory.getBuffer(), 0, memory.getCount());
+          transfer.flush();
+          // We've successfully transferred the data; switch to writing to file
+          out = transfer;
+          success = true;
+        } finally {
+          if (!success && transfer != null) {
+            try {
+              transfer.close();
```
cpovirk, line 255: Would it makes sense to perform the cleanup in the existing `catch (IOException e)` block? That should be the error case that we need to worry about. (And if we're worried about other cases, we could generalize it to also catch `RuntimeException`.) That would also let us attach any exception from `close` as a suppressed exception on `e`.

lmcrean: thanks @cpovirk I've put cleanup in the catch block now:

```java
        } catch (IOException e) {
        if (transfer != null) {
          try {
            transfer.close();
          } catch (IOException closeException) {
            e.addSuppressed(closeException);
          }
        }
        temp.delete();
        throw e;
      }
```

lmcrean: re. `RuntimeException` I get what you mean, may i suggest documenting as a new issue to keep the scope clean.

Once merged can add something like:

# FileBackedOutputStream: extend resource leak fix to RuntimeException

Current fix addresses IOException resource leaks during memory-to-file transition. Should extend to catch RuntimeException (SecurityException, system errors) for complete resource cleanup.

Simple change: `catch (IOException | RuntimeException e)`

RuntimeException scenarios are less common than IOException cases.

guava-tests/test/com/google/common/io/FileBackedOutputStreamTest.java
```diff
@@ -171,4 +171,83 @@ private static boolean isAndroid() {
...
     return OS_NAME.value().startsWith("Windows");
   }
+
+  /**
+   * Test that verifies the resource leak fix for Issue #5756.
+   * This test ensures that when switching from memory to file storage,
+   * the FileOutputStream is properly managed even if an exception occurs.
+   *
+   * Note: Direct testing of the IOException scenario during write/flush is
+   * challenging without mocking. This test verifies that normal operation
```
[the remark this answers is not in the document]
lmcrean: The new tests are regression tests that were added alongside the resource management fix to verify normal operation works correctly.

Specifically:
- `testThresholdCrossing_ResourceManagement()` tests normal threshold crossing behavior through the memory-to-file transition
- `testMultipleThresholdCrossings()` tests normal operation across repeated threshold crossings

PR #7988, opened 2025-09-14
Add tests demonstrating `Iterators.mergeSorted()` instability

lmcrean:
## Purpose

This PR adds test cases that **demonstrate the instability problem** in `Iterators.mergeSorted()` as requested by @kevinb9n in [issue #5773](https://github.com/google/guava/issues/5773#issuecomment-1581520295).

> The first PR that would be useful would have test cases that illustrate the problem -- this is unorthodox, but yes, I mean they actually verify that the problem *does* exist.

## What these tests show

The tests demonstrate that when `mergeSorted()` encounters equal elements from different iterators, it returns them in **unstable order** (dependent on heap implementation details) rather than maintaining a stable merge.

### Example from the issue:
- Input: `[B1, C1]` and `[A2, C2]` (comparing by letter only)
- Expected (stable): `[A2, B1, C1, C2]` - C1 before C2
- Actual (unstable): Sometimes `[A2, B1, C2, C1]` - C2 before C1

## Test failures

**These tests are expected to FAIL** with the current implementation. They verify that:
1. Equal elements don't maintain iterator order
2. The merge is unstable for equal elements

## Next steps

Once this PR is merged, a follow-up PR will:
1. Fix the stability issue in `mergeSorted()`
2. Verify these tests now pass with the fix (no changes needed to the tests themselves)

## Testing

```bash
# Compile the test module
./mvnw compile -pl guava-tests

# Run both new tests together (they will fail, demonstrating the problem)
./mvnw test -pl guava-tests -Dtest="IteratorsTest#testMergeSorted_demonstratesInstability*"

# Or run individually:
./mvnw test -pl guava-tests -Dtest=IteratorsTest#testMergeSorted_demonstratesInstability_issue5773Example
./mvnw test -pl guava-tests -Dtest=IteratorsTest#testMergeSorted_demonstratesInstability_allEqual
```

Related to #5773

### DIFF
--- a/guava-tests/test/com/google/common/collect/IteratorsTest.java
+++ b/guava-tests/test/com/google/common/collect/IteratorsTest.java
@@ -61,6 +61,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.ConcurrentModificationException;
 import java.util.Enumeration;
 import java.util.Iterator;
@@ -68,6 +69,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.NoSuchElementException;
+import java.util.Objects;
 import java.util.RandomAccess;
 import java.util.Set;
 import java.util.Vector;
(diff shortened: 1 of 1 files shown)


cpovirk:
Thanks, I'm able to reproduce the expected failure. I could probably merge a version of this PR that catches the `AssertionFailedError`, but that's arguably a little gross, and it's simpler to just merge https://github.com/google/guava/pull/7989 now that we have the desired confirmation from this PR and from https://github.com/google/guava/pull/7989 that https://github.com/google/guava/pull/7989 will solve the issue.

cpovirk:
(@kluever ended up submitting a passing version of the test for us in https://github.com/google/guava/pull/7992 :) Thanks.)

cpovirk:
(Apparently we had no tests _directly_ for `Iterators.mergeSorted` at all, but we at least had tests for its trivial wrapper, `Iterables.mergeSorted`.)

PR #7989, opened 2025-09-14
 Fix `Iterators.mergeSorted()` to preserve stability for equal elements

### LINKED ISSUE #5773: Make `Iterators.mergeSorted()` stable

lmcrean:
## Problem

`Iterators.mergeSorted()` returns equal elements in unstable order. When merging iterators containing elements that compare as equal, the order depends on heap implementation details rather than maintaining a predictable, stable merge.

## Solution

Track the insertion order of iterators and use it as a tiebreaker when elements compare as equal. Elements from earlier iterators will always appear before equal elements from later iterators.

## Changes

- Added `IndexedIterator` wrapper class to track iterator position
- Modified heap comparator to use insertion order as tiebreaker for equal elements
- Updated Javadoc to document the stable behavior guarantee
- Changed queue type from `Queue<PeekingIterator<T>>` to `Queue<IndexedIterator<T>>`

## Testing

All are passing on local machine

```bash
# Compile the changes
./mvnw clean compile -pl guava

# Run all Iterators tests (583 tests including stability tests from PR #1)
./mvnw test -pl guava-tests -Dtest=IteratorsTest

# Run stability tests specifically
./mvnw test -pl guava-tests -Dtest="IteratorsTest#testMergeSorted_demonstratesInstability*"
```

## Breaking Changes

None. This change only makes the ordering deterministic for equal elements, which was previously undefined.

Fixes #5773
Related to #7988 which is a reproduction of the issue (fails intentionally) and now passes in this PR

### DIFF
--- a/guava/src/com/google/common/collect/Iterators.java
+++ b/guava/src/com/google/common/collect/Iterators.java
@@ -1294,8 +1294,9 @@ public final class Iterators {
    * <p>Callers must ensure that the source {@code iterators} are in non-descending order as this
    * method does not sort its input.
    *
-   * <p>For any equivalent elements across all {@code iterators}, it is undefined which element is
-   * returned first.
+   * <p>For any equivalent elements across all {@code iterators}, elements are returned in the order
+   * of their source iterators. That is, if element A from iterator 1 and element B from iterator 2
+   * compare as equal, A will be returned before B if iterator 1 was passed before iterator 2.
    *
    * @since 11.0
    */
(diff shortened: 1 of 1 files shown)


cpovirk:
Nice, thanks. I'll test this and get it merged, too.

cpovirk:
This turns out to break one test inside Google. I'd figured it would just be a case in which someone had straightforwardly hard-coded the old (non-guaranteed) order, but it looks like there may be something more interesting going on. I still assume that the fault lies with that test (or the prod code that it tests!). It may just take me a little to get that sorted out.

PR #8006, opened 2025-09-18
Refactor: Fix critical state management bug and improve null safety

miladsade96:
### Changes:

1. Correcting state management bug
2. Handling null value properly

### DIFF
--- a/guava/src/com/google/common/base/AbstractIterator.java
+++ b/guava/src/com/google/common/base/AbstractIterator.java
@@ -14,14 +14,14 @@
 
 package com.google.common.base;
 
-import static com.google.common.base.NullnessCasts.uncheckedCastNullableTToT;
-import static com.google.common.base.Preconditions.checkState;
-
 import com.google.common.annotations.GwtCompatible;
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
+import org.jspecify.annotations.Nullable;
+
 import java.util.Iterator;
 import java.util.NoSuchElementException;
-import org.jspecify.annotations.Nullable;
+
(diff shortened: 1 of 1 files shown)


miladsade96:
I have signed the google cla, but still shows me as unsigned. Is there any way to force cla/google to reload or recheck the cla status?

cpovirk:
I clicked the link for rescanning it from https://github.com/google/guava/pull/8006/checks?check_run_id=50725139081, and now it's showing as passing.

If this PR fixes a bug, please add a test that fails without the PR and passes with it.

miladsade96:
@cpovirk 
I refactored the code but the behavior is the same as before.
Is it necessary to add more tests to the test class?

cpovirk:
Ah, OK, I was going off the PR description.

It looks to me like the PR introduces a bug: It no longer allows `computeNext()` to return a `null` to be returned to users. Instead, it treats any `null` as the end of the iteration. (The docs note: ["This class supports iterators that include null elements."](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/collect/AbstractIterator.html).) From what I've seen so far, I don't think we have tests for that case. We probably should.

chaoren:
> critical state management bug

Could you please describe what the bug is in the commit message and PR description?

> Handling null value properly

Please describe where in the existing code was null not properly handled.

> I refactored the code but the behavior is the same as before.

If the behavior is the same as before, then where is the bug?

There's a lot of shuffling around of existing code in the commit and it's not properly formatted, so it's difficult to tell what actually changed from just the diff.

PR #8007, opened 2025-09-19
Update javadoc with more direct FluentIterable.{index,uniqueIndex} equivalents.

### LINKED ISSUE #3119: FluentIterable.index equivalent
http://google.github.io/guava/releases/24.1-jre/api/docs/com/google/common/collect/FluentIterable.html#index-com.google.common.base.Function-
> **Stream equivalent:** `stream.collect(Collectors.groupingBy(keyFunction))` behaves similarly, but returns a mutable `Map<K, List<E>>` instead, and may not preserve the order of entries).

Since 21.0,  `stream.collect(ImmutableListMultimap.toImmutableListMultimap(keyFunction, v -> v))` may be used.

perceptron8:
Fixes #3119

I also tried to unify the documentation a little bit, but I'm happy to restore the "behaves similarly, but.." fragments in `android/guava/src/com/google/common/collect/FluentIterable.java` if they're necessary.

### DIFF
--- a/android/guava/src/com/google/common/collect/FluentIterable.java
+++ b/android/guava/src/com/google/common/collect/FluentIterable.java
@@ -700,9 +700,8 @@ public abstract class FluentIterable<E extends @Nullable Object> implements Iter
    * {@code valueFunction} will be applied to more than one instance of that key and, if it is,
    * which result will be mapped to that key in the returned map.
    *
-   * <p><b>{@code Stream} equivalent:</b> use {@code stream.collect(ImmutableMap.toImmutableMap(k ->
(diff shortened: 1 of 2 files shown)


cpovirk:
Thanks, this difference between the `jre` and `android` Javadoc is likely an oversight on my part from when I made the collectors available to Android users.

cpovirk:
I am proposing to carry over some more `FluentIterable` Javadoc from `guava-jre` to `guava-android` as part of merging this in https://github.com/google/guava/pull/8008, so if you would prefer not to have your name attached to that, let me know.

perceptron8:
I'm cool with that, thanks for asking. :-)

PR #8035, opened 2025-09-30
Fix foreach test assertion to handle HashSet's unordered iteration

annhchen89:
### What does this PR do?
This PR is fixing the ```testForEach()``` test failing caused by HashSet's unordered iteration. The original test compared the list directly, which assumed a consistent iteration order.

However, since the underlying collection was backed by a HashSet, the order of elements was not guaranteed. This caused the assertion to fail even though the correct elements were present.

###  Reproduce Test:
To reproduce the failing test, Nondex can be used.
```
mvn -pl guava-tests edu.illinois:nondex-maven-plugin:2.1.7:nondex -Dtest=com.google.common.collect.SetsFilterHashSetTest#testForEach -DnondexMode=FULL
```
The test fails in ```FULL``` mode because the two lists can be iterating over the same collection but with different internal orders, exposing the nondeterminism.

### The Fix:
This update changes the assertion to verify that all expected elements are present, without assuming any particular order.

### DIFF
--- a/guava-tests/test/com/google/common/collect/FilteredCollectionsTestUtil.java
+++ b/guava-tests/test/com/google/common/collect/FilteredCollectionsTestUtil.java
@@ -106,7 +106,7 @@ public final class FilteredCollectionsTestUtil {
               assertTrue("Unexpected element: " + i, EVEN.apply(i));
               foundElements.add(i);
             });
-        assertEquals(ImmutableList.copyOf(filtered), foundElements);
+        assertThat(foundElements).containsExactlyElementsIn(filtered);
       }
     }
   }


cpovirk:
I haven't been able to look at the PR contents because of the CLA, but if this is similar to https://github.com/google/guava/pull/7387, then I suspect we won't be interested for the reasons discussed there.

cpovirk:
With the CLA signed, I had a look. The changes here do look simpler than the ones in https://github.com/google/guava/pull/7387, so that's a plus. Still, I am comfortable relying on a given `HashSet` in a given VM to continue to iterate in the same order until it's modified. There's no guarantee, but if we see otherwise someday, we'll take care of it then.

PR #8056, opened 2025-10-21
Promote Traverser from beta to not a beta anymore

ZeroOne3010:
The Traverser has been around for eight years already, and your IDE still complains you're using a Beta API.


netdpb:
We are actually working on removing `@Beta` from all of the `common.graph` APIs. Please bear with us.

(Note that many IDEs misinterpret Guava's `@Beta` annotation. It doesn't mean these APIs are not ready for production use. It does mean you shouldn't use them _in open-source libraries_, since we have a lower bar for making incompatible changes. See [the README](https://github.com/google/guava?tab=readme-ov-file#important-warnings) for more details.)

PR #8111, opened 2025-12-08
Optimize getFileExtension by scanning string directly

yuanchaodu:
Avoid creating File object and temporary file name string.
Scan from the end of the path to locate the extension before any path separator.

### DIFF
--- a/guava/src/com/google/common/io/Files.java
+++ b/guava/src/com/google/common/io/Files.java
@@ -794,9 +794,16 @@ public final class Files {
    */
   public static String getFileExtension(String fullName) {
     checkNotNull(fullName);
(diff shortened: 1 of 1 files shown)


cgdecker:
I believe this is technically changing the behavior of this method. For example, `\` isn't a path separator in POSIX file systems so in such a file system the extension of a path like `foo/bar.a\b` with the current implementation would be `a\b`, while the new implementation would return an empty string. While it's certainly debatable whether that behavior is worth preserving (or if it even matters), the method's documentation specifies that it should work exactly the way it's current implemented. You could probably address this by using `File.separator`, but I'm not sure this is really something that's worth optimizing regardless.

PR #8112, opened 2025-12-08
Refactor padStart return statement for readability

yuanchaodu:
Refactor padStart return statement for readability

### DIFF
--- a/guava/src/com/google/common/base/Strings.java
+++ b/guava/src/com/google/common/base/Strings.java
@@ -98,8 +98,7 @@ public final class Strings {
     for (int i = string.length(); i < minLength; i++) {
       sb.append(padChar);
     }
-    sb.append(string);
-    return sb.toString();
+    return sb.append(string).toString();
   }
 
   /**


cgdecker:
Sorry, but this change is really not worth merging. You didn't update the android version of this file either, but even if you did the effort of getting PRs merged internally (we can't just merge these PRs directly on GitHub) is such that the bar for what changes are worth making needs to be higher than this.

cpovirk:
Now that you mention it, I'm trying to remember whether I've sometimes told people _not_ to bother with the `android` version, given that our process for merging PRs involves pulling them into the monorepo and running our merge script there, anyway. (Caveat: If a change required significantly different changes to the `android` and non-`android` flavors, then I could see modifying both.)

(I still agree with declining this, and I think I'd be in favor of declining it if it were proposed internally, too: I could see arguing that the current approach is better because it avoids chaining _consistently_ instead of mixing chaining and non-chaining. I think that reasonable people can disagree, but the difference, as you say, is small enough that I probably wouldn't merge a change in either direction.)

I might try to freshen up `CONTRIBUTING.md`, but I sort of expect to regret the amount of time I ultimately spend on it :)

cgdecker:
Yeah given that it doesn't help much with internal merges, not including `android` changes does seem reasonable.

PR #8143, opened 2026-01-01
Add MediaType.YAML_UTF_8 (see RFC 9512 for yaml.org)

vorburger:
Added YAML media type constant for YAML format.


vorburger:
https://github.com/google/guava/commit/a8e9533b499a446cccf7eff54e7775327fb119ec

PR #8154, opened 2026-01-09
Fix resource leak in FakeTickerTest (Issue #8152)

### LINKED ISSUE #8152: [Resource Leak] ExecutorService is never shut down in FakeTickerTest
### Guava Version

Latest Master

### Description

There is a resource leak in guava-testlib/.../testing/FakeTickerTest.java. In the runConcurrentTest method, an ExecutorService is created using newFixedThreadPool. However, there is no call to executorService.shutdown() or shutdownNow() at the end of the method. Since newFixedThreadPool creates non-daemon threads by default, these threads will remain alive in the JVM even after the test method finishes, potentially leading to thread exhaustion or OOM errors in large test suites.

### Example

```java
// Location: com.google.common.testing.FakeTickerTest.java

private void runConcurrentTest(int numberOfThreads, Callable<@Nullable Void> callable) throws Exception {
  ExecutorService executorService = newFixedThreadPool(numberOfThreads);
  // ... tasks submission ...
  doneLatch.await();
  
  // CRITICAL MISSING: executorService.shutdown();
  // Threads are leaked here.
}
```

### Expected Behavior

The ExecutorService should be explicitly shut down, preferably inside a try-finally block, to ensure all threads are released after the test completes.
```java
try {
    // ... test logic ...
} finally {
    executorService.shutdown();
}
```
### Actual Behavior

The ExecutorService remains active, and its worker threads leak, persisting until the JVM terminates.

### Packages

_No response_

### Platforms

Java 8

### Checklist

QiuYucheng2003:
**Description**
As discussed in Issue #8152, the `ExecutorService` in `FakeTickerTest` was created but never shut down, leading to potential thread leaks.

**Changes**
- Wrapped the test execution logic in a `try-finally` block.
- Added `executorService.shutdown()` in the `finally` block to ensure threads are released even if the test fails.

**Link to Issue**
Fixes #8152

### DIFF
--- a/guava-testlib/test/com/google/common/testing/FakeTickerTest.java
+++ b/guava-testlib/test/com/google/common/testing/FakeTickerTest.java
@@ -164,8 +164,7 @@ public class FakeTickerTest extends TestCase {
   @GwtIncompatible // concurrency
   private void runConcurrentTest(int numberOfThreads, Callable<@Nullable Void> callable)
           throws Exception {
-    ExecutorService executorService = newFixedThreadPool(numberOfThreads);
-    try { //   1.  <---try
+    try (ExecutorService executorService = newFixedThreadPool(numberOfThreads)) {
       CountDownLatch startLatch = new CountDownLatch(numberOfThreads);
       CountDownLatch doneLatch = new CountDownLatch(numberOfThreads);
       for (int i = numberOfThreads; i > 0; i--) {
@@ -184,8 +183,6 @@ public class FakeTickerTest extends TestCase {
                         });
       }
       doneLatch.await();
-    } finally {     // 2.  <---finally
-      executorService.shutdown();
     }
   }
 }


guava-testlib/test/com/google/common/testing/FakeTickerTest.java
```diff
@@ -163,25 +163,29 @@ public void testConcurrentAutoIncrementStep() throws Exception {
...
-                @Override
-                public @Nullable Void call() throws Exception {
-                  startLatch.countDown();
-                  startLatch.await();
-                  callable.call();
-                  doneLatch.countDown();
-                  return null;
-                }
-              });
+    try { //   1.  <---try
```
netdpb: `ExecutorService` implements `AutoCloseable`, so you could just use a try-with-resources.

PR #8173, opened 2026-01-26
testlib: Add concurrent collection spliterator tests via derived suite builders

### LINKED ISSUE #7855: testlib: add tests for Map.entrySet().stream()

nickita-khylkouski:
## Summary

Adds testlib support for verifying `Spliterator` characteristics on `ConcurrentMap` views.

**Architecture (per @chaoren's feedback):**
- Add `ConcurrentCollectionSpliteratorTester` - tests that a collection's spliterator has the `CONCURRENT` characteristic
- Add `ConcurrentSetTestSuiteBuilder` and `ConcurrentCollectionTestSuiteBuilder` - reusable builders that include the concurrent tester
- Override `createDerivedEntrySetSuite`, `createDerivedKeySetSuite`, `createDerivedValueCollectionSuite` in `ConcurrentMapTestSuiteBuilder` to use the concurrent builders
- Add spec invariant checks to `CollectionSpliteratorTester` (applies to ALL collections):
  - `CONCURRENT` and `SIZED` must not coexist
  - `CONCURRENT` and `IMMUTABLE` must not coexist
  - `SUBSIZED` requires `SIZED`
  - `SORTED` requires `ORDERED`

## Motivation

Per [issue #7855](https://github.com/google/guava/issues/7855) and maintainer guidance from @chaoren:
- `ConcurrentMap` implementations should return spliterators with the `CONCURRENT` characteristic
- The Java specification mandates several characteristic invariants

Currently, some Guava `ConcurrentMap` implementations (e.g., `LocalCache`, `MapMakerInternalMap`) return spliterators that:
- Do NOT have `CONCURRENT` (should have it)
- DO have `SIZED` (incompatible with `CONCURRENT` per spec)

This PR adds test infrastructure to document the expected behavior. A follow-up PR can fix the implementations.

## Testing

- Classes compile successfully
- Tests follow the existing `ConcurrentMap*Tester` pattern (plain `@GwtCompatible`, `@NullMarked`, no reflection methods)
- Verified JDK's `ConcurrentHashMap` correctly returns `CONCURRENT=true, SIZED=false`

Fixes #7855

### DIFF
--- a/guava-testlib/src/com/google/common/collect/testing/ConcurrentCollectionTestSuiteBuilder.java
+++ b/guava-testlib/src/com/google/common/collect/testing/ConcurrentCollectionTestSuiteBuilder.java
@@ -29,7 +29,7 @@ import java.util.List;
  * <p>This builder adds {@link ConcurrentCollectionSpliteratorTester} to verify that the
  * collection's spliterator has the {@link java.util.Spliterator#CONCURRENT} characteristic.
  *
- * @author Guava Authors
+ * @author Nickita Khylkouski
  */
 @GwtIncompatible
 public class ConcurrentCollectionTestSuiteBuilder<E> extends CollectionTestSuiteBuilder<E> {
--- a/guava-testlib/src/com/google/common/collect/testing/ConcurrentSetTestSuiteBuilder.java
+++ b/guava-testlib/src/com/google/common/collect/testing/ConcurrentSetTestSuiteBuilder.java
@@ -29,7 +29,7 @@ import java.util.List;
  * <p>This builder adds {@link ConcurrentCollectionSpliteratorTester} to verify that the set's
  * spliterator has the {@link java.util.Spliterator#CONCURRENT} characteristic.
  *
- * @author Guava Authors
+ * @author Nickita Khylkouski
  */
 @GwtIncompatible
 public class ConcurrentSetTestSuiteBuilder<E> extends SetTestSuiteBuilder<E> {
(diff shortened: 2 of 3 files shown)


guava-testlib/src/com/google/common/collect/testing/testers/ConcurrentMapSpliteratorTester.java
```diff
@@ -0,0 +1,166 @@
...
+
+  /**
+   * Tests that the spliterator returned by {@code values().spliterator()} has the {@code
+   * CONCURRENT} characteristic.
+   */
+  public void testValuesSpliteratorHasConcurrent() {
+    Spliterator<?> spliterator = getMap().values().spliterator();
+    assertTrue(
+        "values().spliterator() should have CONCURRENT characteristic",
+        spliterator.hasCharacteristics(Spliterator.CONCURRENT));
```
chaoren: I think we can add these to `CollectionSpliteratorTester` instead.

Per https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Spliterator.html#CONCURRENT,
`CONCURRENT` is incompatible with both `SIZED` and `IMMUTABLE`, and that should apply to *all* Spliterators, not just concurrent ones.

Also, https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Spliterator.html#SUBSIZED says

> A Spliterator that does not report `SIZED` as required by `SUBSIZED` is inconsistent

We can check that as well.

nickita-khylkouski:
Done - moved the spec invariant checks to CollectionSpliteratorTester and set up the derived suite overrides so the concurrent builders get used for entrySet/keySet/values. Rebased on latest master too. Let me know if anything else needs adjusting.

PR #8174, opened 2026-01-26
collect: Return null from Spliterator.getComparator() for natural ordering

### LINKED ISSUE #6187: Spliterators for sorted collections cause inefficient stream operations when natural ordering is used

nickita-khylkouski:
## Summary
- Return `null` from `Spliterator.getComparator()` when comparator is natural ordering
- Check for both `Ordering.natural()` and `Comparator.naturalOrder()` (both singletons)
- Apply fix to `ImmutableSortedSet.spliterator()` and `CollectSpliterators.indexed()`
- Add unit tests for the new behavior

## Motivation
Per the [Spliterator.getComparator() contract](https://docs.oracle.com/javase/8/docs/api/java/util/Spliterator.html#getComparator--), `null` should be returned to indicate natural ordering. The JDK stream implementation [clears the SORTED flag](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/stream/StreamOpFlag.java#L751-L755) when `getComparator()` returns non-null, which [prevents the sorted() optimization](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/stream/SortedOps.java#L136-L139).

This causes unnecessary sorting when calling `stream().sorted()` on naturally-ordered `ImmutableSortedSet`:

```java
ImmutableSortedSet<Integer> sortedSet = ImmutableSortedSet.of(1, 2, 3, 4);
// sorted() should be a no-op, but wasn't because getComparator() returned Ordering.natural()
sortedSet.stream().sorted().collect(Collectors.toList());
```

Fixes #6187

## Testing
Added 3 unit tests to `ImmutableSortedSetTest`:
- `testSpliteratorComparator_naturalOrdering` - verifies `getComparator()` returns null for natural ordering
- `testSpliteratorComparator_customComparator` - verifies custom comparator is returned
- `testAsListSpliteratorComparator_naturalOrdering` - verifies `asList().spliterator()` also returns null

Also verified locally that stream sorted() optimization now works correctly.

### DIFF
--- a/guava-tests/test/com/google/common/collect/ImmutableSortedSetTest.java
+++ b/guava-tests/test/com/google/common/collect/ImmutableSortedSetTest.java
@@ -55,6 +55,7 @@ import java.util.Iterator;
 import java.util.NoSuchElementException;
 import java.util.Set;
 import java.util.SortedSet;
+import java.util.Spliterator;
 import java.util.TreeSet;
 import java.util.function.BiPredicate;
 import java.util.stream.Collector;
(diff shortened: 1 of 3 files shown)


nickita-khylkouski:
Superseded by PR #8182 which covers this as part of comprehensive spliterator audit for issue #8165

PR #8179, opened 2026-01-28
testlib: Use SUPPORTS_REMOVE in CollectionRemoveIfTester

### LINKED ISSUE #6076: [Bug] GuavaTestLib RemoveIf Flag Inconsistencies
I have been working on a Test Library that basically adds support for Primitive Variants of Guavas Test Lib.
And some of my implementations require to change how iterators are checked and I noticed when disabling that the removeIf tests get disabled too.

While looking through that class I noticed that tests check if "SUPPORTS_ITERATOR_REMOVE" flag is set, but if you have the "SUPPORTS_REMOVE" flag missing it tries to test if removeIf throws exceptions.

There is clearly some inconsistencies in there.
https://github.com/google/guava/blob/master/guava-testlib/src/com/google/common/collect/testing/testers/CollectionRemoveIfTester.java

My suggested fix is make them all "SUPPORTS_REMOVE", since the iterator has nothing to do with the removeIf function.
While that there could be a case made for it, though at that point a dedicated flag for lambdas/functions could be made.

Would be nice if this could be addressed, because I kinda either need to copy your class just to fix it or not test removeIf.

nickita-khylkouski:
## Summary
- Changed `CollectionRemoveIfTester` to use `SUPPORTS_REMOVE` instead of `SUPPORTS_ITERATOR_REMOVE` in all test method annotations
- Removed unused import for `SUPPORTS_ITERATOR_REMOVE`

## Motivation
The `removeIf()` method is a `Collection` method that can be overridden independently of iterator support. The default implementation uses `Iterator.remove()`, but collections can provide their own optimized implementation that doesn't rely on iterator removal.

Therefore, `CollectionRemoveIfTester` should require `SUPPORTS_REMOVE` (the general collection removal capability) rather than `SUPPORTS_ITERATOR_REMOVE` (the specific iterator removal capability). This aligns with how other collection testers work (e.g., `CollectionRemoveTester`, `ListRemoveTester`, `CollectionClearTester`).

This addresses issue #6076 where the incorrect flag was causing tests to be skipped for collections that support `removeIf()` but don't support iterator removal.

## Testing
- Verified that `SUPPORTS_ITERATOR_REMOVE` is no longer present in any `@CollectionFeature.Require` annotations
- Verified that `SUPPORTS_REMOVE` is present in all expected locations (4 test methods)
- Class compiles successfully
- Changes are minimal and focused (5 insertions, 6 deletions)

Fixes #6076

### DIFF
--- a/guava-testlib/src/com/google/common/collect/testing/testers/CollectionRemoveIfTester.java
+++ b/guava-testlib/src/com/google/common/collect/testing/testers/CollectionRemoveIfTester.java
@@ -17,7 +17,6 @@
 package com.google.common.collect.testing.testers;
 
 import static com.google.common.collect.testing.features.CollectionFeature.FAILS_FAST_ON_CONCURRENT_MODIFICATION;
-import static com.google.common.collect.testing.features.CollectionFeature.SUPPORTS_ITERATOR_REMOVE;
 import static com.google.common.collect.testing.features.CollectionFeature.SUPPORTS_REMOVE;
 import static com.google.common.collect.testing.features.CollectionSize.SEVERAL;
 import static com.google.common.collect.testing.features.CollectionSize.ZERO;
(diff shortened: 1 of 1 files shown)


nickita-khylkouski:
Closing this PR - the change from `SUPPORTS_ITERATOR_REMOVE` to `SUPPORTS_REMOVE` is incorrect.

`Collection.removeIf()` default implementation calls `Iterator.remove()`, so the correct feature flag is `SUPPORTS_ITERATOR_REMOVE`. Changing to `SUPPORTS_REMOVE` causes 1753 test errors because many Guava collections support `Collection.remove(Object)` but not `Iterator.remove()` (e.g., filtered views using `UnmodifiableIterator`).

PR #8180, opened 2026-01-28
Semgrep Rule Fix

Akhil-gi:
We have used our AI-Guardian product to identify and remediate a Semgrep rule violation

  ```
  Rule ID: unsafe-reflection
    Rule Message: If an attacker can supply values that the application then uses to determine which class to instantiate or which method to invoke, the potential exists for the attacker to create control flow paths through the application that were not intended by the application developers. This attack vector may allow the attacker to bypass authentication or access control checks or otherwise cause the application to behave in an unexpected manner.
    File Path: /tools/scanResult/unzipped-271687943/android/guava-tests/benchmark/com/google/common/base/EnumsBenchmark.java
    Line: 49
```


eamonnmcmanus:
Please don't waste our time with things like this. Even the most cursory examination shows that this code is in a benchmark.

PR #8183, opened 2026-01-28
BaseEncoding: Make encodingStream().close() idempotent

### LINKED ISSUE #5284: BaseEncoding.encodingStream().close() is non-idempotent

nickita-khylkouski:
## Summary
- Add `boolean closed` field to the `OutputStream` returned by `encodingStream()`
- Make `write()`, `flush()`, and `close()` `synchronized` per @cpovirk's guidance
- `close()` returns early if already closed (idempotent)
- `write()`/`flush()` throw `IOException` after close, matching `AppendableWriter` and `CharSequenceReader` patterns

## Motivation
`BaseEncoding.encodingStream().close()` violates the `Closeable.close()` contract which states: "If the stream is already closed then invoking this method has no effect."

Calling `close()` twice writes additional encoded data to the underlying Writer:
```java
StringWriter w = new StringWriter();
OutputStream out = BaseEncoding.base64().encodingStream(w);
out.write(0);
out.close();
System.out.println(w); // "AA==" (correct)
out.close();
System.out.println(w); // "AA==A===" (wrong - extra data)
```

This is problematic because double-close is common due to try-with-resources blocks and wrapping streams that close their underlying streams.

Fixes #5284

## Testing
- 3 new tests in `BaseEncodingTest`:
  - `testEncodingStreamCloseIsIdempotent` — verifies close() twice doesn't change output
  - `testEncodingStreamWriteAfterClose` — verifies write() throws IOException after close
  - `testEncodingStreamFlushAfterClose` — verifies flush() throws IOException after close
- All 43 tests in `BaseEncodingTest` pass (`./mvnw test -pl guava-tests -Dtest=BaseEncodingTest`)
- Added `IOException` to `ReflectionFreeAssertThrows` to support the new test assertions

### DIFF
--- a/guava/src/com/google/common/io/BaseEncoding.java
+++ b/guava/src/com/google/common/io/BaseEncoding.java
@@ -645,9 +645,13 @@ public abstract class BaseEncoding {
         int bitBuffer = 0;
         int bitBufferLength = 0;
         int writtenChars = 0;
+        boolean closed = false;
 
         @Override
-        public void write(int b) throws IOException {
+        public synchronized void write(int b) throws IOException {
+          if (closed) {
+            throw new IOException("Stream is closed");
+          }
           bitBuffer <<= 8;
           bitBuffer |= b & 0xFF;
           bitBufferLength += 8;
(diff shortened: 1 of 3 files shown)


guava/src/com/google/common/io/BaseEncoding.java
```diff
@@ -645,9 +645,13 @@ public OutputStream encodingStream(Writer out) {
         int bitBuffer = 0;
         int bitBufferLength = 0;
         int writtenChars = 0;
+        boolean closed = false;
 
         @Override
-        public void write(int b) throws IOException {
+        public synchronized void write(int b) throws IOException {
```
softpapers: "Synchronization is required for reliable communication between threads as well as for mutual exclusion. " - Item 78 of Effective Java Programming by Joshua Blosch.

What justifies the need for the write of an abstract class to be synchronised?

PR #8189, opened 2026-01-30
eventbus: Catch StackOverflowError in Subscriber.dispatchEvent()

### LINKED ISSUE #7728: Error being undetected by the SubscriberExceptionHandler
### API(s)

```java
`com.google.common.eventbus.Subscriber#dispatchEvent`
```

### How do you want it to be improved?

```java
  final void dispatchEvent(final Object event) {
    executor.execute(
        new Runnable() {
          @Override
          public void run() {
            try {
              invokeSubscriberMethod(event);
            } catch (InvocationTargetException e) {
              bus.handleSubscriberException(e.getCause(), context(event));
            } catch(Throwable e) {
              bus.handleSubscriberException(e, context(event));
            }
          }
        });
  }
```

### Why do we need it to be improved?

The `Subscriber#dispatchEvent` did not catch `Throwable`, resulting in the `Error` being undetected by the `SubscriberExceptionHandler`

### Example

```java
-
```

### Current Behavior

-

### Desired Behavior

`Error` can be detected by the `SubscriberExceptionHandler`

### Concrete Use Cases

-

### Checklist

- [x] I have read and understood the [contribution guidelines](https://github.com/google/guava/wiki/HowToContribute#feature-requests).

- [x] I have read and understood [Guava's philosophy](https://github.com/google/guava/wiki/PhilosophyExplained), and I strongly believe that this proposal aligns with it.

nickita-khylkouski:
## Summary

When a subscriber method throws `StackOverflowError` directly (not wrapped in `InvocationTargetException`), the error was not caught by the `SubscriberExceptionHandler`, making debugging difficult.

## Changes

- **Fix:** Added a catch block for `StackOverflowError` in `Subscriber.dispatchEvent()`, allowing the handler to log the error for debugging purposes.
- **Test:** Added `testSubscriberThrowsStackOverflowError()` to verify that `StackOverflowError` is properly caught and handled.

## Motivation

Reported by @Jiabao-Sun in #7728:

> When encountering ERRORs such as `StackOverflowError`, losing the problem stack trace can make troubleshooting extremely difficult.

The current implementation only catches `InvocationTargetException`, so if a subscriber method throws `StackOverflowError` directly, it propagates up without being logged by the exception handler.

## Design Decision

This change specifically catches `StackOverflowError` (not all `Throwable`) to follow Guava's existing pattern of handling `StackOverflowError` specially. See `Platform.rethrowIfErrorOtherThanStackOverflow()` which treats `StackOverflowError` differently from other `Error` types.

Other `Error` types (like `OutOfMemoryError`, `LinkageError`) are intentionally not caught, as they typically indicate more serious JVM-level issues that should propagate.

## Testing

The new test:
1. Creates an EventBus with a RecordingSubscriberExceptionHandler
2. Registers a subscriber that throws StackOverflowError
3. Posts an event to trigger the subscriber
4. Verifies the error was caught by the handler with correct context

Fixes #7728

### DIFF
--- a/guava-tests/test/com/google/common/eventbus/EventBusTest.java
+++ b/guava-tests/test/com/google/common/eventbus/EventBusTest.java
@@ -163,6 +163,31 @@ public class EventBusTest extends TestCase {
     eventBus.post(EVENT);
   }
 
+  public void testSubscriberThrowsStackOverflowError() throws Exception {
+    RecordingSubscriberExceptionHandler handler = new RecordingSubscriberExceptionHandler();
+    EventBus eventBus = new EventBus(handler);
+    StackOverflowError error = new StackOverflowError("deep recursion");
+    Object subscriber =
+        new Object() {
+          @Subscribe
+          public void throwStackOverflowOn(String message) {
+            throw error;
+          }
+        };
+    eventBus.register(subscriber);
+    eventBus.post(EVENT);
+
+    assertEquals("StackOverflowError should be caught by handler.", error, handler.exception);
+    assertEquals("EventBus should be available.", eventBus, handler.context.getEventBus());
+    assertEquals("Event should be available.", EVENT, handler.context.getEvent());
+    assertEquals("Subscriber should be available.", subscriber, handler.context.getSubscriber());
+    assertEquals(
+        "Method should be available.",
+        subscriber.getClass().getMethod("throwStackOverflowOn", String.class),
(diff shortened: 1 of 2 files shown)


nickita-khylkouski:
Withdrawing this PR based on code review. The approach violates Guava's error handling philosophy by swallowing StackOverflowError. Errors should be allowed to propagate as they indicate system-level problems that should crash the application before entering a corrupted state.

PR #8198, opened 2026-02-04
Make BloomFilter.bitSize() public

### LINKED ISSUE #6866: BloomFilter: add method to get the binary size in public API
### 1. What are you trying to do?

I am using Guava's bloom filters as part of a persistent file format, i.e. the raw byte array of the bloom filter lives somewhere in the file. It would be beneficial to the efficiency to know the length of the byte array produced by the bloom filter beforehand (i.e. without actually serializing it).

There is already a method called `bitSize` in the BloomFilter, but unfortunately it is not public. The method also doesn't include the two bytes from the strategy and the number of hash functions, as well as the integer for the length of the `bits.data` array.

### 2. What's the best code you can write to accomplish that without the new feature?

```java
public int getByteSizeOf(BloomFilter<*> bloomFilter) {
    return serialize(bloomFilter).length;
}

public byte[] serialize(BloomFilter<*> bloomFilter){
    try(var baos = new ByteArrayOutputStream()) {
        bloomFilter.writeTo(baos);
        return baos.toByteArray();
    }
}
```

The method `getSizeOf` is very inefficient because it actually serializes the bloom filter to get its size. It would be nice if we could do it without the serialization.

### 3. What would that same code look like if we added your feature?

BloomFilter<*> bloom = ...;
var size = bloom.getSizeInBytes();

### (Optional) What would the method signatures for your feature look like?

```java
public class BloomFilter<T> {

    public int getSizeInBytes();

}
```

### Concrete Use Cases

Serialization of the bloom filter as a building block for more complex formats.

### Packages

com.google.common.hash

### Checklist

- [X] I agree to follow the [code of conduct](https://github.com/google/.github/blob/master/CODE_OF_CONDUCT.md).

- [X] I have read and understood the [contribution guidelines](https://github.com/google/guava/wiki/HowToContribute#feature-requests).

- [X] I have read and understood [Guava's philosophy](https://github.com/google/guava/wiki/PhilosophyExplained), and I strongly believe that this proposal aligns with it.

- [X] I have visited the [idea graveyard](https://github.com/google/guava/wiki/IdeaGraveyard), and did not see anything similar to this idea.

nickita-khylkouski:
## Summary

This PR makes the existing `bitSize()` method in `BloomFilter` public. Previously it was package-private with `@VisibleForTesting`.

### Motivation

Users who embed Bloom filters in custom file formats need to know the serialization size before writing to pre-allocate space (e.g., for memory-mapped files or fixed-size records). Currently the only way to determine the size is to actually serialize the filter, which is inefficient.

### Changes

- Changed `bitSize()` visibility from package-private to `public`
- Removed `@VisibleForTesting` annotation
- Added comprehensive Javadoc explaining:
  - The method's purpose
  - How to calculate serialization size: `bitSize() / 8 + 6` bytes
  - The 6-byte header breakdown (1 byte strategy, 1 byte hash functions, 4 bytes array length)
- Added `@since 35.0` annotation
- Applied changes to both JRE and Android flavors

### Tests

Added `testBitSizeMatchesSerializationSize()` test that verifies `bitSize()` correctly predicts the `writeTo()` output size across various configurations.

## Test plan

- [x] New test `testBitSizeMatchesSerializationSize()` validates the size calculation formula
- [x] Existing `testBitSize()` test continues to pass (now testing public API)
- [ ] Run full Guava test suite

Fixes #6866

### DIFF
--- a/android/guava-tests/test/com/google/common/hash/BloomFilterTest.java
+++ b/android/guava-tests/test/com/google/common/hash/BloomFilterTest.java
@@ -353,6 +353,47 @@ public class BloomFilterTest extends TestCase {
     }
   }
 
+  /**
+   * Tests that bitSize() can be used to predict the serialization size produced by writeTo().
+   *
+   * <p>The serialization format consists of a 6-byte header (1 byte strategy, 1 byte hash
+   * functions, 4 bytes array length) followed by the bit array data (bitSize / 8 bytes).
+   */
+  public void testBitSizeMatchesSerializationSize() throws Exception {
+    int[] expectedInsertionValues = {1, 10, 100, 1000, 10000};
+    double[] fppValues = {0.01, 0.03, 0.1};
+
+    for (int expectedInsertions : expectedInsertionValues) {
+      for (double fpp : fppValues) {
+        BloomFilter<String> bf =
+            BloomFilter.create(Funnels.unencodedCharsFunnel(), expectedInsertions, fpp);
+
+        // Add some elements
+        for (int i = 0; i < expectedInsertions / 2; i++) {
+          bf.put("element" + i);
+        }
+
+        // Calculate expected size based on bitSize()
+        // Header: 1 byte (strategy) + 1 byte (hash functions) + 4 bytes (array length) = 6 bytes
+        // Data: bitSize / 8 bytes
+        long predictedSize = bf.bitSize() / 8 + 6;
+
+        // Serialize and measure actual size
+        ByteArrayOutputStream out = new ByteArrayOutputStream();
+        bf.writeTo(out);
+        int actualSize = out.size();
+
+        assertEquals(
+            "Serialization size mismatch for expectedInsertions="
+                + expectedInsertions
+                + " fpp="
+                + fpp,
+            predictedSize,
+            actualSize);
+      }
+    }
+  }
+
   public void testApproximateElementCount() {
     int numInsertions = 1000;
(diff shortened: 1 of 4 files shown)


kluever:
I wonder if it'd be more convenient for users if we exposed an API that tells you how many _bytes_ (instead of _bits_) the underlying `AtomicLongArray` is? We'd want to figure out a good name for it (I'm not convinced `byteSize()` is very good :-)

jrtom:
_waves_

I'm not convinced that `byteSize()` is the best name, but at least it's easily digestible.  ;)

More ~serially~ seriously: I do agree that providing the size in bytes is better than the size in bits, at least for the proposed purpose.  That said: if the sole purpose of  providing it is to inform serialization users of the size of the object to be serialized, maybe the reported size should both (a) be in bytes and (b) include the header as well, and be named something like `serializationSize()`.

nickita-khylkouski:
Great suggestions! I agree — returning bytes with the header included is a cleaner API. Users shouldn't need to do the `bitSize() / 8 + 6` math themselves.

For naming, a few options:
- `serializationSize()` — @jrtom's suggestion, clean and direct
- `serializedSizeInBytes()` — explicit about units, but verbose
- `writtenSize()` — ties to `writeTo()`, but less clear in isolation

I lean toward `serializationSize()` for simplicity. Happy to update the PR once you settle on a name — or if you'd prefer a different approach entirely.

PR #8278, opened 2026-03-16
fix: Omit sources from generated Javadoc

### LINKED ISSUE #7597: Omit sources from generated Javadoc
When we build Javadoc, we [set `linksource`](https://github.com/google/guava/blob/ff7c071b0aeff0306451f74aa69b2c2c5803f932/pom.xml#L324). That causes the Javadoc tool to generate links from text like `public interface Multimap` in https://guava.dev/Multimap. Those links go to pages like https://guava.dev/releases/snapshot-jre/api/docs/src-html/com/google/common/collect/Multimap.html#line-164.

I somewhat suspect that no one cares about these pages: If they want source, they'll look in GitHub or in the source jar for a release or snapshot.

Additionally, these pages force the generated Javadoc to change even for _implementation_ changes (example [output changes](https://guava.dev/releases/snapshot-jre/api/docs/src-html/com/google/common/collect/Multimap.html#line-164) from [input changes](https://github.com/google/guava/commits/master/guava/src/com/google/common/io/Closeables.java)).

And _any_ additional page presumably slows down downloading history, switching branches, and deploying the guava.dev. site.

I'm not sure I'll even bother to remove `linksource`, but I'm thinking about it, so let me know if that would cause a problem for you. If I do try it, and if it goes well, I might even consider removing some of the existing sources....

JiwaniZakir:
Fixes #7597

The maven-javadoc-plugin was embedding source code links into generated Javadoc output. The `linksource` parameter defaults to `true` in some plugin versions, causing source files to be included in the Javadoc artifact. Sets `linksource` to `false` in the `maven-javadoc-plugin` configuration block in `pom.xml` to explicitly disable this behavior. Verified by inspecting the resulting Javadoc output to confirm source files are no longer embedded.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -328,6 +328,7 @@
               <additionalOption>-XDignore.symbol.file</additionalOption>
               <additionalOption>-Xdoclint:-html</additionalOption>
             </additionalOptions>
+            <linksource>false</linksource>
             <source>8</source>
           </configuration>
           <executions>


cpovirk:
This PR does not change any behavior.

JiwaniZakir:
That's the intent - this is a pure refactor/cleanup with no behavioral changes. something looks off.

PR #8292, opened 2026-03-23
Prevent newline (`\n`) in MediaType.parse() quoted parameters (header safety)

artahir-dev:
This PR addresses an issue in `MediaType.parse()` where bare line feed (`\n`) characters were permitted within quoted parameter values (e.g., `charset="utf-8\n"`).

While carriage return (`\r`) was already excluded, line feed (`\n`) was not covered by `QUOTED_TEXT_MATCHER`. In scenarios where parsed `MediaType` values are incorporated into HTTP headers without proper sanitization, this behavior could contribute to HTTP header injection risks.

This change updates `QUOTED_TEXT_MATCHER` to also reject `\n`, aligning with safe header construction practices.

Tests have been extended to assert that inputs containing bare `\n` now result in `IllegalArgumentException`.

### DIFF
--- a/guava-tests/test/com/google/common/net/MediaTypeTest.java
+++ b/guava-tests/test/com/google/common/net/MediaTypeTest.java
@@ -379,6 +379,7 @@ public class MediaTypeTest extends TestCase {
     assertThrows(IllegalArgumentException.class, () -> MediaType.parse("text/plain; a=1; b"));
     assertThrows(IllegalArgumentException.class, () -> MediaType.parse("text/plain; a=1; b="));
     assertThrows(IllegalArgumentException.class, () -> MediaType.parse("text/plain; a=\u2025"));
+    assertThrows(IllegalArgumentException.class, () -> MediaType.parse("text/plain; a=\"\n\""));
   }
 
   // https://github.com/google/guava/issues/6663
--- a/guava/src/com/google/common/net/MediaType.java
+++ b/guava/src/com/google/common/net/MediaType.java
@@ -85,7 +85,7 @@ public final class MediaType {
           .and(CharMatcher.isNot(' '))
           .and(CharMatcher.noneOf("()<>@,;:\\\"/[]?="));
 
-  private static final CharMatcher QUOTED_TEXT_MATCHER = ascii().and(CharMatcher.noneOf("\"\\\r"));
+  private static final CharMatcher QUOTED_TEXT_MATCHER = ascii().and(CharMatcher.noneOf("\"\\\r\n"));
 
   /*
    * This matches the same characters as linear-white-space from RFC 822, but we make no effort to


chaoren:
From https://datatracker.ietf.org/doc/html/rfc822

```
     quoted-string = <"> *(qtext/quoted-pair) <">; Regular qtext or
                                                 ;   quoted chars.
     
     qtext       =  <any CHAR excepting <">,     ; => may be folded
                     "\" & CR, and including
                     linear-white-space>
```

Only excluding `"`, `\`, and `\r` is correct according to the spec for `qtext`.

artahir-dev:
Thank you for the quick review and for pointing out the RFC 822 specification. You're absolutely right that the `qtext` grammar only explicitly excludes `"`, `\`, and CR (`\r`), so LF (`\n`) is technically permitted in quoted strings per that spec.

However, the security concern I was trying to address is that when a `MediaType` instance later gets serialized into an HTTP header (e.g., `Content-Type`), a bare LF in a quoted parameter could allow header injection attacks, just like CR does. Since we already forbid CR for this reason, allowing LF creates an inconsistent safety boundary: both characters are dangerous in HTTP headers, but only one is blocked.

I understand that keeping `MediaType.parse()` strictly spec‑compliant is valuable, and I agree that modifying the parser itself might not be the right layer for this hardening. Would you be open to an alternative approach, such as:

1. Adding a validation method (e.g., `MediaType.isSafeForHttpHeader()`) that rejects both CR and LF, and documenting that when constructing HTTP headers from untrusted input, callers should use that method?
2. Or performing the LF check at the point where a `MediaType` is converted to a header string (e.g., in `HttpHeaders` or similar), rather than inside the parser?

I’d be happy to rework the PR to follow the direction you prefer. Let me know what makes the most sense for Guava's design.

chaoren:
> a bare LF in a quoted parameter could allow header injection attacks

Can you show an example of this?

> Since we already forbid CR for this reason

Who's "we"? The spec excludes it, so we do as well. Where is the reason given in the spec?

Upon closer reading of https://datatracker.ietf.org/doc/html/rfc822:

     quoted-string = <"> *(qtext/quoted-pair) <">; Regular qtext or
                                                 ;   quoted chars.
     
     qtext       =  <any CHAR excepting <">,     ; => may be folded
                     "\" & CR, and including
                     linear-white-space>

     linear-white-space =  1*([CRLF] LWSP-char)  ; semantics = SPACE
                                                 ; CRLF => folding

     LWSP-char   =  SPACE / HTAB                 ; semantics = SPACE

It seems to be saying that CR (by itself) is excepted, but `linear-white-space` is included, which can include CRLF.

Indeed, according to https://datatracker.ietf.org/doc/html/rfc822#section-3.4.5
     
        Where permitted (i.e., in words in structured fields)  quoted-
        strings  are  treated  as a single symbol.  That is, a quoted-
        string is equivalent to an atom, syntactically.  If a  quoted-
        string  is to be "folded" onto multiple lines, then the syntax
        for folding must be adhered to.  (See the "Lexical Analysis of
        Messages"  section  on "Folding Long Header Fields" above, and
        the section on "Case  Independence"  below.)   Therefore,  the
        official  semantics  do  not  "see" any bare CRLFs that are in
        quoted-strings; however particular parsing programs  may  wish
        to  note  their presence.  For such programs, it would be rea-
        sonable to interpret a "CRLF LWSP-char" as being a CRLF  which
        is  part  of the quoted-string; i.e., the CRLF is kept and the
        LWSP-char is discarded.  Quoted CRLFs (i.e., a backslash  fol-
        lowed  by  a CR followed by a LF) are also subject to rules of
        folding, but the presence of the quoting character (backslash)
        explicitly  indicates  that  the  CRLF  is  data to the quoted
        string.  Stripping off the first following LWSP-char  is  also
        appropriate when parsing quoted CRLFs.

CRLFs are explicitly allowed in quoted strings. It looks like our parser is actually stricter than necessary since we're not allowing CRLFs when we should.

artahir-dev:
Hey @chaoren, thanks again for taking the time to walk through the RFC details so thoroughly. I really appreciate it.

You’re absolutely right; I misread the spec initially. After digging back into those sections on `qtext` and `linear-white-space`, it’s clear that bare `\n` is technically allowed in quoted strings. Trying to block it in the parser would make Guava stricter than the standard requires, which could end up breaking valid use cases. That isn't the goal here.

I also see your point about the parser already being a bit overzealous by rejecting standalone `\r` when `CRLF` is permitted via folding rules. Trying to solve header injection at the parsing layer was probably the wrong move since it mixes syntax concerns with context-specific safety.

Since the PR is closed, I’m happy to drop this approach. If there’s interest, I’d be glad to explore adding a validation helper (something like `isSafeForHttpHeader()`) or pushing the check to the serialization layer instead, where headers are actually constructed. That feels like a cleaner way to handle it without compromising spec compliance.

Either way, thanks for the sharp review. It helped me see the bigger picture here.

PR #8297, opened 2026-03-25
Add overrides of `and`,`or`,`negate` for ObjectPredicate entries

MukjepScarlet:
I wanted to add an issue for this but I think it's simple enough, so I directly write my idea down.

### DIFF
--- a/guava/src/com/google/common/base/Predicates.java
+++ b/guava/src/com/google/common/base/Predicates.java
@@ -317,6 +317,22 @@ public final class Predicates {
       }
 
       @Override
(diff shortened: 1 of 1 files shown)


chaoren:
Why do these need to be overridden? `java.util.function.Predicate` already provides default implementations for all of these.

cpovirk:
Historically, we took pains to provide extreme micro optimizations and predictable `equals` behavior. I think we did it mostly "because we can," and it was likely identifiable as a mistake at the time. Nowadays, anyone who comes to rely on such behavior will have a harder time migrating to lambda `Predicate` implementations \[edit: internal CL 889291176\], so I think it's become even more regrettable.

PR #8298, opened 2026-03-26
Fix inconsistent weigher() javadoc in CacheBuilder

### LINKED ISSUE #1690: Inconsistency in CacheBuilder.weigher() and maximumWeight() javadoc
_[Original issue](https://code.google.com/p/guava-libraries/issues/detail?id=1690) created by **shtratos** on 2014-03-06 at 08:45 AM_

---

Here: https://google.github.io/guava/apidocs/com/google/common/cache/CacheBuilder.html#weigher(com.google.common.cache.Weigher)

maximumWeight() javadoc says:
... When weight is zero, elements will be evicted immediately after being loaded into cache. 

Note that weight is only used to determine whether the cache is over capacity; it has no effect on selecting which entry should be evicted next. ...

Which contradicts to weigher() javadoc:
... Entry weight is taken into consideration by maximumWeight(long) when determining which entries to evict, ...

When the weight of an entry is zero it will not be considered for size-based eviction (though it still may be evicted by other means).

daguimu:
## Problem

The `weigher()` javadoc in `CacheBuilder` states that entry weight is used "when determining which entries to evict", contradicting the `maximumWeight()` javadoc which correctly states that "weight is only used to determine whether the cache is over capacity; it has no effect on selecting which entry should be evicted next."

## Root Cause

The `weigher()` javadoc was written with imprecise wording. The implementation in `LocalCache.Segment.getNextEvictable()` confirms that eviction candidate selection is based on access recency (LRU via the `accessQueue`), not entry weight. Weight only affects whether the total capacity threshold has been exceeded, triggering eviction.

## Fix

- Updated the `weigher()` javadoc to say weight is used "when determining whether the cache is over capacity" instead of "when determining which entries to evict"
- Applied the same fix to both JRE and Android variants

## Impact

Javadoc-only change. No behavioral change. Resolves the documented inconsistency between `weigher()` and `maximumWeight()` javadoc.

Fixes #1690

### DIFF
--- a/android/guava/src/com/google/common/cache/CacheBuilder.java
+++ b/android/guava/src/com/google/common/cache/CacheBuilder.java
@@ -549,10 +549,10 @@ public final class CacheBuilder<K, V> {
 
   /**
    * Specifies the weigher to use in determining the weight of entries. Entry weight is taken into
-   * consideration by {@link #maximumWeight(long)} when determining which entries to evict, and use
-   * of this method requires a corresponding call to {@link #maximumWeight(long)} prior to calling
-   * {@link #build}. Weights are measured and recorded when entries are inserted into the cache, and
-   * are thus effectively static during the lifetime of a cache entry.
+   * consideration by {@link #maximumWeight(long)} when determining whether the cache is over
+   * capacity, and use of this method requires a corresponding call to {@link #maximumWeight(long)}
+   * prior to calling {@link #build}. Weights are measured and recorded when entries are inserted
+   * into the cache, and are thus effectively static during the lifetime of a cache entry.
    *
    * <p>When the weight of an entry is zero it will not be considered for size-based eviction
    * (though it still may be evicted by other means).
--- a/guava/src/com/google/common/cache/CacheBuilder.java
+++ b/guava/src/com/google/common/cache/CacheBuilder.java
@@ -549,10 +549,10 @@ public final class CacheBuilder<K, V> {
 
   /**
    * Specifies the weigher to use in determining the weight of entries. Entry weight is taken into
-   * consideration by {@link #maximumWeight(long)} when determining which entries to evict, and use
-   * of this method requires a corresponding call to {@link #maximumWeight(long)} prior to calling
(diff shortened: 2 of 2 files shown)


chaoren:
@ben-manes, looks like the Caffeine docs have the same inconsistency:

https://github.com/ben-manes/caffeine/blob/fbadda61eac012f7553e6a62e05ace7acd63f0f1/caffeine/src/main/java/com/github/benmanes/caffeine/cache/Caffeine.java#L473-L474

```
   * Specifies the weigher to use in determining the weight of entries. Entry weight is taken into
   * consideration by {@link #maximumWeight(long)} when determining which entries to evict, and use
```

https://github.com/ben-manes/caffeine/blob/fbadda61eac012f7553e6a62e05ace7acd63f0f1/caffeine/src/main/java/com/github/benmanes/caffeine/cache/Caffeine.java#L451-L452

```
   * Note that weight is only used to determine whether the cache is over capacity; it has no effect
   * on selecting which entry should be evicted next.
```

ben-manes:
That is a fair confusion, though *"consideration"* means its left to `maximumWeight` to decide how it may be used. That saying it is only used for the capacity limit is the algorithmic details that is the policy's choice of useful information, whereas a Weigher just provides the information. I don't think `.weigher(...)` should include that detail, but I can see how the word `consideration` might be too subtle.

We did collaborate to investigate using the weight as part of the eviction decision and shared our findings ([paper](https://dl.acm.org/doi/10.1145/3507920?cid=99659224047)). There are some modest gains but also adds complexity wrt concurrency, and for an on-heap Java cache it was unclear how beneficial the enhancement would be in practice so we backlogged it.

chaoren:
I think these two statements in particular are in conflict:

> weight is taken into consideration ... when determining which entries to evict

> [weight] has no effect on selecting which entry should be evicted next

Maybe `weigher` doesn't need to say anything about how the weight is going to be used, just that it would used by `maximumWeight`. Declaring that `maximumWeight` would use it to determine which entries to evict directly contradicts the documentation on `maximumWeight`. I think we should at least remove that part.

PR #8301, opened 2026-03-27
fix: computeIfAbsent should not update write timestamp for existing entries

### LINKED ISSUE #3700: LocalCache computeIfAbsent updates write timestamp on existing entries, affecting expireAfterWrite caches
It seems that there is a bug in `LocalCache`'s implementation of `Map.computeIfAbsent(K, Function<? super K, ? extends V>)`, specifically affecting an `expireAfterWrite` cache.

**Expected behaviour:**
If there is an existing entry for key `K`, calling `computeIfAbsent(K, ...)` should not modify the existing entry for K (unless the entry is expired and cleaned up). _**In particular, the write timestamp of entry K should not be changed.**_ Of course, if there is no existing entry for `K`, a new value is computed and a write timestamp set.

**Actual behaviour:**
If there is an existing entry for key `K`, calling `computeIfAbsent(K, ...)` will update the write timestamp (and write queue) for that existing entry. (As expected, if there is no existing entry for `K`, a new value is computed and a write timestamp set.)

**Implication:**
Because the write timestamp of the entry is reset, if an entry is accessed using `computeIfAbsent()` repeatedly in intervals less that the `expireAfterWrite` duration, it will never expire, acting more like an `expireAfterAccess` cache.

**Related code in `LocalCache`:**
where `recordWrite()` is called: [LocalCache.java#L2248](https://github.com/google/guava/blob/cc6cff2df9d66d4d9cd18f378cb75d03f7002852/guava/src/com/google/common/cache/LocalCache.java#L2248)
`computeIfAbsent()` implementation: [LocalCache.java#L4201](https://github.com/google/guava/blob/cc6cff2df9d66d4d9cd18f378cb75d03f7002852/guava/src/com/google/common/cache/LocalCache.java#L4201)

Proposed test case in `LocalCacheTest`, which currently will fail:
```java
  public void testComputeIfAbsent_ExistingEntryWriteTimeNotChanged() {
    LocalCache<Object, Object> map =
          makeLocalCache(
                createCacheBuilder()
                      .expireAfterWrite(1, MINUTES)
                      .expireAfterAccess(1, MINUTES));

    String key = "a";
    map.put(key, "b");
    ReferenceEntry<Object, Object> entry = map.getEntry(key);
    Object originalValue = entry.getValueReference().get();
    long originalWriteTime = entry.getWriteTime();
    long originalAccessTime = entry.getAccessTime();

    map.computeIfAbsent(key, k -> "c");
    assertEquals(originalValue, map.get(key));
    assertEquals("Write time should not be changed, since the value was not changed",
          originalWriteTime, entry.getWriteTime());
    assertThat(entry.getAccessTime()).isGreaterThan(originalAccessTime);
  }
```

Versions:
- Discovered using `guava:27.0-jre`
- Introduced in [Release Java 8 changes to Guava (73e382f)](https://github.com/google/guava/commit/73e382fa877f80994817a136b0adcc4365ccd904#diff-fe898d98ea517870f437b629cdd91698)
- This issue still exists in master.

**Specific use case:**
A polling endpoint which checks an authentication cache (expire after write) that is accessed via Spring Boot's `@Cacheable` AOP abstraction. The implementation of this abstraction uses the cache's `Map.computeIfAbsent()` method when synchronized access is needed:
Spring caching abstraction, synchronized access: [CacheAspectSupport.java#L381](https://github.com/spring-projects/spring-framework/blob/6fa9871a7001b9f3a7bcb325dcd6981cfdf2c025/spring-context/src/main/java/org/springframework/cache/interceptor/CacheAspectSupport.java#L381)
Actual call to `computeIfAbsent()` in Spring: [ConcurrentMapCache.java#L144](https://github.com/spring-projects/spring-framework/blob/6fa9871a7001b9f3a7bcb325dcd6981cfdf2c025/spring-context/src/main/java/org/springframework/cache/concurrent/ConcurrentMapCache.java#L144)

daguimu:
## Problem

`LocalCache.computeIfAbsent()` incorrectly resets the write timestamp when called on an existing key, even though the value is not changed. This causes `expireAfterWrite` caches to behave like `expireAfterAccess` caches — entries never expire as long as they are accessed via `computeIfAbsent()` at intervals shorter than the write expiry duration.

This is particularly problematic for Spring Boot applications using `@Cacheable` with synchronized access, which delegates to `computeIfAbsent()` internally.

## Root Cause

`computeIfAbsent()` delegates to `Segment.compute()`. When the computed value is the same reference as the existing value (`newValue == valueReference.get()`), the code path at `LocalCache.java:2278` called `recordWrite(e, 0, now)`, which resets the write timestamp via `entry.setWriteTime(now)`. This incorrectly extends the `expireAfterWrite` deadline.

## Fix

Replaced the `recordWrite(e, 0, now)` call with inline logic that:
- Updates the access time (for `expireAfterAccess` correctness)
- Re-adds the entry to both `accessQueue` and `writeQueue` (required because they were removed earlier in the `compute()` method at lines 2251-2252)
- Does NOT reset the write timestamp, preserving the original `expireAfterWrite` deadline

## Tests Added

| Change Point | Test |
|---|---|
| Do not reset write time when value unchanged | `testComputeIfAbsent_existingEntryWriteTimeNotChanged()` — verifies `getWriteTime()` is unchanged after `computeIfAbsent` on existing key |
| Entry properly expires despite repeated computeIfAbsent | `testComputeIfAbsent_existingEntryExpiresAfterWrite()` — verifies entry expires after write duration even with repeated `computeIfAbsent` calls |
| Regression: compute() with new value still updates write time | `testCompute_returningNewValueUpdatesWriteTime()` — verifies that `compute()` returning a new value still correctly updates write time |

## Impact

Only affects the code path in `Segment.compute()` where the computed value is the same reference as the existing value. This path is exercised by `computeIfAbsent()` on existing keys, and also by `compute()`/`computeIfPresent()`/`merge()` when their function returns the existing value unchanged. The new-value and removal paths are unaffected.

Fixes #3700

### DIFF
--- a/guava-tests/test/com/google/common/cache/LocalCacheTest.java
+++ b/guava-tests/test/com/google/common/cache/LocalCacheTest.java
@@ -795,6 +795,76 @@ public class LocalCacheTest extends TestCase {
     assertThat(notifications).isEmpty();
   }
 
+  public void testComputeIfAbsent_existingEntryWriteTimeNotChanged() {
+    FakeTicker ticker = new FakeTicker();
+    LocalCache<Object, Object> map =
+        makeLocalCache(
+            createCacheBuilder()
+                .concurrencyLevel(1)
+                .ticker(ticker)
+                .expireAfterWrite(10, NANOSECONDS));
+
+    Object key = new Object();
+    Object value = new Object();
+    map.put(key, value);
+    ReferenceEntry<Object, Object> entry = map.getEntry(key);
+    long originalWriteTime = entry.getWriteTime();
+
+    ticker.advance(5, NANOSECONDS);
+
+    // computeIfAbsent on an existing key should NOT update write time
+    Object result = map.computeIfAbsent(key, k -> new Object());
+    assertThat(result).isSameInstanceAs(value);
+    assertThat(entry.getWriteTime()).isEqualTo(originalWriteTime);
+  }
+
+  public void testComputeIfAbsent_existingEntryExpiresAfterWrite() {
+    FakeTicker ticker = new FakeTicker();
+    LocalCache<Object, Object> map =
+        makeLocalCache(
+            createCacheBuilder()
+                .concurrencyLevel(1)
+                .ticker(ticker)
+                .expireAfterWrite(10, NANOSECONDS));
+
+    Object key = new Object();
+    Object value = new Object();
+    map.put(key, value);
+
+    // Repeatedly call computeIfAbsent before expiry
+    for (int i = 0; i < 5; i++) {
+      ticker.advance(3, NANOSECONDS);
+      map.computeIfAbsent(key, k -> new Object());
+    }
+
+    // After 15ns total (> 10ns expiry), the entry must be expired
+    assertThat(map.get(key)).isNull();
+  }
+
+  public void testCompute_returningNewValueUpdatesWriteTime() {
+    FakeTicker ticker = new FakeTicker();
+    LocalCache<Object, Object> map =
+        makeLocalCache(
+            createCacheBuilder()
+                .concurrencyLevel(1)
+                .ticker(ticker)
+                .expireAfterWrite(10, NANOSECONDS));
+
+    Object key = new Object();
+    Object value = new Object();
+    map.put(key, value);
+    ReferenceEntry<Object, Object> entry = map.getEntry(key);
+    long originalWriteTime = entry.getWriteTime();
+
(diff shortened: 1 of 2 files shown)


koppor:
Note that the issue says: Migrate to https://github.com/ben-manes/caffeine

    implementation("com.github.ben-manes.caffeine:caffeine:3.2.4")

PR #8371, opened 2026-04-23
Add ConcurrentMap spliterator checks

### LINKED ISSUE #7855: testlib: add tests for Map.entrySet().stream()

holly-agyei:
Adds ConcurrentMap-specific testlib coverage for derived keySet and entrySet spliterators, and updates LocalCache to satisfy it in both the JRE and Android source trees.

Fixes #7855

Implementation:
- add a new ConcurrentMapSpliteratorTester that checks keySet() and entrySet() spliterators report CONCURRENT and do not report SIZED
- wire that tester into ConcurrentMapTestSuiteBuilder so ConcurrentMap suites pick it up automatically
- add a focused ConcurrentMapTestSuiteBuilder self-test that verifies both the passing and failing cases
- override LocalCache derived-set spliterators in both JRE and Android variants to use concurrent unknown-size spliterators

Process:
- read and followed the repository guidance and matched the surrounding testlib and cache code style
- kept the scope to the issue and added regression coverage for the new behavior

Validation:
- direct junit.textui.TestRunner execution of com.google.common.collect.testing.ConcurrentMapTestSuiteBuilderTest passed
- direct javac compilation of the changed testlib and LocalCache sources passed
- full Maven test execution was not possible in this environment because Maven project loading failed on unresolved build extensions during local verification

### DIFF
--- a/guava-testlib/src/com/google/common/collect/testing/testers/ConcurrentMapSpliteratorTester.java
+++ b/guava-testlib/src/com/google/common/collect/testing/testers/ConcurrentMapSpliteratorTester.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright (C) 2026 The Guava Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.common.collect.testing.testers;
+
+import com.google.common.annotations.GwtCompatible;
+import com.google.common.collect.testing.AbstractMapTester;
+import java.util.Spliterator;
+import java.util.concurrent.ConcurrentMap;
+import org.jspecify.annotations.NullMarked;
+import org.junit.Ignore;
+
+/**
+ * A generic JUnit test which tests spliterator characteristics on derived views of a concurrent
+ * map. Can't be invoked directly; please see {@link
+ * com.google.common.collect.testing.ConcurrentMapTestSuiteBuilder}.
+ *
+ * @author Louis Wasserman
+ */
+@GwtCompatible
+@Ignore("test runners must not instantiate and run this directly, only via suites we build")
+// @Ignore affects the Android test runner, which respects JUnit 4 annotations on JUnit 3 tests.
+@SuppressWarnings("JUnit4ClassUsedInJUnit3")
+@NullMarked
+public class ConcurrentMapSpliteratorTester<K, V> extends AbstractMapTester<K, V> {
+  @Override
+  protected ConcurrentMap<K, V> getMap() {
+    return (ConcurrentMap<K, V>) super.getMap();
+  }
+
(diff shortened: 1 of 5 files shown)


holly-agyei:
CI fix: the earlier run failed because the `spliterator()` override added to `android/guava/src/com/google/common/cache/LocalCache.java` referenced `java.util.Spliterator` / `Spliterators.spliteratorUnknownSize(...)`, which Animal Sniffer flagged against the Android API 23 signature (`gummy-bears-api-23`).

Pushed a follow-up commit (bcd722f9) that annotates the new method with the package-local `@IgnoreJRERequirement`, following the convention already used for `spliterator()` in `android/guava/src/com/google/common/collect/ImmutableCollection.java` and primitive array wrappers. The main `guava/src` variant is unchanged because it's checked against the `java18` signature, which includes `Spliterator`.

Verified locally:
- `./mvnw animal-sniffer:check -pl guava -f android/pom.xml` — BUILD SUCCESS
- `./mvnw install -DskipTests=true -f android/pom.xml` — BUILD SUCCESS (all 5 reactor modules)
- `./mvnw -pl guava-tests test -Dtest='LocalCacheTest,CacheBuilderSpecTest,CacheLoadingTest' -f android/pom.xml` — 8,139 tests pass
- `./mvnw -pl guava-tests test -Dtest='LocalCacheTest,CacheBuilderSpecTest,CacheLoadingTest,ConcurrentMapTestSuiteBuilderTest,ConcurrentHashMultisetTest,MapMakerInternalMapTest'` (main variant) — 13,998 tests pass

cgdecker:
I think `LocalCache.Values` (which extends `AbstractCollection`) should also have its `spliterator()` implementation changed in a similar way (`CONCURRENT` and `NONNULL`).

guava-testlib/src/com/google/common/collect/testing/testers/ConcurrentMapSpliteratorTester.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.concurrent.ConcurrentMap;
+import org.jspecify.annotations.NullMarked;
+import org.junit.Ignore;
+
+/**
+ * A generic JUnit test which tests spliterator characteristics on derived views of a concurrent
+ * map. Can't be invoked directly; please see {@link
+ * com.google.common.collect.testing.ConcurrentMapTestSuiteBuilder}.
+ *
+ * @author Louis Wasserman
```
cgdecker: This `@author` tag shouldn't be here. Besides being inaccurate here, we don't use `@author` tags at all anymore in new code.

holly-agyei: @cgdecker, thanks for the catch. Both fixed in d9e665df5.

On the @author tag: I had copied the file header from a sibling tester (MapSpliteratorTester) as a starting point and missed removing the @author Louis Wasserman line. Sorry about that. I Dropped the tag entirely since new Guava code doesn't use them anymore.

holly-agyei: @cgdecker, thanks for the catch. Both fixed in https://github.com/google/guava/commit/d9e665df576fb44448c57d7339318ce4cc727626.

On the https://github.com/author tag: I had copied the file header from a sibling tester (MapSpliteratorTester) as a starting point and missed removing the https://github.com/author Louis Wasserman line. Sorry about that. I Dropped the tag entirely since new Guava code doesn't use them anymore.

PR #8380, opened 2026-04-27
aggregate-related cases out of FuturesTest into a dedicated test

akashkumar916:
Split the aggregate-related cases out of FuturesTest into a dedicated test class so the main file is easier to navigate. Keep AggregateFutureStateFallbackAtomicHelperTest explicitly wired to the aggregate-focused test set, rather than relying on the old “run everything in FuturesTest” behavior.

Double-check that the fallback test still covers the relevant AggregateFutureState paths and doesn’t accidentally run too little after the split. Mirror the same structure in the Android tree and run the targeted futures test suites. 

The main goal is to improve navigation without reducing the effective fallback-helper coverage you called out above


chaoren:
I don't think we need to do this.

PR #8442, opened 2026-05-22
Add fallback variants to Gradle module metadata for non-JVM consumers

### LINKED ISSUE #8398: Gradle unable to resolve Guava in a project, with a VariantSelectionByAttributesException error
### Guava Version

33.5.0

### Description

In a gradle project I have a build that uses some Java tooling as part of its 
build machinery, but is not, itself, a Java project. One of the tooling 
dependencies (javaparser) has a dependency on Guava.

For historical reasons that project's `build.gradle.kts` applied the 
`java-library` plugin. When I removed it because its default tasks were causing 
problems with other tasks and, as noted, no Java artifacts were being produced,
the build failed with an error:

    Variant Selection Exception: 
        org.gradle.internal.component.resolution.failure.exception.VariantSelectionByAttributesException 
        caused by Resolution Failure: 
        org.gradle.internal.component.resolution.failure.type.NoCompatibleVariantsFailure

Gradle can no longer resolve a Guava variant that it can use. Note that `javaparser` is explicitly depending on the `-jre` Guava version. (https://github.com/javaparser/javaparser/blob/11a99ae61df28b19fb9b64e5f7c216fa71dc4020/pom.xml#L432-L437)

There's a minimal repro here: https://github.com/Saxonica/guava-variant-problem

This project contains two Gradle sub-projects, `java_lib` and 
`non_java` whose `build.gradle.kts` differ only in the plugin they apply.

`non_java` applies the `base` plugin, and `java_lib` applies the `java-library` 
plugin. They both contain a simple configuration which declares a dependency on
Guava. 

Running `./gradlew java_lib:dependencies` shows that Guava resolves fine, and
`./gradlew non_java:depdendencies` shows the failure.

This is defintely unexpected behvaiour, and seems like buggy behaviour, though I
am really not sure if this is a Gradle problem or a Guava Gradle metadata
problem.

I imagine this is related to #7059, #6801 and friends

### Example

```java
Clone https://github.com/Saxonica/guava-variant-problem and run the following:


./gradlew non_java:dependencies`
```

### Expected Behavior

A properly resolved dependency tree, whose `macguffin` configuration's dependency tree should be identical to the result of running `./gradlew java_lib:dependencies`:

```
> Task :non_java:dependencies

------------------------------------------------------------
Project ':non_java'
------------------------------------------------------------

macguffin
\--- com.google.guava:guava:33.6.0-jre
     +--- com.google.guava:failureaccess:1.0.3
     +--- com.google.guava:listenablefuture:9999.0-empty-to-avoid-conflict-with-guava
     +--- org.jspecify:jspecify:1.0.0
     +--- com.google.errorprone:error_prone_annotations:2.47.0
     \--- com.google.j2objc:j2objc-annotations:3.1

macguffinRuntime (n)
\--- com.google.guava:guava:33.6.0-jre (n)

(n) - A dependency or dependency configuration that cannot be resolved.

A web-based, searchable dependency report is available by adding the --scan option.

BUILD SUCCESSFUL in 414ms
```

### Actual Behavior

```
> Task :non_java:dependencies

------------------------------------------------------------
Project ':non_java'
------------------------------------------------------------

macguffin
\--- com.google.guava:guava:33.6.0-jre FAILED

macguffinRuntime (n)
\--- com.google.guava:guava:33.6.0-jre (n)

(n) - A dependency or dependency configuration that cannot be resolved.

A web-based, searchable dependency report is available by adding the --scan option.

BUILD SUCCESSFUL in 414ms
```

### Packages

_No response_

### Platforms

_No response_

### Checklist

giruuuuj:
Add apiElements and runtimeElements variants without the org.gradle.jvm.environment attribute. This allows projects without the jvm-ecosystem plugin to resolve Guava as a dependency.

Fixes #8398

When a Gradle project applies only the base plugin, Gradle variant selection fails because Guava variants all declare org.gradle.jvm.environment. The fallback variants omit this attribute so Gradle can resolve them without the JVM ecosystem plugin.

RELNOTES=Add fallback Gradle module metadata variants for non-JVM consumers

### DIFF
--- a/android/guava/src/com/google/common/collect/ImmList.java
+++ b/android/guava/src/com/google/common/collect/ImmList.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright (C) 2026 The Guava Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.google.common.collect;
+
+import com.google.common.annotations.GwtCompatible;
+import java.util.List;
+
+/**
+ * A {@link List} whose contents will never change. {@code ImmList} is the interface counterpart of
+ * {@link ImmutableList}, allowing alternative implementations.
+ *
+ * <p>All mutation methods ({@link #add}, {@link #remove}, {@link #clear}, etc.) throw {@link
+ * UnsupportedOperationException}.
+ *
+ * @param <E> the element type
+ * @since 9999.0
+ */
+@GwtCompatible
+public interface ImmList<E> extends List<E> {}
--- a/android/guava/src/com/google/common/collect/ImmutableList.java
+++ b/android/guava/src/com/google/common/collect/ImmutableList.java
@@ -63,7 +63,7 @@ import org.jspecify.annotations.Nullable;
 @GwtCompatible
 @SuppressWarnings("serial") // we're overriding default serialization
 public abstract class ImmutableList<E> extends ImmutableCollection<E>
-    implements List<E>, RandomAccess {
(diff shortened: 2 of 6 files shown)


netdpb:
I feel like the solution in https://github.com/google/guava/issues/8398#issuecomment-4366214325 is the correct solution for rare users like that. We changed #8398 to a documentation issue because it's probably a good idea to document that fact.

I think the solution you're proposing here is a very significant change, and I don't think it's necessary given that there seems to be a supported Gradle solution.

Issue #8295, opened 2026-03-23
ClassPathTest.testLocationsFrom_idempotentScan failure


cpovirk:
```
Error:  Failures: 
Error:    ClassPathTest.testLocationsFrom_idempotentScan:492 unexpected (5)
#1      : target/surefire-reports/com.google.common.primitives.UnsignedLongsTest.txt
#2      : target/surefire-reports/com.google.common.reflect.AbstractInvocationHandlerTest.txt
#3      : target/surefire-reports/TEST-com.google.common.reflect.AbstractInvocationHandlerTest.xml
#4      : target/surefire-reports/com.google.common.reflect.ClassPathTest-output.txt
#5      : target/surefire-reports/TEST-com.google.common.primitives.UnsignedLongsTest.xml
---
expected: [benchmark/com/google/common/cache/SegmentBenchmark.java, benchmark/com/google/common/cache/ChainBenchmark.java, ...
```

https://github.com/google/guava/actions/runs/23457793433/job/68251479033

I guess that Surefire is putting its reports directory on the classpath? That seems weird. And then somehow some reports are getting written during the execution of an individual method in `ClassPathTest` (and not just `ClassPathTest-output.txt`)? That also seems weird.

I'm going to hope that this goes away on its own, but I expect that we'll see it again....

Aaron-PintoTamayo:
Hi cpovirk, I been looking at this for a while and dug into the source code and believe found the root cause. ClassPath.getClassLoaderUrls() calls parseJavaClassPath() which reads the java.class.path system property. Surefire appears to be adding target/surefire-reports/ to that system property during test execution, which is why it shows up in the classpath scan. Between the two scanResources() calls in testLocationsFrom_idempotentScan, Surefire writes new report files to that directory, causing the second scan to find files the first scan didn't. A potential fix could be to verify whether Surefire is indeed modifying java.class.path and if so, whether we can configure it not to include the reports directory. I could be wrong will investigate more.

Aaron-PintoTamayo:
I’m already investigating this and have found some more findings will share shortly.

Aaron-PintoTamayo:
[...]

PR #8443, opened 2026-05-22
Fix ClassPathTest flakiness by moving Surefire reports directory

giruuuuj:
Surefire writes report files to target/surefire-reports/ during test execution, and that directory can appear on java.class.path. This causes ClassPathTest.testLocationsFrom_idempotentScan to flake: new report files written between the test's two scanResources() calls make the second scan find files the first scan did not.\n\nMove the reports directory to target/surefire-reports-classpath-safe (a path that does not end up on the classpath).\n\nFixes #8295\n\nRELNOTES=n/a

### DIFF
--- a/android/guava-tests/pom.xml
+++ b/android/guava-tests/pom.xml
@@ -84,6 +84,19 @@
       </plugin>
       <plugin>
         <artifactId>maven-surefire-plugin</artifactId>
+        <configuration>
+          <!--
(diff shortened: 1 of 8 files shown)


cpovirk:
This does not appear to fix the problem: https://github.com/google/guava/issues/8295#issuecomment-4523392753

PR #8447, opened 2026-05-25
hash: guard BloomFilter.readFrom() against corrupt dataLength

adilburaksen:
## Problem

`BloomFilter.readFrom()` reads a signed 32-bit `dataLength` from the
stream and passes it directly to `Math.multiplyExact(dataLength, 64L)`
to size the backing `AtomicLongArray`, with no upper-bound validation.

A crafted stream can trigger two denial-of-service conditions:

**1. Large positive `dataLength`** — e.g. `Integer.MAX_VALUE` (7 bytes total):
```
0x01 0x01 0x7F 0xFF 0xFF 0xFF  (+ EOF)
```
`Math.multiplyExact(2147483647, 64L)` = 137 billion bits (fits in `long`, no exception).
`Ints.checkedCast(137438953408 / 64)` = 2,147,483,647 (fits in `int`, no exception).
`new AtomicLongArray(2147483647)` → allocates **~17.2 GB** → `OutOfMemoryError`.

**2. Negative `dataLength`** — e.g. `readInt()` returns `-1`:
`multiplyExact(-1, 64) = -64`. `LockFreeBitArray` constructor throws
`"data length is zero!"` instead of a clear corrupt-input message.

## Fix

Add two `checkArgument` guards immediately after `readInt()`:

```java
checkArgument(dataLength >= 0,
    "Corrupt BloomFilter: dataLength (%s) must be >= 0", dataLength);
checkArgument(dataLength <= Integer.MAX_VALUE / 64,
    "Corrupt BloomFilter: dataLength (%s) is unreasonably large (max: %s)",
    dataLength, Integer.MAX_VALUE / 64);
```

`checkArgument` is already statically imported. The thrown
`IllegalArgumentException` is caught by the existing `catch(Exception e)`
block and re-thrown as `IOException` with the
`"Unable to deserialize BloomFilter"` message — preserving the existing
API contract for callers.

The cap `Integer.MAX_VALUE / 64 = 33,554,431` limits the backing
`AtomicLongArray` to ~256 MB, which is sufficient for any realistic
Bloom filter use case.

## Impact

Any service that deserializes a `BloomFilter` from an untrusted source
(distributed cache, message queue, file upload, network protocol) is
vulnerable to a 7-byte payload that crashes the JVM process.

### DIFF
--- a/guava/src/com/google/common/hash/BloomFilter.java
+++ b/guava/src/com/google/common/hash/BloomFilter.java
@@ -653,6 +653,12 @@ public final class BloomFilter<T extends @Nullable Object> implements Predicate<
       strategyOrdinal = din.readByte();
       numHashFunctions = toUnsignedInt(din.readByte());
       dataLength = din.readInt();
+      checkArgument(dataLength >= 0, "Corrupt BloomFilter: dataLength (%s) must be >= 0", dataLength);
+      checkArgument(
+          dataLength <= Integer.MAX_VALUE / 64,
+          "Corrupt BloomFilter: dataLength (%s) is unreasonably large (max: %s)",
+          dataLength,
+          Integer.MAX_VALUE / 64);
 
       /*
        * We document in BloomFilterStrategies that we must not change the ordering, and we have a


adilburaksen:
@googlebot I signed it!

adilburaksen:
@googlebot I signed it!

swornim619:
hi @adilburaksen I reported this same issue to Google  on May 5, 2026 and also have an OSS-Fuzz PR open for fuzzing coverage: https://github.com/google/oss-fuzz/pull/15484

cpovirk:
I am neither the on-call person this week nor the primarily author of `hash`, so I'm hoping to leave this to one of those people, but I have various thoughts, which I'll leave here in the hope that they're helpful to one of those people:

- Reading a `BloomFilter` from a source that the caller doesn't know is safe seems likely to be rare, though I admit I can't rule it out. This is true to an even greater extent here than with normal serialization because the caller would need to explicitly request deserialization of a `BloomFilter` specifically with `readFrom`, not merely have `BloomFilter` on the classpath when a Java deserialization action allows the deserialization of anything on the classpath.
- Plain old Java deserialization for `BloomFilter` presumably has the same behavior because [its `SerializedForm` has an array member](https://github.com/google/guava/blob/a1d62bed86c4c4e318ac0b564d6ce641209823dc/guava/src/com/google/common/hash/BloomFilter.java#L588).
   - We might have other classes that have the same behavior. I am going to try to identify any.
   - But Java serialization at least allows you to put a [serialization filter](https://openjdk.org/jeps/290) in place to set various limits.
- Contrary to the original post, I would expect that Bloom filters can legitimately be huge, so I don't think there's any fixed limit we could set that would be both small enough to prevent DoS and large enough to allow all legitimate usage.
   - One alternative is to require callers of `readFrom` to pass in a maximum allowed size. This would be close to the serialization-filter approach. (Maybe we could even integrate automatically with Java serialization's own limits, but users might want different limits for `BloomFilter.readFrom` than for Java serialization.)
   - Another alternative is to grow the backing array only as actual data arrives, `ArrayList`-style. This at least requires an attacker to hang around to send a lot of data if it wants to force a lot of allocation—an improvement but not something that's easy to feel totally comfortable about. However, this requires doubling how much memory is held at once during deserialization. That's especially sad if someone does legitimately want gigabytes of data.

cpovirk:
We do have a number of `Serializable` classes that have (or whose `SerializedForm`/`SerialForm` double has) array members.

That said, I notice that `java.util.ArrayList` [will happily allocate an array as large as the serialization filter allows](https://github.com/openjdk/jdk/blob/ea99ed616f067268ebfd5b78e428e1835a84fe08/src/java.base/share/classes/java/util/ArrayList.java#L985), so I think we have done our duty there by implicitly respecting that. (I think that `ArrayList` has to check manually only because of its custom serialization, which it seems might relate to `clone()` or something? Ideally we'd confirm that.)

Still, here's a list of the examples that I found (beyond `BloomFilter` itself):

- `ArrayTable`
- `CompoundOrdering`
- `ImmutableList`
- `ImmutableMultiset`
- `ImmutableSet`
- `ImmutableSortedMultiset`
- `ImmutableSortedSet`
- `ImmutableTable`
- `Lists.OnePlusArrayList`
- `Lists.TwoPlusArrayList`
- `HashCode.BytesHashCode`
- (Google-internal `RandomAccessZipStream.ZipStreamEntry`)
- various `primitives` `*ArrayAsList` classes

Again, I think we can ignore Java serialization of this form: The biggest danger there would be if we _didn't_ respect the serialization filter, as was the case long ago for some of our classes that manually serialized size+contents but hopefully is no longer. (Hmm, but it's possible that we should separately consider adding manual checks against the serialization filter for them, too? Or how much is that best addressed through other checks done through the serialization filter, like the check on overall stream size?)

adilburaksen:
Thanks for the detailed analysis! I noticed @swornim619 had already reported this issue independently — small world. Even so, happy to keep the PR going if it's useful as a concrete fix. The distinction you're drawing is exactly right. The classes you listed go through `ObjectInputStream`, so the JEP 290 serialization filter can bound allocations naturally (via `maxarray`, `maxrefs`, stream size limits, etc.). `BloomFilter.readFrom()` is different — it reads a raw `int` from a `DataInputStream` and calls `new long[numLongs]` directly, completely outside the serialization filter machinery. There's no filter hook in that path.\n\nIf there are other Guava types with similar custom `readFrom()` / `DataInputStream`-based deserialization that allocate based on a caller-supplied size, I'm happy to extend this PR to cover them. Otherwise the scope here seems correct: just the `readFrom()` path on `BloomFilter`.

PR #8470, opened 2026-06-06
Document that Closer is not thread-safe

### LINKED ISSUE #3900: Closer is not thread-safe
The underlying stack is not thread safe. I encountered this issue in a scenario where lazily loaded items registered on a closer concurrently.
If not fixed it should at least be documented as not thread-safe.

Mike014:
[Fixes #3900](https://github.com/google/guava/issues/3900)

**Closer** uses a mutable ArrayDeque internally with no synchronization, 
making it **not thread-safe**. _This was undocumented_.

Added a **Javadoc** note to **both** the **JRE** and **Android versions** of Closer, 
following the existing pattern used by **[FileBackedOutputStream](https://github.com/google/guava/blob/master/guava/src/com/google/common/io/FileBackedOutputStream.java)**  
("This class is thread-safe.") — but stating the opposite.

Changes:
- guava/src/com/google/common/io/Closer.java
- android/guava/src/com/google/common/io/Closer.java

### DIFF
--- a/android/guava/src/com/google/common/io/Closer.java
+++ b/android/guava/src/com/google/common/io/Closer.java
@@ -74,6 +74,8 @@ import org.jspecify.annotations.Nullable;
  * <p>An exception that is suppressed is added to the exception that <i>will</i> be thrown using
  * {@code Throwable.addSuppressed(Throwable)}.
  *
+ * <p>This class is not thread-safe.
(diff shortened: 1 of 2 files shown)


cpovirk:
We may eventually figure out whether we want this, but I don't know yet.

PR #8479, opened 2026-06-14
fix(Splitter): throw IllegalStateException if trimmer already specified

Derylfabiensyah:
## Summary

Implements TODO(kevinb) in Splitter.trimResults(CharMatcher) - adds a checkState guard that throws IllegalStateException when trimResults() is called on a Splitter that already has a non-default trimmer configured.

## Problem

Previously, calling trimResults() multiple times on a Splitter would silently override the previous trimmer setting. This could lead to subtle bugs where a developer accidentally configures two different trimmers and the first one is silently discarded.

## Solution

- Added checkState(this.trimmer == CharMatcher.none(), ...) guard in trimResults(CharMatcher) to detect duplicate trimmer specification
- - Added @throws IllegalStateException to the Javadoc
- - Added import static com.google.common.base.Preconditions.checkState
## Files Changed

- android/guava/src/com/google/common/base/Splitter.java
- - guava/src/com/google/common/base/Splitter.java
Both android and main guava source trees received identical changes.

### DIFF
--- a/android/guava/src/com/google/common/base/Splitter.java
+++ b/android/guava/src/com/google/common/base/Splitter.java
@@ -16,6 +16,7 @@ package com.google.common.base;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
 import static java.util.Collections.unmodifiableList;
 import static java.util.Collections.unmodifiableMap;
 
@@ -349,10 +350,14 @@ public final class Splitter {
    * @param trimmer a {@link CharMatcher} that determines whether a character should be removed from
    *     the beginning/end of a subsequence
    * @return a splitter with the desired configuration
+   * @throws IllegalStateException if a trimmer was already specified on this splitter
    */
-  // TODO(kevinb): throw if a trimmer was already specified!
   public Splitter trimResults(CharMatcher trimmer) {
     checkNotNull(trimmer);
+    checkState(
+        this.trimmer == CharMatcher.none(),
+        "A trimmer was already specified: %s",
+        this.trimmer);
     return new Splitter(strategy, omitEmptyStrings, trimmer, limit);
   }
 
--- a/guava/src/com/google/common/base/Splitter.java
+++ b/guava/src/com/google/common/base/Splitter.java
@@ -16,6 +16,7 @@ package com.google.common.base;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkNotNull;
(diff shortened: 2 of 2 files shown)


cpovirk:
This is an interesting one because it's not entirely clear what the desired behavior would be:
- throw always
- throw if the trimmer is different from the original trimmer (though we can't always tell unless we were to do something very expensive)
- `or` together the two trimmers
- apply the trimmers in order (e.g., "trim spaces on the outside, then trim quotes," leaving behind any spaces inside the quotes, different from the previous bullet)
- \[edit: And the current behavior is itself defensible.\]

It also turns out that at least the first two options would make some people's code always throw. The other options are of course also changes to semantics, so they could cause trouble, too. In hindsight, we should have done something about this TODO in the past 16.5 years. At this point, we may be safest leaving it alone.

Derylfabiensyah:
Hi @cpovirk, thank you for the feedback and for sharing these insights! 

I completely understand the concern regarding backward compatibility and the risk of breaking existing code if we throw a runtime exception after all these years. 

To address this, here is a breakdown of why this validation is useful, along with some potential alternatives that avoid runtime breaking changes:

### 1. Why the Validation is Useful (The "Silent Override" Problem)
* **API Misdirection:** Developers calling `.trimResults(A).trimResults(B)` likely expect both trimmers to be applied (either composed via `or` or executed sequentially). The current behavior silently discards the first trimmer, which is a classic silent bug that can be very hard to debug in production.
* **Fulfilling the Original Intent:** The original codebase explicitly has a `// TODO(kevinb): throw if a trimmer was already specified!`, showing that the API designers recognized this as an issue that should prevent misuse.

---

### 2. Proposed Alternatives to Avoid Breaking Runtime Changes

If throwing an `IllegalStateException` at runtime is deemed too risky for backward compatibility, here are three alternative approaches we could take:

#### Option A: Implement a Compile-Time Check in Error Prone (Recommended)
Instead of modifying the runtime behavior of Guava, we could write a static analysis rule in **Error Prone** (e.g., under the Guava/Splitter checks). 
* **How it works:** It will flag any chain containing duplicate `.trimResults()` calls during compilation.
* **Pro:** Zero runtime risk for existing production code, but prevents developers from writing new bugs or copying bad patterns.

#### Option B: Compose the Trimmers (Union Matcher)
Instead of throwing or overriding, we could combine the existing trimmer with the new one:
```java
public Splitter trimResults(CharMatcher trimmer) {
  checkNotNull(trimmer);
  CharMatcher combinedTrimmer = (this.trimmer == CharMatcher.none()) 
      ? trimmer 
      : this.trimmer.or(trimmer);
  return new Splitter(strategy, omitEmptyStrings, combinedTrimmer, limit);
}

PR #8499, opened 2026-06-22
Saturate count() and entryIterator() in Multisets.sum() to avoid integer overflow

vasiliy-mikhailov:
`Multisets.sum(a, b)`'s `size()` already guards against overflow with `IntMath.saturatedAdd`, but `count()` and `entryIterator()` add the per-element counts with a plain `+`. When two multisets each hold close to `Integer.MAX_VALUE` of an element, the sum overflows to a **negative count** — which violates the `Multiset` contract (counts are non-negative) and makes `immutableEntry`'s `checkNonnegative` throw `IllegalArgumentException` (`count cannot be negative but was: -2147483648`).

This applies `IntMath.saturatedAdd` in `count()` and `entryIterator()`, matching what `size()` already does, so the count saturates at `Integer.MAX_VALUE`. Adds regression tests for both paths.

### DIFF
--- a/guava-tests/test/com/google/common/collect/MultisetsTest.java
+++ b/guava-tests/test/com/google/common/collect/MultisetsTest.java
@@ -161,6 +161,37 @@ public class MultisetsTest extends TestCase {
     assertThat(sum(ms1, ms2)).containsExactly("a", "b", "a");
   }
 
+  public void testSumEntryIteratorOverflow() {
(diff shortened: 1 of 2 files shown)


eamonnmcmanus (approved):
Thanks, this seems like a great idea. Because of how we sync with our internal codebase, there will be a separate PR with these changes which will close this one. You will be credited on that PR.

ALIFKA-HUB:
alus

PR #8519, opened 2026-06-26
refactor(Joiner): rename internal toString helper to toCharSequence

Derylfabiensyah:
Renamed the package-private helper method `toString(Object)` to `toCharSequence(Object)` in `Joiner.java` to resolve cpovirk's TODO. 

This clarifies the return type (`CharSequence`) and avoids shadowing the standard `Object.toString()` signature for internal calls. Both the standard and Android version source trees have been updated.


guava/src/com/google/common/base/Joiner.java
```diff
@@ -506,8 +506,7 @@ public MapJoiner useForNull(String nullText) {
     }
   }
 
-  // TODO(cpovirk): Rename to "toCharSequence."
-  CharSequence toString(@Nullable Object part) {
+  CharSequence toCharSequence(@Nullable Object part) {
```
ricardoofnl, line 509: The comment block further down in this method (around line 524) still references the old name:

> Its implementation avoids calling this `toString(Object)` method in the first place.

Since this PR renames the method, it would be good to update that mention to `toCharSequence(Object)` as well.

ricardoofnl (commented):
Nice mechanical rename. I checked all call sites in both trees and nothing was missed. There is one small leftover, which I flagged inline in both files.

android/guava/src/com/google/common/base/Joiner.java
```diff
@@ -470,8 +470,7 @@ public MapJoiner useForNull(String nullText) {
     }
   }
 
-  // TODO(cpovirk): Rename to "toCharSequence."
-  CharSequence toString(@Nullable Object part) {
+  CharSequence toCharSequence(@Nullable Object part) {
```
ricardoofnl, line 473: Same as in the JRE flavor: the comment block further down in this method (around line 488) still says `toString(Object)` and should be updated to `toCharSequence(Object)` as part of this rename.

Derylfabiensyah:
Hi @ricardoofnl, thanks for pointing that out! I've updated the comment block references in both `guava/src/com/google/common/base/Joiner.java` and `android/guava/src/com/google/common/base/Joiner.java` from `toString(Object)` to `toCharSequence(Object)`.

Issue #2648, opened 2016-11-17
symlink cycles create failures in ClassPath scan


jct-amzn:
Starting in at least r19, https://google.github.io/guava/releases/19.0/api/docs/com/google/common/reflect/ClassPath.html will eagerly traverse circular symlinks during the ClassPath scan.

Depending on the host platform and JVM settings, this can lead to OutOfMemoryError, StackOverflowError, or (depending on the depth limits in underlying file system) inconsistent resource listings.

The symlink cycle doesn't have to occur in the actual class path.  ClassPath  eagerly scans the Class-Path attribute in the manifests of any encountered jars.  So relative paths encountered there, or an empty Class-Path attribute value, causes recursive scans of arbitrary file system paths.

I do have a failing test case and candidate fix prepared.  I can submit a pull request for that, but wanted to file a bug report first, since the "correct" behavior is a little ambiguous in this case.

cgdecker:
Looks like this was fixed by 7c11b8d9220b710d5598b774e22cc774f68f40e7.

PR #8554, opened 2026-07-15
ClassPath: Reject manifest Class-Path entries that escape the JAR's parent directory

arunmanni-ai:
## Summary

`getClassPathFromManifest()` resolves relative `Class-Path` manifest entries via `new URL(jarFile.toURI().toURL(), path)` without validating that the resolved path remains within the originating JAR's parent directory. This allows a manifest containing entries like `../sensitive.jar` to cause `ClassPath.from()` to recursively scan and expose resources from JARs outside the intended classpath scope.

Issue Tracker: 520917227

## Problem

The existing safeguards (protocol restriction to `file:` URLs, deduplication via canonical path) do not prevent relative path traversal. Resources from out-of-scope JARs can be discovered and read via `ResourceInfo.asByteSource()` without triggering classloader security mechanisms.

This is relevant in environments that process untrusted or semi-trusted JARs, such as plugin systems, build tools, or CI/CD pipelines.

## Fix

Added a containment check in `getClassPathFromManifest()` that validates the canonical path of each resolved relative Class-Path entry against the parent directory of the originating JAR file. Entries that escape the directory boundary are logged as warnings and skipped.

Absolute `file:` URLs (used by Maven surefire and similar tools) are not affected by this check.

## Tests

Added three test cases:
- `testGetClassPathFromManifest_pathTraversalRejected` — single `../` traversal
- `testGetClassPathFromManifest_deepPathTraversalRejected` — nested `../../` traversal
- `testGetClassPathFromManifest_mixedValidAndTraversal` — mix of valid and traversal entries

All 46 existing ClassPathTest cases pass with this change.

## Related

- #2648 — addressed symlink cycles in ClassPath scanning (different root cause, fixed deduplication only)

### DIFF
--- a/guava-tests/test/com/google/common/reflect/ClassPathTest.java
+++ b/guava-tests/test/com/google/common/reflect/ClassPathTest.java
@@ -400,6 +400,27 @@ public class ClassPathTest extends TestCase {
         .containsExactly(fullpath("base/relative.jar"));
   }
 
+  public void testGetClassPathFromManifest_pathTraversalRejected() throws IOException {
+    File jarFile = new File("base/some.jar");
+    Manifest manifest = manifestClasspath("../outside.jar");
+    assertThat(ClassPath.getClassPathFromManifest(jarFile, manifest)).isEmpty();
+  }
+
+  public void testGetClassPathFromManifest_deepPathTraversalRejected() throws IOException {
+    File jarFile = new File("base/nested/some.jar");
+    Manifest manifest = manifestClasspath("../../outside.jar");
+    assertThat(ClassPath.getClassPathFromManifest(jarFile, manifest)).isEmpty();
+  }
+
+  public void testGetClassPathFromManifest_mixedValidAndTraversal() throws IOException {
+    if (isWindows()) {
+      return;
+    }
+    File jarFile = new File("base/some.jar");
+    Manifest manifest = manifestClasspath("valid.jar ../escape.jar");
+    assertThat(ClassPath.getClassPathFromManifest(jarFile, manifest))
+        .containsExactly(fullpath("base/valid.jar"));
+  }
   public void testGetClassName() {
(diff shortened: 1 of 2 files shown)


arunmanni-ai:
Fixes #8555

arunmanni-ai:
Hi @cpovirk, thanks for reviewing. Just wanted to confirm — is this being handled internally, or was there an issue with the approach? Happy to make any changes if needed.

cpovirk:
We've seen the same uptick in security reports as many other projects, and that has included from people who spray questionable reports around to many projects, so I've been erring on the side of conserving our time by closing reports without comment when they look like false positives to me. Since I didn't see evidence that *you* were filing a lot of reports, I wondered if you'd be the first person that we heard back from, and sure enough, here we are :) So maybe I'll learn that there's more going on here than I realize.

Anyway, as suggested above, I don't think this is a vulnerability: If the attacker has the ability to put a hostile jar onto your classpath, then you almost certainly have [bigger problems](https://www.google.com/search?q=site%3Ahttps%3A%2F%2Fdevblogs.microsoft.com%2Foldnewthing+%22It+rather+involved+being+on+the+other+side+of+the+airtight+hatchway). Even if the jar were to not be allowed to contain code, only Class-Path entries, then the attacker can use that to put *other* jars on the classpath, which can themselves contain code or other resources. Overall, `ClassPath` is just providing a convenient API to the classes and resources that Java is already willing to load for you. Additionally, as you've noted, Class-Path entries are a valid thing for jars to use, so we "have to" follow them (again, as Java itself does). Once we are willing to do that, we can get into trouble without `../` segments, not just by reading jars from the same directory but also by reading from `file:///`.

Now, security is always a complicated thing, and I may be overlooking something. I just don't see a case being made so far here that there is a problem.

arunmanni-ai:
Thanks for the detailed explanation, I appreciate you taking the time. That makes sense from Guava's perspective.

PR #8597, opened 2026-08-08
Clarify BloomFilter.put return value under concurrency

### LINKED ISSUE #3055: BloomFilter multithread problem: put() NOT atomic
Hello team!

I just found a problem using `com.google.common.hash.BloomFilter` with more than one thread. Long story short: `put()` is not atomic so I am able to reproduce a **false negative**.

This is the config in `pom.xml`:
```xml
<dependency>
   <groupId>com.google.guava</groupId>
   <artifactId>guava</artifactId>
   <version>24.0-jre</version>
</dependency>
```

I created a simple [Test.java](https://github.com/google/guava/files/1781560/Test.java.txt) class that proves the point. The fix is also included in the file: just uncomment lines 56 and 58 (make `put()` synchronized).

Looking at the `BloomFilter` javadoc I see: 
```
 * As of Guava 23.0, this class is thread-safe and lock-free. It internally uses atomics and
 * compare-and-swap to ensure correctness when multiple threads are used to access it.
```

So I suppose that I'm not doing anything "wrong". Can you please confirm the issue or tell me where I'm using the library wrong?

Thanks!

NgoQuocViet2001:
Fixes #3055.

`BloomFilter.put` is lock-free, so concurrent calls for the same object can each set a different subset of bits and both return `true`. The existing Javadoc instead states that a `true` result definitely marks the first insertion.

This change scopes the first-insertion and `mightContain` return-value guarantees to nonconcurrent calls and documents the concurrent case. Runtime behavior is unchanged.

Tests:
- Ran: `.\mvnw.cmd -B -ntp clean install` (865,730 tests; 0 failures or errors)

### DIFF
--- a/guava/src/com/google/common/hash/BloomFilter.java
+++ b/guava/src/com/google/common/hash/BloomFilter.java
@@ -187,11 +187,13 @@ public final class BloomFilter<T extends @Nullable Object> implements Predicate<
    * Puts an element into this {@code BloomFilter}. Ensures that subsequent invocations of {@link
    * #mightContain(Object)} with the same element will always return {@code true}.
    *
-   * @return true if the Bloom filter's bits changed as a result of this operation. If the bits
-   *     changed, this is <i>definitely</i> the first time {@code object} has been added to the
-   *     filter. If the bits haven't changed, this <i>might</i> be the first time {@code object} has
-   *     been added to the filter. Note that {@code put(t)} always returns the <i>opposite</i>
-   *     result to what {@code mightContain(t)} would have returned at the time it is called.
+   * @return true if the Bloom filter's bits changed as a result of this operation. In the absence
+   *     of concurrent calls to {@code put}, if the bits changed, this is <i>definitely</i> the first
+   *     time {@code object} has been added to the filter. If the bits haven't changed, this
+   *     <i>might</i> be the first time {@code object} has been added to the filter. Note that, in the
+   *     absence of concurrent calls to {@code put}, {@code put(t)} always returns the
+   *     <i>opposite</i> result to what {@code mightContain(t)} would have returned at the time it is
+   *     called. With concurrent calls, multiple calls may return true for the same object.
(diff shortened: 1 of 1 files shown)


MuhammadNiazAli (approved):
Yeh bhi ek achha open PR hai (google/guava #8597)  chaliye is ka review likhte hain:

Review:

This is a meaningful correctness clarification for concurrent usage of BloomFilter.put(), and the change is well-scoped to just the Javadoc.

Core addition is the right one: the new sentence "In the absence of concurrent calls to put, ... With concurrent calls, multiple calls may return true for the same object" directly addresses a subtle race that the old doc glossed over. Without this qualifier, a reader could reasonably (and incorrectly) assume the "opposite of mightContain" guarantee always holds — which breaks down under concurrent puts racing on the same bits.
Why this matters in practice: someone using the return value of put() to decide "am I the first caller to add this object" (a common dedup pattern with Bloom filters) needs to know that under concurrency this signal isn't reliable multiple threads can all see true. Calling that out explicitly, rather than leaving it implied, is exactly the kind of thing that prevents a subtle production bug.
Wording is precise: "In the absence of concurrent calls to put" as the qualifier for the deterministic behavior, followed by the concurrent caveat at the end, reads clearly and doesn't require re-reading the whole doc block to understand which part applies when.
One thing worth double-checking: the doc doesn't say whether all concurrent callers can get true, or just "more than one" — i.e., is it possible for zero callers to get true even though the object was in fact newly added? If that's a real possibility (depends on the implementation's synchronization), it might be worth clarifying so readers don't assume "at least one caller gets true" as a hard guarantee. If the implementation guarantees at least one true, that's worth stating too, since it's the more useful invariant for callers.
Scope is clean  single doc comment, no logic touched, matches the PR title exactly.

Good, precise fix. Approving, with the note above as an optional follow-up for extra clarity.

NgoQuocViet2001:
I signed it!

PR #8607, opened 2026-08-13
Docs: warn not to mutate previous in AbstractSequentialIterator

### LINKED ISSUE #1916: Should add a warning to the AbstractSequentialIterator class documentation
It's vital implementations do not mutate the previous value when overriding the `computeNext` method, otherwise the initial value will never be seen.

This also means you can't reuse the object passed to the `computeNext` method, since that object will be mutated before being returned through the iterator.

Hashim1999164:
## Summary
Fixes #1916

Documents that `computeNext` must not mutate its `previous` argument (and must not reuse that instance as the following element), because `next()` returns the same instance after `computeNext` returns.

## Test plan
* [ ] Review the class and `computeNext` Javadoc wording
* [ ] Confirm no code behavior changes

### DIFF
--- a/guava/src/com/google/common/collect/AbstractSequentialIterator.java
+++ b/guava/src/com/google/common/collect/AbstractSequentialIterator.java
@@ -36,6 +36,11 @@ import org.jspecify.annotations.Nullable;
  *     };
  * }
  *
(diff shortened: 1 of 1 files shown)


cpovirk:
https://github.com/google/guava/issues/1916#issuecomment-5302509631

PR #8608, opened 2026-08-13
Docs: note ClassPath.getResources keeps one entry per path

### LINKED ISSUE #2712: Documentation of ClassPath#getResources() is misleading
Javadocs on ClassPath#getResources()  reads:

> Returns all resources loadable from the current class path, including the class files of all
> loadable classes but excluding the "META-INF/MANIFEST.MF" file

This is not exactly true for files which has same path relative to JAR.
Eg. if I want to load files all resources named "META-INF/schema.xml" from all JAR files on classpath, it will pick only last one, due to nature of backing ImmutableSet.
It should be either mentioned in javadocs, or allow "duplicates" for example by using ImmutableList or by changing ResourceInfo#equals to consider URL instead of just resourceName;

Hashim1999164:
## Summary
Fixes #2712

Clarifies `ClassPath.getResources()` so callers know that `ResourceInfo` equality is by resource name and class loader, and that the returned `ImmutableSet` therefore keeps only one resource for a given path even when that path appears in multiple jars or directories.

## Test plan
* [ ] Review the updated `getResources` Javadoc
* [ ] Confirm no code behavior changes

### DIFF
--- a/guava/src/com/google/common/reflect/ClassPath.java
+++ b/guava/src/com/google/common/reflect/ClassPath.java
@@ -139,8 +139,12 @@ public final class ClassPath {
   }
 
   /**
-   * Returns all resources loadable from the current class path, including the class files of all
+   * Returns resources loadable from the current class path, including the class files of all
    * loadable classes but excluding the "META-INF/MANIFEST.MF" file.
+   *
+   * <p><b>Note:</b> {@link ResourceInfo} instances are equal when they share the same resource
+   * name and {@link ClassLoader}. Because this method returns an {@link ImmutableSet}, only one
+   * resource is kept for a given path even if that path appears in multiple jars or directories.
    */
   public ImmutableSet<ResourceInfo> getResources() {
     return resources;


MuhammadNiazAli (approved):
Yeh naya repo/PR hai google/guava #8608 (abhi tak Open hai, achha hai review submit ho sakega). Chaliye review likhte hain:

Review:

This is a small but genuinely useful doc fix  clarifying getResources()'s dedup behavior can save someone real debugging time if they assumed every resource on the classpath shows up in the result.

Removing "all" from the first line: Good catch. The old wording "Returns all resources loadable..." — was misleading precisely because the method does not return every resource; it keeps only one entry per path. Dropping "all" makes the opening line accurate instead of contradicting the new paragraph you added right below it.
New paragraph is clear and specific: "instances are equal when they share the same resource name and ClassLoader... only one resource is kept for a given path even if that path appears in multiple jars or directories" this explains both why (equality based on name + ClassLoader) and what the practical effect is (last-one-wins-ish dedup across jars). That's exactly the kind of detail that's easy to discover the hard way and hard to find in existing docs.
One small thing worth checking: since dedup is based on equality of ResourceInfo (name + ClassLoader), it might be worth a one-line note on which entry wins when the same path appears in multiple jars (first one found on the classpath order, presumably) this could matter for someone debugging a "which jar's resource am I actually getting" issue. Not blocking, but it's the natural next question a reader would have after this paragraph.
Scope is right: pure Javadoc change, one file, no logic touched exactly what a "Docs:" prefixed PR should be.

Nit aside, this is a solid, well-scoped documentation clarification. Looks good to approve.

PR #8609, opened 2026-08-13
Use HashCode#getBytesInternal in combinedOrdered / combineUnordered

kilink:
The methods combinedOrdered and combineUnordered in Hashing can use getBytesInternal.

### DIFF
--- a/android/guava/src/com/google/common/hash/Hashing.java
+++ b/android/guava/src/com/google/common/hash/Hashing.java
@@ -635,7 +635,7 @@ public final class Hashing {
     int bits = iterator.next().bits();
     byte[] resultBytes = new byte[bits / 8];
     for (HashCode hashCode : hashCodes) {
-      byte[] nextBytes = hashCode.asBytes();
+      byte[] nextBytes = hashCode.getBytesInternal();
       checkArgument(
           nextBytes.length == resultBytes.length, "All hashcodes must have the same bit length.");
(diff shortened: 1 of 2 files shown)


cpovirk:
Makes sense. I see a modestly significant amount of time in the `combine*` methods inside Google (though apparently negligible _RAM_, noting that I don't have any numbers for any impact on GC frequency). It doesn't _appear_ to be in `asBytes()`. (And the tiny bit of time I do see in `asBytes()` is for `LongHashCode`.) But there's a lot of time that's not attributed to any child method, and I don't know whether that indicates limitations of profiling or actual work done directly in `combine*`.

Anyway, I'll get this merged, and I'll check back in a month or so to see if the impact is noticeable for us.

cpovirk:
Huh, _all_ the numbers here look negligible, before and after :) I may have had an off-by-a-factor-of-a-thousand error here, imagining that I saw a "k" somewhere that I didn't? Anyway, given that, it's hard to read much into the diffs. I actually see numbers that are _worse_ (but again, in the range that totally does not matter at all), so I'm even more inclined than before to write this off to "limitations of profiling." I suppose that it is not completely and totally out of the question that, in the cases that `getBytesInternal` is going to end up delegating to `asBytes`, anyway, the additional indirection slows us down. But did I mention that the numbers suggest that none of this difference matters to us? :)

PR #8627, opened 2026-08-24
Update actions/setup-java to v6

brunoborges:
Updates all pinned `actions/setup-java` references in the CI workflow from v5.6.0 to the v6.0.0 commit.

Validation: verified all active workflow references use setup-java v6 and ran `git diff --check`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
       # - 11 (sometimes) to *download* to support anyone who runs JDiff or our Gradle integration tests (including our doc snapshots and our Java 11 CI test run) but not to use directly
       # - 26 for running Javadoc and javac (to help people who build Guava locally and might not use a recent JDK to run Maven)
       - name: 'Set up JDKs'
-        uses: actions/setup-java@03ad4de0992f5dab5e18fcb136590ce7c4a0ac95 # v5.6.0
+        uses: actions/setup-java@dd06d9cba3e5552c54d9f8ea23572deb30010f7c # v6.0.0
         with:
           java-version: |
             ${{ matrix.java }}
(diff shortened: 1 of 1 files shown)


cpovirk:
I'm guessing that you're sending us a PR so soon after [release](https://github.com/actions/setup-java/releases/tag/v6.0.0) because of the [changes to automatically verify checksums](https://github.com/actions/setup-java/blob/main/README.md#v6:~:text=JDK%20downloads%20now%20automatically%20verify%20authoritative%20checksums) (including for Temurin, which is [what we use](https://github.com/google/guava/blob/6393ebbdb62f05edc2d320a7bc44b86a046642a8/.github/workflows/ci.yml#L49)). It's possible that you're also making any manual upgrades to the configuration that the upgrade requires, though I can't confirm without the CLA.

Dependabot should propose the basic dependency bump [within a month of when our seven-day cooldown expires](https://github.com/google/guava/blob/6393ebbdb62f05edc2d320a7bc44b86a046642a8/.github/dependabot.yml#L30-L32). I'm open to doing this earlier if you think it's worthwhile to sort through the CLA.

cpovirk:
While touching setup-java, Gemini suggests that we may be able to remove our [`-ntp`](https://github.com/google/guava/blob/6393ebbdb62f05edc2d320a7bc44b86a046642a8/.github/workflows/ci.yml#L53-L56) flag, since that's nowadays added by default. Apparently that default goes back to [5.5.0](https://github.com/actions/setup-java/commit/6c4d4a5025dca6ad7cc86c839fadbcb66762ed3b).

brunoborges:
my apologies for this PR. It was only supposed to hit a set of repos with a specific dependabot configuration that would prevent getting the alert of this new release

PR #8643, opened 2026-08-30
Use `SUPPORTS_REMOVE` for `CollectionRemoveIfTester` feature requirements.

### LINKED ISSUE #6076: [Bug] GuavaTestLib RemoveIf Flag Inconsistencies

Divyansh151005:
## Problem

`CollectionRemoveIfTester` was inconsistent about which collection feature flags gate `removeIf` tests:

- Positive `removeIf` tests required `SUPPORTS_ITERATOR_REMOVE`.
- Unsupported-case tests correctly used `absent = SUPPORTS_REMOVE`.

Collections that support `removeIf` (and `remove`) but not iterator removal therefore skipped all `removeIf` coverage in Guava Testlib.

Fixes #6076

## Solution

Require `SUPPORTS_REMOVE` for the positive `removeIf` tests, matching `CollectionRemoveTester` and the unsupported tests in the same class.

Add `CollectionRemoveIfFeatureTest`, which builds a suite for a collection with `SUPPORTS_REMOVE` but without `SUPPORTS_ITERATOR_REMOVE` and verifies that `removeIf` tests are included.

## Testing

```
$ export JAVA_HOME=~/.m2/jdks/jdk-26.0.2+10/Contents/Home
$ ./mvnw -B -Dtoolchain.skip -P!standard-with-extra-repos test \
    -Dmaven.javadoc.skip=true -Dsurefire.toolchain.version=17 \
    -Dtest=CollectionRemoveIfFeatureTest -DfailIfNoTests=false -pl guava-testlib

[INFO] Running com.google.common.collect.testing.CollectionRemoveIfFeatureTest
[INFO] Tests run: 1, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

### DIFF
--- a/guava-testlib/src/com/google/common/collect/testing/testers/CollectionRemoveIfTester.java
+++ b/guava-testlib/src/com/google/common/collect/testing/testers/CollectionRemoveIfTester.java
@@ -17,7 +17,6 @@
 package com.google.common.collect.testing.testers;
 
 import static com.google.common.collect.testing.features.CollectionFeature.FAILS_FAST_ON_CONCURRENT_MODIFICATION;
-import static com.google.common.collect.testing.features.CollectionFeature.SUPPORTS_ITERATOR_REMOVE;
 import static com.google.common.collect.testing.features.CollectionFeature.SUPPORTS_REMOVE;
 import static com.google.common.collect.testing.features.CollectionSize.SEVERAL;
 import static com.google.common.collect.testing.features.CollectionSize.ZERO;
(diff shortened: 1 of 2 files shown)


cgdecker:
We can't review your PR unless you sign the CLA.

Issue #8661, opened 2026-09-05
MultiReader.read() can throw StackOverflowError with many empty CharSources


cuishuang:
### Guava Version

33.7.1-jre

### Description

`MultiReader.read(char[], int, int)` recursively calls itself whenever the current reader reaches EOF.

`CharSource.concat(...)` uses `MultiReader`, so concatenating a large number of empty `CharSource` instances can exhaust the Java call stack even though the input is valid.

The corresponding `MultiInputStream` implementation already advances through empty streams using a loop.

### Example

```java
import com.google.common.io.CharSource;
import java.io.Reader;
import java.util.Collections;

final class MultiReaderStackOverflow {
  public static void main(String[] args) throws Exception {
    try (Reader reader =
        CharSource.concat(
                Collections.nCopies(100_000, CharSource.empty()))
            .openStream()) {
      System.out.println(reader.read());
    }
  }
}
```

### Expected Behavior

The read should return -1 after advancing through all empty sources.

### Actual Behavior

The read operation can throw StackOverflowError because each empty reader adds one recursive stack frame.

### Packages

com.google.common.io

### Checklist

- [x] I agree to follow the [code of conduct](https://github.com/google/.github/blob/master/CODE_OF_CONDUCT.md).

- [x] I can reproduce the bug with the [latest](https://github.com/google/guava/releases/latest) version of Guava available.

Issue #2996, opened 2017-11-21
MultiInputStream produces stack overflow


jkoroid:
Using `guava-23.0` (and earlier versions), `com.google.common.io.MultiInputStream` produces a stack overflow when it is given sufficiently many empty `ByteSource` instances.

Test cases that reproduce the behavior:

```java
import java.io.InputStream;
import java.util.List;

import org.junit.Before;
import org.junit.Test;

import com.google.common.collect.Lists;
import com.google.common.io.ByteSource;

public class MultiInputStreamTest {

  /** Adjust this according to the thread stack size. */
  private static final int LARGE_ENOUGH_FOR_CURRENT_STACK_SIZE = 1000000;

  private ByteSource       concatenated;

  public @Before void setup() throws Exception {
    final List<ByteSource> emptySources = Lists.newArrayList();
    for (int i = 0; i < LARGE_ENOUGH_FOR_CURRENT_STACK_SIZE; i++) {
      emptySources.add(ByteSource.empty());
    }
    // produce a ConcatenatedByteSource which in turn produces
    // a MultiInputStream
    concatenated = ByteSource.concat(emptySources);
  }

  public @Test void stackOverflowErrorOnEmptySources1() throws Exception {
    try (InputStream in = concatenated.openStream()) {
      in.read();
    }
  }

  public @Test void stackOverflowErrorOnEmptySources2() throws Exception {
    try (InputStream in = concatenated.openStream()) {
      in.read(new byte[1], 0, 1);
    }
  }

}
```

The condensed exception stack traces produced by the respective test case (which appear to be truncated):

```
$ uniq -c stack1
   1 java.lang.StackOverflowError
   1 	at com.google.common.io.ByteSource$ByteArrayByteSource.openStream(ByteSource.java:574)
   1 	at com.google.common.io.MultiInputStream.advance(MultiInputStream.java:65)
   1 	at com.google.common.io.MultiInputStream.read(MultiInputStream.java:89)
1021 	at com.google.common.io.MultiInputStream.read(MultiInputStream.java:90)
```

```
$ uniq -c stack2
   1 java.lang.StackOverflowError
   1 	at com.google.common.io.MultiInputStream.close(MultiInputStream.java:52)
   1 	at com.google.common.io.MultiInputStream.advance(MultiInputStream.java:63)
   1 	at com.google.common.io.MultiInputStream.read(MultiInputStream.java:102)
1021 	at com.google.common.io.MultiInputStream.read(MultiInputStream.java:103)
```

PR #8662, opened 2026-09-05
Avoid recursion when MultiReader advances past empty readers

cuishuang:
Addresses  https://github.com/google/guava/issues/8661

### Summary

- Replace recursive EOF advancement in `MultiReader.read(char[], int, int)` with an iterative loop.
- Add a regression test covering 100,000 consecutive empty `CharSource` instances.

This prevents `StackOverflowError` for valid concatenations containing many empty sources and matches the existing iterative behavior in `MultiInputStream`.


eamonnmcmanus:
Thanks! This closely parallels the fix for #2996 back in 2017.

PR #8666, opened 2026-09-08
fix: avoid stack overflow in MultiReader

### LINKED ISSUE #8661: MultiReader.read() can throw StackOverflowError with many empty CharSources

rksharma-owg:
Fixes #8661

## Summary
- replace recursive empty-source traversal in `MultiReader.read` with an iterative loop
- add a regression test covering 100,000 empty sources

## Validation
- `JAVA_HOME=/opt/homebrew/opt/openjdk@17 ./mvnw -pl guava-tests -DskipAndroid=true -Dcheckstyle.skip=true test`
- focused `MultiReaderTest` passed after the fix

The regression was reproduced before the fix as a `StackOverflowError` while reading the concatenated source.

### DIFF
--- a/guava-tests/test/com/google/common/io/MultiReaderTest.java
+++ b/guava-tests/test/com/google/common/io/MultiReaderTest.java
@@ -79,6 +79,16 @@ public class MultiReaderTest extends TestCase {
     assertThat(CharStreams.toString(joinedReader)).isEqualTo(expectedString);
   }
 
+  public void testReadWithManyEmptySourcesDoesNotOverflowStack() throws Exception {
+    ImmutableList.Builder<CharSource> sources = ImmutableList.builder();
+    for (int i = 0; i < 100_000; i++) {
+      sources.add(newCharSource(""));
+    }
+
+    Reader joinedReader = CharSource.concat(sources.build()).openStream();
+    assertEquals(-1, joinedReader.read());
+  }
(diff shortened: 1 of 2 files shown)


rksharma-owg:
Closing as duplicate of #8662, which addresses the same MultiReader recursion/StackOverflowError issue.

PR #8677, opened 2026-09-17
Fix ArrayStoreException in ObjectArrays.concat(element, array) / concat(array, element)

### LINKED ISSUE #3768: ObjectArrays.concat(T[], T) and (T, T[]) can throw ArrayStoreException
I doubt that anyone has ever hit this in practice. (I didn't see any bugs filed internally about it, anyway, nor any mentions of it on the web.) But I thought it was neat :)

```java
$ cat ConcatTs.java 
import static com.google.common.collect.ObjectArrays.concat;

public class ConcatTs {
    String[] array = new String[0];
    Object[] concatenated = concat(array, 0);
  }
}
```

```
 $ javac -cp $HOME/.m2/repository/com/google/guava/guava/28.2-jre/guava-28.2-jre.jar ConcatTs.java && java -cp .:$HOME/.m2/repository/com/google/guava/guava/28.2-jre/guava-28.2-jre.jar ConcatTs
Exception in thread "main" java.lang.ArrayStoreException: java.lang.Integer
  at com.google.common.collect.ObjectArrays.concat(ObjectArrays.java:102)
  at ConcatTs.main(ConcatTs.java:6)
```

Interestingly, we got this right in `concat(T[], T[], Class<T>)` -- which requires a `Class` to use to create the array -- back in CL 3734425. But we got it wrong in the new methods in CL 7931853.

Junnie18:
## Summary

`ObjectArrays.concat(T element, T[] array)` and `ObjectArrays.concat(T[] array, T element)` build their result array with the same runtime component type as the input array (via `newArray`/`Arrays.copyOf`), then store `element` directly into it. If `element`'s runtime type isn't assignable to that component type — which erasure allows the caller to do, since both methods are generic — storing it throws `ArrayStoreException` instead of succeeding:

```java
String[] array = new String[0];
Object[] concatenated = ObjectArrays.<Object>concat(array, 0); // throws ArrayStoreException
```

This is inconsistent with `ObjectArrays.concat(T[], T[], Class<T>)`, which never has this problem because it always builds its result array from the explicitly supplied `Class<T>` rather than inferring a component type from one of its array arguments.

## Fix

Each of the two single-element overloads now checks whether `element` is storable in an array with the same component type as the input array before allocating the result that way. If it isn't, it falls back to a plain `Object[]` (unchecked-cast back to `T[]`), which mirrors the existing `unsoundlyCovariantArray` pattern already used elsewhere in this file for the same class of unsoundness. In the overwhelmingly common case (element and array are actually type-compatible), behavior and the returned array's runtime type are unchanged.

No public API was changed.

## Testing

Added 4 regression tests to `ObjectArraysTest` (prepend/append, with both an empty and a non-empty backing array) that concatenate a `String[]` with an `Integer` via an explicit `<Object>` type witness, reproducing the exact shape of bug report's repro. Verified:
- Before the fix: all 4 new tests fail with `ArrayStoreException`, confirming they exercise the reported bug.
- After the fix: `mvn test -pl guava,guava-tests -am -Dtest=ObjectArraysTest` → 30/30 pass.
- Ran the full `com.google.common.collect.*Test` suite (207 test classes) afterward to check for regressions: all pass.
- Formatted the changed source file with `google-java-format`.

Fixes #3768.


ionik0:
Hey @Junnie18 ... go to https://cla.developers.google.com/  and sign in , or the googleBot  will block it, individual CLA is a few minutes.

cpovirk:
As noted above, we can't look at the code without a CLA. But judging from the description, it sounds like this change would cause us to return the wrong type.

I don't think the underlying problem here has a real "fix" short of deprecating the API and migrating all callers to a fixed API. Given that I don't think we've heard of anyone who hit this problem for real, we are unlikely to do even that.

